PR #8, opened 2022-03-29
Dev

Luthaf:


ceriottm:
Let's do all the renaming here, so we don't have to keep updating stuff on the examples side. There was also `sparse -> keys` and `Descriptor -> SparseTensor` (both TBD)

Luthaf:
Replaced by #14

PR #11, opened 2022-04-01
Allow gradients to use a super-set of the values components

Luthaf:
This should make gradients of equivariant descriptors easier to manage


Luthaf:
#13 feels like a better solution to this

PR #22, opened 2022-05-10
Implement serialization of TensorMap based on numpy's NPZ format

Luthaf:


PicoCentauri:
Nice! I can do a review if you like.

Luthaf:
That would be great. I split this PR into three independent commits, it should help with the review!

equistore/io.py
```diff
@@ -0,0 +1,134 @@
...
+    for more information on the format.
+
+    :param path: path of the file where to save the data
+    :param tensor: tensor to save
+    :param use_numpy: should we use numpy or the native implementation? The main
+        difference is that numpy should be able to save more dtypes for the data
+        than the native implementation, which is limited to float64.
+    """
+    if not path.endswith(".npz"):
+        # warning?
```
PicoCentauri, line 45: I would add the ending explicitly then.

equistore/io.py
```diff
@@ -0,0 +1,134 @@
...
+    """Save the given :py:class:`equistore.TensorMap` to a file at ``path``.
+
+    :py:class:`equistore.TensorMap` are serialized using numpy's ``.npz``
+    format, i.e. a ZIP file without compression (storage method is ``STORED``),
+    where each file is stored as a ``.npy`` array. See the C API documentation
+    for more information on the format.
+
+    :param path: path of the file where to save the data
+    :param tensor: tensor to save
+    :param use_numpy: should we use numpy or the native implementation? The main
```
PicoCentauri, line 40: why is `use_numpy` false by default. The docstring sounds like it is beneficial to use numpy.

equistore/io.py
```diff
@@ -0,0 +1,134 @@
...
+        return array
+    elif _is_torch_array(array):
+        return array.cpu().numpy()
+    else:
+        raise ValueError("unknown array type passed to `equistore.io.save`")
+
+
+def _tensor_map_to_dict(tensor_map):
+    result = {}
+    result["keys"] = tensor_map.keys
```
PicoCentauri, line 67:
```python
result = {"keys": tensor_map.keys}
```

to save a line

PicoCentauri (commented):
Looks good and structured. The inner file structure looks also understandable. `npy_headers.py` is a 1to1 copy? I was wondering that there is no check for latin1 and utf8...

equistore/io.py
Luthaf: I don't really like "magic" behavior like this, I would rather send a warning and have the user fix its code.

equistore/io.py
Luthaf: The non-numpy version should be slightly faster since it construct Rust struct directly instead of building blocks in Python first (although I did not benchmark this). Also, using the non-numpy version by default allows to find bugs/issues in the native version faster =)

Luthaf:
> npy_headers.py is a 1to1 copy?

Yes, I just added a bit of error propagation integration at the end.

> I was wondering that there is no check for latin1 and utf8...

I'm not sure what you mean? Rust `str`/`String` types are guaranteed to be UTF8, so that removes the need to do a bunch of encoding checks.

equistore/io.py
PicoCentauri: A warning and saving the file without the npz ending?

equistore/io.py
PicoCentauri: Okay then we should maybe add the higher perfomance of the native implementation to the docstring. Python users are used to the fact that numpy is the fastest you can do :-).

PicoCentauri:
Makes sense. I just had a quick look at the docu of npz and they said that they changed from latin1 to utf8 in version 3.

equistore/io.py
Luthaf: Hmm, I checked how numpy handles this, and it automagically adds the extension to the file ... Let's go for a warning and saving the file with an extension to follow numpy then.

equistore/io.py
PicoCentauri: Really cool. So we should add this information to the docs and motivate people use the native format.

Luthaf: I've added a small sentence on this, it should be good to go =)

PR #23, opened 2022-05-19
Rust API improvement for integration with rascaline

Luthaf:


PicoCentauri:
Looks clean and solid. I have no objections and you probably tested everything because there is not a single test case added ;-). But I think this is okay for now.

Luthaf:
> you probably tested everything because there is not a single test case added ;-)

yeah, I used these function in a rascaline branch to compute power spectrum & friends. Let me see if I can add some test to this PR, it is always good to have some

Issue #32, opened 2022-06-21
Add a `__repr__` function to TensorMap and TensorBlock


Luthaf:
Right now we get not very informative output:

```py
>>> tensor
<equistore.tensor.TensorMap at 0x136bcd6f0>

>>> tensor.block(0)
<equistore.block.TensorBlock at 0x1044768c0>
```

---
Suggested by @davidetisi

Luthaf:
Implemented in #33

PR #33, opened 2022-06-22
Added first attempt of __repr__ for TensorMap and TensorBlock

DavideTisi:
In the spirit of issue #32. I added an idea for two `__repr__` functions for TensorMap and TensorBlock.
In TensorMap it basically prints the keys and its names:
```
>> Tensor
TensorMap 
keys: [(0, 1, 1), (0, 1, 6), (0, 1, 8), (0, 6, 1), (0, 6, 6), (0, 6, 8), (0, 8, 1), (0, 8, 6), (0, 8, 8)]
key_names: ['spherical_harmonics_l', 'species_center', 'species_neighbor']
```
For the TensorBock, the function prints the names of the samples, components and properties:

```
>> Tensor.block(0)
TensorBlock 
samples: ['structure', 'center']
component: ['spherical_harmonics_m']
properties: ['n']
```

The example are taken from a spherical harmonic expansion Tensor.

### DIFF
--- a/tests/python/blocks.py
+++ b/tests/python/blocks.py
@@ -1,4 +1,5 @@
 import unittest
+import os
 
 import numpy as np
 
@@ -6,6 +7,16 @@ from equistore import Labels, TensorBlock
 
 
 class TestBlocks(unittest.TestCase):
+    def test_repr_(self):
+        block = TensorBlock(
+            values=np.full((3, 2), -1.0),
+            samples=Labels(["samples"], np.array([[0], [2], [4]], dtype=np.int32)),
+            components=[],
+            properties=Labels(["properties"], np.array([[5], [3]], dtype=np.int32)),
+        )
+        expected = "TensorBlock \nsamples: ['samples']\ncomponent: []\nproperties: ['properties']\ngradients: no"
+        self.assertTrue(block.__repr__() == expected)
+
     def test_block_no_components(self):
         block = TensorBlock(
             values=np.full((3, 2), -1.0),
--- a/tests/python/tensor.py
+++ b/tests/python/tensor.py
@@ -1,7 +1,8 @@
 import unittest
+import os
 
 import numpy as np
-from utils import test_tensor_map
+from utils import test_tensor_map, test_large_tensor_map
 
 
 class TestTensorMap(unittest.TestCase):
@@ -9,10 +10,39 @@ class TestTensorMap(unittest.TestCase):
         tensor = test_tensor_map()
         self.assertEqual(tensor.keys.names, ("key_1", "key_2"))
         self.assertEqual(len(tensor.keys), 4)
+        self.assertEqual(len(tensor), 4)
         self.assertEqual(tuple(tensor.keys[0]), (0, 0))
         self.assertEqual(tuple(tensor.keys[1]), (1, 0))
         self.assertEqual(tuple(tensor.keys[2]), (2, 2))
         self.assertEqual(tuple(tensor.keys[3]), (2, 3))
+
+    def test_print(self):
+        """
+        Test routine for the print function of the TensorBlock.
+        It compare the reults with those in a file.
+        """
+        tensor = test_tensor_map()
+        repr = tensor.__repr__()
+        expected = """TensorMap with 4 blocks
+keys: ['key_1' 'key_2']
+          0       0
+          1       0
+          2       2
+          2       3"""
+        self.assertTrue(expected == repr)
+
+        tensor = test_large_tensor_map()
+        _print = tensor.__repr__()
+        expected = """TensorMap with 12 blocks
+keys: ['key_1' 'key_2']
+          0       0
+          1       0
+          2       2
+       ...
+          1       5
+          2       5
+          3       5"""
+        self.assertTrue(expected == _print)
 
     def test_labels_names(self):
         tensor = test_tensor_map()
--- a/tests/python/utils.py
+++ b/tests/python/utils.py
@@ -79,3 +79,59 @@ def test_tensor_map():
     )
 
     return TensorMap(keys, [block_1, block_2, block_3, block_4])
+
+
+def test_large_tensor_map():
+    """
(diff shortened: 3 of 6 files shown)


Luthaf:
The build failure is unrelated, and should be fixed on master. Can you rebase this branch to bring in the fix?

Some small suggestions for the output
```bash
TensorMap with 9 blocks  # add the number of blocks
# how about a vertical layout? takes more space but might be easier to read
keys: ['spherical_harmonics_l', 'species_center', 'species_neighbor'] 
               0,                        1,               1,
               0,                        1,               6,

                                        ... # skip some keys if they are more than 10?
                                        
               0,                        8,               6,
               0,                        8,               8,
```

If you agree with this, maybe we can skip keys only in `__repr__` and not in `__print__` (similarly to how numpy prints arrays)
---

For the tensor block, I would also add some information about the gradients

DavideTisi:
I agree for sure with the addition of the number of blocks.
the vertical layout how you would suggest aligning the numbers with just tabs?
I mean if the keys_names changes it would be difficult are different to put the number under the label.

Regarding the skip, I agree to put it only in __repr__ (actually that's why I haven't put it, normally I do not like when it skips stuff.

I can complete the TensorBlock with gradients. Here do we put full infos or just if there are gradients or not? and then a __repr__ (or similar) function for the gradient only?

Luthaf:
>  the vertical layout how you would suggest aligning the numbers with just tabs?
> I mean if the keys_names changes it would be difficult are different to put the number under the label.

There is some code doing this kind of alignment in rust, here: https://github.com/lab-cosmo/equistore/blob/8823476d864409618afb23c96dc86affc767b160/src/labels.rs#L229-L244

It is not too bad: you can use the length of the different variables as padding when writing the values. I think python formatting can take the padding width as an extra parameter (not sure what's the syntax for it though).

> I can complete the TensorBlock with gradients. Here do we put full infos or just if there are gradients or not? and then a repr (or similar) function for the gradient only?

I think just having which gradients are present or not is fine, and then having another `__repr__` function in the `Gradient` proxy object is a great idea!

DavideTisi:
The TensorMap prints the number of blocks and the values in a table-like way.
The padding for sure can be improved (also the code is not super good-looking).
 
If the number of blocks is < 10, both functions do not skip any value.
```
>> Tensor
TensorMap with 9 blocks 
keys: ['spherical_harmonics_l' 'species_center' 'species_neighbor]
                  0                   1                1
                  0                   1                6
                  0                   1                8
                  0                   6                1
                  0                   6                6
                  0                   6                8
                  0                   8                1
                  0                   8                6
                  0                   8                8

>> print(Tensor)
TensorMap with 9 blocks 
keys: ['spherical_harmonics_l' 'species_center' 'species_neighbor]
                  0                   1                1
                  0                   1                6
                  0                   1                8
                  0                   6                1
                  0                   6                6
                  0                   6                8
                  0                   8                1
                  0                   8                6
                  0                   8                8
```

If the number of blocks is > 10.


```
>> Tensor
TensorMap with 63 blocks 
keys: ['spherical_harmonics_l' 'species_center' 'species_neighbor]
                  0                   1                1
                  1                   1                1
                  2                   1                1
               ...
                  4                   8                8
                  5                   8                8
                  6                   8                8
```

`print` function never skip.


The TensorBlock output now gives info on the gradients

If there are no gradients:
```
TensorBlock 
samples: ['structure', 'center']
component: ['spherical_harmonics_m']
properties: ['n']
gradients: No
```

If there are gradients, it gives the list of parameters wrt the gradients are computed:

```
>> Tensor.block(0)
TensorBlock 
samples: ['structure', 'center']
component: ['spherical_harmonics_m']
properties: ['n']
gradients: ['positions']
```

Finally the gradients block:
```
>> Tensor.block(0).gradient('positions')
Gradient TensorBlock 
parameter: 'positions'
samples: ['sample', 'structure', 'atom']
component: ['gradient_direction', 'spherical_harmonics_m']
properties: ['n']
```

Luthaf:
I like the new output! 

The build will fail again, you'll need to reformat the code with [black](https://github.com/psf/black). You can do this manually with `tox -e format` or integrate it automatically in your editor with https://github.com/microsoft/vscode-black-formatter (I know you are using vscode). The automatic integration will automatically format all python files when you save them.

Luthaf (commented):
A couple of small suggestions, but the output looks fine to me. Could you also add some basic tests?

tests/python/blocks.py
```diff
@@ -1,11 +1,25 @@
...
+            values=np.full((3, 2), -1.0),
+            samples=Labels(["samples"], np.array([[0], [2], [4]], dtype=np.int32)),
+            components=[],
+            properties=Labels(["properties"], np.array([[5], [3]], dtype=np.int32)),
+        )
+        ROOT = os.path.dirname(__file__)
+        tfile = os.path.join(ROOT, "..", "repr_string_block.data")
+        with open(tfile, "r") as f:
+            s = f.readlines()
+        self.assertTrue(block.__repr__() == "".join(s))
```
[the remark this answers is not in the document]
DavideTisi: you're right

tests/python/utils.py
```diff
@@ -3,10 +3,12 @@
 from equistore import Labels, TensorBlock, TensorMap
 
 
-def test_tensor_map():
+def test_tensor_map(iflong: bool = False):
```
Luthaf, line 6: You can remove the `iflong` parameter now =)

tests/python/tensor.py
```diff
@@ -1,19 +1,40 @@
...
+    def test_print(self):
+        """
+        Test routine for the print function of the TensorBlock.
+        It compare the reults with those in a file.
+        """
+        tensor = test_tensor_map()
+        repr = tensor.__repr__()
+        ROOT = os.path.dirname(__file__)
+        tfile = os.path.join(ROOT, "..")
+        with open(tfile + "/repr_string.data", "r") as f:
```
Luthaf, line 28: could you move the data out of the files and directly inside the code here as well?

PR #35, opened 2022-07-01
Fix use after free when accessing rust array inside a TensorBlock

Luthaf:
This fixes the issue I saw in https://github.com/lab-cosmo/scikit-cosmo/pull/138#issuecomment-1172454872

### DIFF
--- a/equistore/block.py
+++ b/equistore/block.py
@@ -5,7 +5,12 @@ from typing import Generator, List, Tuple
 
 from ._c_api import eqs_array_t, eqs_labels_t
 from ._c_lib import _get_library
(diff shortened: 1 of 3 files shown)


ceriottm:
Very nice. Does this come at a substantial cost (e.g. overhead every time we access the object), or is just "cosmetic"?

Luthaf:
It is mostly cosmetic. Previously we would return a `numpy.ndarray` instance from the `values` property, loosing any link with the memory owner; and now we return a sub-class of `numpy.ndarray` which also holds a reference to the memory owned (`parent` in the code). 

Everything else works the same, we just keep a reference to the parent around

PR #37, opened 2022-07-08
TensorSlice helper to select samples and/or features from a Tensormap

ceriottm:
This facilitates a bit the usage of TensorMap, and serves as an example of one could implement utilities to manipulate equistore objects.


Luthaf:
Replaced by #77

PR #46, opened 2022-09-07
added __getitem()__ and added a check for multiple positional arguments in tensor.block and tensor.blocks

DavideTisi:
-- added a few checks in the length of `args` before `tensor.block(0,5)==tensor.block(0)`, because it checked only that the first argument was an `int` and then returned the result. Now the code raises an error:

`raise ValueError(
    f"only one non-keyword argument is supported, {len(args)} are given"
)`

This behaviour was actually true only if the first positional argument was an `int` because a similar check was done by the `blocks_matching` function, which was not called if `args[0]` was `int`. I also left the check inside `blocks_matching` even though now it is probably never called.

-- added the `__getitem__()` function to `TensorMap`.
  Now, `tensor[0]` gives the same result of `tensor.block(0)`
  it works also for keys: `tensor[tensor.keys[0]]` is equivalent to `tensor.block(tensor.keys[0])`. Different behaviour if no input is given:
```
     tensor[] 
           ^ 
SyntaxError: invalid syntax
```
while 
```
tensor.block()
  File "/Users/davidetisi/Documents/EPFL_postdoc/equistore/.tox/tests/lib/python3.10/site-packages/equistore/tensor.py", line 161, in block
    raise ValueError(
ValueError: more than one block matched {}, use `TensorMap.blocks` if you want to get all of them

```

tensor.block() fails because it selects all the blocks, e.g. tensor.blocks() gives all the block
I think this difference is acceptable.

I added tests for the additions above.

### DIFF
--- a/equistore/tensor.py
+++ b/equistore/tensor.py
@@ -96,6 +96,14 @@ class TensorMap:
         )
         return result
 
+    def __getitem__(self, *args) -> TensorBlock:
+        """This is equivalent to self.block(*args)"""
+        if args and isinstance(args[0], tuple):
+            raise ValueError(
+                f"only one non-keyword argument is supported, {len(args[0])} are given"
+            )
+        return self.block(*args)
+
     @property
     def keys(self) -> Labels:
         """The set of keys labeling the blocks in this tensor map."""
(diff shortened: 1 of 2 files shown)


Luthaf (approved):

Luthaf:
Looks good!

> I also left the check inside blocks_matching even though now it is probably never called.

blocks_matching is a public function, and can be called directly (if the user needs both the set of keys & corresponding blocks for a given selection)

PR #51, opened 2022-10-04
Tensor operations: solve and dot 

DavideTisi:
implementation of the operation: solve and dot between TensorMap and TensorBlock.

-  dot: gives a TensorMap ``y = dot(X,w)`` where ``y`` has the same keys as `X` and `w`.
The blocks inside `y` are the dot product of the corresponding blocks in `X` and `w`, they have the same samples of the blocks in `X` and the properties of the blocks in `w`.
- solve: given two TensorMaps it finds the `w` s.t.  `y = dot(X,w)`. The keys are always the same for all the TensorMaps. The TensorBlocks of `w` have the samples of `y` and the properties of `x`.


DavideTisi:
thinks to do 

- [ ] try linear model with soap
- [ ] now solve does automatically lstq if the input matrices are not squares, we might want to have two functions.
- [ ] run test with pythorch

DavideTisi:
I also added few help functions that we might want to eliminate:

- has_any_gradient () -> return len(self.gradients_list()) > 0
I was just tired of typing len(self.gradients_list()) > 0, I didn't use that much more in a notebook I have where I tested stuff

- compare_blocks: it compares two TensorBlocks creating a dictionary I used in the tests and to debug it is a bit of a mess and created for that purpose, but it was useful to find bugs

equistore/operations/_dispatch.py
```diff
@@ -0,0 +1,121 @@
...
+    of a system of linear equations.
+    Computes the vector x that approximately
+    solves the equation ``array1 @ x = array2``.
+    This function has the equivalent
+    behaviour of ``numpy.linalg.lstsq(array1,array2)``.
+    """
+    if isinstance(array1, np.ndarray) and isinstance(array2, np.ndarray):
+        return np.linalg.lstsq(array1, array2, rcond=rcond)[0]
+    elif isinstance(array1, TorchTensor) and isinstance(array2, TorchTensor):
+        result = torch.linalg.lstsq(array1, array2)[0]
```
Luthaf, line 99: I think torch lstsq should have a rcond parameter as well?

equistore/operations/_dispatch.py
```diff
@@ -0,0 +1,121 @@
...
+        return result
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def vstack(arrays: tuple):
```
Luthaf, line 105: this feels like the wrong type for arrays. Why not List[Union[Tensor, np.ndarray]], or similar? Or not putting in the type at all?

equistore/operations/dot.py
```diff
@@ -0,0 +1,71 @@
...
+
+from . import _dispatch
+
+
+def _dot_block(block1: TensorBlock, block2: TensorBlock) -> TensorBlock:
+    """dot product (row) among two `TensorBlocks`.
+    The `TensorBlocks` should have the same `properties`
+    """
+
+    assert np.all(block1.properties == block2.properties)
```
Luthaf, line 13: I think this should be an exception with a better error message, since the user is likely to encounter it. I tend to prefer using assertions for internal consistency checks

equistore/operations/lstsq.py
```diff
@@ -0,0 +1,71 @@
...
+        if len(X.gradients_list()) != len(Y.gradients_list()) or (
+            not np.all(
+                [parameter in Y.gradients_list() for parameter in X.gradients_list()]
+            )
+        ):
+            raise ValueError("The two input TensorBlock should have the same gradients")
+
+        for parameter, Xgradient in X.gradients():
+            X_grad_data_reshape = Xgradient.data.reshape(-1, X.values.shape[-1])
+            # TODO: this assume the atoms are in the same order in X_grad &
```
Luthaf, line 51: With TensorMap, we should be able to check this by checking the samples. The todo was here because the prototype used numpy arrays

equistore/operations/normalize.py
```diff
@@ -0,0 +1,63 @@
+from equistore import TensorBlock, TensorMap
+
+from . import _dispatch
+
+
+def normalize_by_sample(tensor: TensorMap) -> TensorMap:
```
Luthaf, line 6: let's keep this one out of this PR for now

equistore/operations/solve.py
```diff
@@ -0,0 +1,72 @@
...
+
+def solve(X: TensorMap, Y: TensorMap) -> TensorMap:
+    """
+    Solve a linear system among two :py:class:`TensorMap`.
+    Solve the linear equation set X * w = Y for the unknown w.
+    Where X , w, Y are all :py:class:`TensorMap`
+    """
+
+    if len(X) != len(Y) or (not np.all([key in Y.keys for key in X.keys])):
+        raise ValueError("The two input TensorMap should have the same keys")
```
Luthaf, line 16: this feels like an operation we will have to do a lot, should we extract it to a separate function?

Luthaf:
https://github.com/lab-cosmo/equistore/pull/52 should fix the remaining build error.

equistore/operations/_dispatch.py
DavideTisi: yes, it does but this is a thing i wanted to talk about.
The value `None` which I the default in both cases is different between torch and NumPy. In particular, the torch default is the “old“ NumPy default.

[NumPy](https://numpy.org/doc/stable/reference/generated/numpy.linalg.lstsq.html): 

> `rcond` float, optional
> Cut-off ratio for small singular values of a. For the purposes of rank determination, singular values are treated as zero if they are smaller than rcond times the largest singular value of a.
 The previous default of -1 will use the machine precision as rcond parameter, the new default will use the machine precision times max(M, N). To silence the warning and use the new default, use rcond=None, to keep using the old behavior, use rcond=-1.


[Pytorch](https://pytorch.org/docs/stable/generated/torch.linalg.lstsq.html):

>  If rcond= None (default), rcond is set to the machine precision of the dtype of A.

In pytorch they suggest to set manually a value as default to avoid changing behaviour in the future: 

> The default value of rcond may change in a future PyTorch release. It is therefore recommended to use a fixed value to avoid potential breaking changes

Do we want to do that?

equistore/operations/_dispatch.py
DavideTisi: you mean in vstack? ok, I think that since numpy uses tuples in the example I just putted tuple but yes maybe don't put anything is better

equistore/operations/normalize.py
DavideTisi: ah ok... we put it in a new branch?

equistore/operations/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .dot import _dot_block, dot  # noqa
+from .lstsq import _lstsq_block, lstsq  # noqa
+from .normalize import normalize_by_sample  # noqa
+from .remove_gradients import remove_gradients  # noqa
+from .solve import _solve_block, solve  # noqa
```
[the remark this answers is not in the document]
DavideTisi: not strong reasons, but someone might want to use them outside or not?

Luthaf (commented):
The code looks good overall, but the documentation is a bit lacking for some function (in particular, the parameter names differ between the internal and user-facing function). 

The documentation should also be added to the sphinx build!

PR #55, opened 2022-10-12
Stop keeping reference to TensorBlock in freshly allocated array

### LINKED ISSUE #53: Equistore is too keen on keeping references to tensor blocks, resulting in memory leaks
Code below is an example - any large dataset will make the problem clear. Basically values.sum() returns a _RustNDArray that holds a reference to the parent block, so the memory for the environment-level descriptors is never freed. Wrapping the .sum() in a np.array() solves the problem, but I see no reason why the reference should be kept around.

```
import numpy as np
import rascaline, equistore, itertools
import ase.io as aseio
import tqdm.notebook as tqdm

frames = aseio.read("./shiftml2_training_structures.xyz", ":")

hypers = {
    "cutoff": 6.0,
    "max_radial": 6,
    "max_angular": 4,
    "atomic_gaussian_width": 0.5,
    "radial_basis": {"Gto": {}},
    "cutoff_function": {"ShiftedCosine": {"width": 0.5}},
    "center_atom_weight": 1.0,    
}

calculator = rascaline.SoapPowerSpectrum(**hypers)

species = np.unique(np.concatenate([f.numbers for f in frames]))
feats_frm = []
for f in tqdm.tqdm(frames): 
    rho2i = calculator.compute(f)
    rho2i.keys_to_properties(equistore.Labels(['species_neighbor_1', 'species_neighbor_2'],
                                              values=np.array(list(itertools.product(species, species)), dtype=np.int32)) )
    print(rho2i.block(0).values.shape)
    feats_frm.append(rho2i.block(0).values.sum(axis=0))
```

Luthaf:
Fix #53

### DIFF
--- a/equistore/data/extract.py
+++ b/equistore/data/extract.py
@@ -168,3 +168,17 @@ class _RustNDArray(np.ndarray):
     def __array_finalize__(self, obj):
         # keep the parent around when creating sub-views of this array
         self._parent = getattr(obj, "_parent", None)
+
+    def __array_wrap__(self, new):
+        self_ptr = self.ctypes.data
+        self_size = self.nbytes
+
+        new_ptr = new.ctypes.data
+
+        if self_ptr <= new_ptr <= self_ptr + self_size:
+            # if the new array is a view inside memory owned by self, wrap it in
+            # a _RustNDArray
+            return super().__array_wrap__(new)
+        else:
+            # return the ndarray straight away
+            return new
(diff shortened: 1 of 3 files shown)


Luthaf:
@ceriottm could you check this fixes the issue on your side, using the prebuilt wheels from CI (cf above)?

ceriottm:
Tested, and I can confirm this fixes #53

PicoCentauri (approved):

PR #57, opened 2022-10-17
Small refactor for Python code

Luthaf:
Move the code from `/equistore` to `python/src/equistore`, making space for a future torch module to live in `/python/torch/...`.


PicoCentauri:
Your plan is that the torch module is independent of the python equistore package itself but also be installable via pip with independent setup files?

setup.cfg
```diff
@@ -15,5 +15,10 @@ python_requires = >=3.6
 
 [options]
 zip_safe = False
-packages = find:
 install_requires = numpy
+packages=find:
```
PicoCentauri, line 19: This is deprecated/renamed and will raise a `Package would be ignored"` warning during installation.

```suggestion
packages = find_namespace:
```

PicoCentauri (changes requested):

Luthaf:
> Your plan is that the torch module is independent of the python equistore package itself but also be installable via pip with independent setup files?

I'm not sure. We could have an equistore-torch package, which conditionally re-export as `equistore.torch` if it is installed, or have the torch code inside the equistore package, always build wheels for it, and conditionally load the corresponding module if torch is installed on the user's machine.

The first option makes it a bit more complex for users (they now need to install more than one package), while the second one makes it a bit harder for developers (they need to build the torch code even if they don't care) and easier for developers (any breaking change on equistore is immediately propagated to equistore-torch)

setup.cfg
Luthaf: This is not how I understand https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#custom-discovery. `find:` is for standard packages and `find_namespace:` for namespace packages

PicoCentauri: If pip doesn't complain, it doesn't look like in my local test, I will also not.

PicoCentauri:
> > Your plan is that the torch module is independent of the python equistore package itself but also be installable via pip with independent setup files?
> 
> I'm not sure. We could have an equistore-torch package, which conditionally re-export as `equistore.torch` if it is installed, or have the torch code inside the equistore package, always build wheels for it, and conditionally load the corresponding module if torch is installed on the user's machine.
> 
> The first option makes it a bit more complex for users (they now need to install more than one package), while the second one makes it a bit harder for developers (they need to build the torch code even if they don't care) and easier for developers (any breaking change on equistore is immediately propagated to equistore-torch)

So the more or less the same "problem/task" we have to do with equisolve? I would go for one package if this is somehow feasable. To package might lead to a lot of confusion and also code sharing might be a bit harder...

If this is a bit more effort for the developer I am okay.

PicoCentauri (approved):

PR #59, opened 2022-10-21
Miscelaneous improvements

Luthaf:
Mainly a follow up to #57, and improving the error message for wrong label names


PicoCentauri (approved):

Luthaf:
I would tend to keep separate commits, since they touch different places of the code

PR #74, opened 2022-11-18
added reduce function over the samples.

DavideTisi:
it takes as input the tensorMap and the names of the samples you want to sum over. for example if sample_name = ["structure"] it sum over all the sample with the same structure if sample_name = ["structure","center"] it sums over all the rows with the same values for the "structure" and the "center"

it works with the gradients too.

### DIFF
--- a/docs/src/reference/python/operations/sum_over_samples.rst
+++ b/docs/src/reference/python/operations/sum_over_samples.rst
@@ -1,4 +1,4 @@
 sum_over_samples
 ================
 
-.. autofunction:: equistore.operations.reduce_over_samples.sum_over_samples
+.. autofunction:: equistore.operations.sum_over_samples
--- a/docs/src/reference/python/operations/sum_over_samples.rst
+++ b/docs/src/reference/python/operations/sum_over_samples.rst
@@ -1,4 +1,0 @@
-sum_over_samples
-================
-
-.. autofunction:: equistore.operations.sum_over_samples
--- a/docs/src/reference/python/operations/sum_over_samples.rst
+++ b/docs/src/reference/python/operations/sum_over_samples.rst
@@ -1,0 +1,4 @@
+sum_over_samples
+================
+
+.. autofunction:: equistore.operations.reduce_over_samples.sum_over_samples
--- a/python/src/equistore/operations/_dispatch.py
+++ b/python/src/equistore/operations/_dispatch.py
@@ -40,6 +40,25 @@ def allclose(a, b, rtol, atol, equal_nan=False):
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
+def bincount(input, weights=None, minlength=0):
+    """Count number of occurrences of each value in array of non-negative ints.
+    Equivalent of ``numpy.bitcount(input, weights, minlength)``
+
+    Args:
+        input (array_like): Input array.
+        weights (array_like, optional): Weights, array of the same shape as input.
+                                        Defaults to None.
+        minlength (int, optional): A minimum number of bins for the output array.
+                                        Defaults to 0.
+    """
+    if isinstance(input, np.ndarray):
+        return np.bincount(input, weights=weights, minlength=minlength)
+    elif isinstance(input, TorchTensor):
+        return torch.bincount(input, weights=weights, minlength=minlength)
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
 def norm(array, axis=None):
     """Compute the 2-norm (Frobenius norm for matrices) of the input array.
 
@@ -145,5 +164,46 @@ def vstack(arrays):
     elif isinstance(arrays[0], TorchTensor):
         _check_all_same_type(arrays, TorchTensor)
         return torch.vstack(arrays)
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def index_add(output_array, input_array, index):
+    """Accumulates in `output_array`
+    the elements of `array`
+    by adding to the indices in the order given in index.
+
+    it is equivalent of torch's:
+
+    output_array.index_add_(0, torch.tensor(index),input_array)
+
+    """
+    if len(index.shape) != 1:
+        raise ValueError("index should be 1D array")
+    if isinstance(input_array, np.ndarray):
+        _check_all_same_type([output_array], np.ndarray)
+        np.add.at(output_array, index, input_array)
+    elif isinstance(input_array, TorchTensor):
+        _check_all_same_type([output_array], TorchTensor)
+        output_array.index_add_(0, torch.tensor(index), input_array)
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def zeros_like(array, shape=None):
+    """Create an zeros_like with the same size of array.
+    if shape is not None it overrides the shape of the result.
+
+    It is equivalent of np.zeros_like(array, shape=shape).
+    requires_grad is used only in torch"""
+    if isinstance(array, np.ndarray):
+        return np.zeros_like(array, shape=shape)
+    elif isinstance(array, TorchTensor):
+        if shape is None:
+            shape = array.size()
+
+        return torch.zeros(
+            shape, dtype=array.dtype, layout=array.layout, device=array.device
+        )
     else:
         raise TypeError(UNKNOWN_ARRAY_TYPE)
--- a/python/src/equistore/operations/reduce_over_samples.py
+++ b/python/src/equistore/operations/reduce_over_samples.py
@@ -107,9 +107,9 @@ def _reduce_over_samples_block(
 
 def sum_over_samples(tensor: TensorMap, sample_names: List[str]) -> TensorMap:
     """Create a new :py:class:`TensorMap` with the same keys as
-    as the input `tensor`, and each :py:class:`TensorBlock` is obtained summing
-    the corresponding input :py:class:`TensorBlock` over the rows with the same
-    sample_names.
+    as the input ``tensor``, and each :py:class:`TensorBlock` is obtained
+    summing the corresponding input :py:class:`TensorBlock` over the rows with
+    the same ``sample_names``.
 
     For example if ``sample_name = ["structure"]``, the function sums over all
     the sample with the same structure, and if ``sample_name = ["structure",
@@ -141,9 +141,9 @@ def sum_over_samples(tensor: TensorMap, sample_names: List[str]) -> TensorMap:
 
 def mean_over_samples(tensor: TensorMap, sample_names: List[str]) -> TensorMap:
     """Create a new :py:class:`TensorMap` with the same keys as
-    as the input `tensor`, and each :py:class:`TensorBlock` is obtained
+    as the input ``tensor``, and each :py:class:`TensorBlock` is obtained
     averaging the corresponding input :py:class:`TensorBlock` over the rows with
-    the same sample_names.
+    the same ``sample_names``.
 
     For example if ``sample_name = ["structure"]``, the function averages over
     all the sample with the same structure, and if ``sample_name = ["structure",
--- a/python/src/equistore/operations/reduce_over_samples.py
+++ b/python/src/equistore/operations/reduce_over_samples.py
@@ -38,17 +38,17 @@ def _reduce_over_samples_block(
         block_values, shape=(new_samples.shape[0],) + other_shape
     )
 
+    _dispatch.index_add(
+        values_result,
+        block_values,
+        index,
+    )
+
     if reduction == "mean":
         bincount = _dispatch.bincount(index)
-
-        _dispatch.index_add(
-            values_result,
-            block_values[:, ...]
-            / bincount[index].reshape((-1,) + (1,) * len(other_shape)),
-            index,
+        values_result = values_result / bincount.reshape(
+            (-1,) + (1,) * len(other_shape)
         )
-    elif reduction == "sum":
-        _dispatch.index_add(values_result, block_values, index)
 
     result_block = TensorBlock(
         values=values_result,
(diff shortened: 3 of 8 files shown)


python/src/equistore/operations/_dispatch.py
```diff
@@ -129,3 +129,15 @@ def vstack(arrays):
         return torch.vstack(arrays)
     else:
         raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def index_add(array, shape0, index):
```
Luthaf, line 134: This function feels like it is doing the job of two functions: `zero_like` and then `index_add`. How about splitting it in two?

python/src/equistore/operations/_dispatch.py
```diff
@@ -129,3 +129,15 @@ def vstack(arrays):
...
+    if isinstance(array, np.ndarray):
+        result = np.zeros((shape0,) + array.shape[1:], dtype=np.float64)
+        np.add.at(result, index, array)
+    elif isinstance(array, TorchTensor):
+        result = torch.zeros((shape0,) + array.shape[1:], dtype=torch.float64)
```
Luthaf, line 139: This should also allow for different dtypes/devices

python/src/equistore/operations/reduce.py
```diff
@@ -0,0 +1,99 @@
...
+    :param sample_name: names of samples to sum
+    """
+
+    try:
+        blocks = []
+        for key, block in tensor:
+            blocks.append(_reduction_sample_block(block=block, sample_name=sample_name))
+        return TensorMap(tensor.keys, blocks)
+    except ValueError:
+        print("it is not possible to perform the reduction for all the blocks")
```
Luthaf, line 99: Printing when you caught an exception is not a good practise: the user will likely miss the error and can not do anything about it. Why not just let the exception go all the way to the user here?

python/tests/operations/reduce.py
```diff
@@ -0,0 +1,298 @@
...
+        self.assertTrue(
+            np.all(
+                np.sum(bl1.values[26:32], axis=0) == reduce_tensor_ps.block(0).values[6]
+            )
+        )
+        self.assertTrue(
+            np.all(
+                np.sum(bl1.values[32:38], axis=0) == reduce_tensor_ps.block(0).values[7]
+            )
+        )
```
Luthaf, line 67: I would not test everything one by one, it makes the test harder to read and modify later if we need to. How about testing the first new sample and one in the middle only?

python/src/equistore/operations/reduce.py
```diff
@@ -0,0 +1,99 @@
...
+from . import _dispatch
+
+# TODO maybe add a way to reduce only the rows
+#  with sample[sample_name] = idx
+# for example not summing over all the sample
+# that have the same 'structure'
+# but only those with 'structure' = 0
+
+
+def _reduction_sample_block(block: TensorBlock, sample_name: list) -> TensorBlock:
```
Luthaf, line 15:
```suggestion
def _reduction_sample_block(block: TensorBlock, sample_names: list) -> TensorBlock:
```

since there are multiple names

Luthaf: also, the right type should be

```suggestion
def _reduction_sample_block(block: TensorBlock, sample_name: List[str]) -> TensorBlock:
```

python/src/equistore/operations/reduce.py
```diff
@@ -0,0 +1,99 @@
...
+    """Create a new :py:class:`TensorBlock` summing the `properties` among
+    the selected `samples`.
+    The output :py:class:`TensorBlocks` have the same components of the input one.
+    :param block: :py:class:`TensorBlock` -> input block
+    :param sample_name: list -> names of samples to sum
```
Luthaf, line 20: you do not need to put the types here again, sphinx should be able to extract them from the signature

python/src/equistore/operations/reduce.py
```diff
@@ -0,0 +1,99 @@
...
+    """
+    for isample in sample_name:
```
Luthaf, line 22: `isample` is a weird name for me: I expect stuff with a `i` in the name to be integers. Why not
```suggestion
    for sample_name in sample_names:
```

or 
```suggestion
    for sample in sample_names:
```

python/src/equistore/operations/reduce.py
```diff
@@ -0,0 +1,99 @@
...
+
+    for parameter, gradient in block.gradients():
+        # here I copy because I want to modify the samples array
+        samples = (
+            gradient.samples.view(dtype=np.int32)
+            .reshape(gradient.samples.shape[0], -1)
+            .copy()
+        )
+        for i in range(len(samples)):
+            samples[i, 0] = index[samples[i, 0]]
```
Luthaf, line 56: is there a magical numpy function that would do this without a for loop?

python/src/equistore/operations/reduce.py
DavideTisi: this function does not show up in the doc. only resolve that acts on TensorMap does. I can put also this public

Luthaf: No, I think this is fine being private. My comment was more about making docstrings consistent in the whole codebase

python/tests/operations/reduce.py
DavideTisi: i left few of them at the beginning, in the middle and at the end

python/src/equistore/operations/reduce.py
DavideTisi: it think samples[:, 0] = index[samples[:, 0]] works

python/src/equistore/operations/_dispatch.py
```diff
@@ -129,3 +129,49 @@ def vstack(arrays):
...
+    it is equivalent of torch's:
+
+    output_array.index_add_(0, torch.tensor(index),input_array)
+
+    """
+    if len(index.shape) != 1:
+        raise ValueError("index should be 1D array")
+    if isinstance(input_array, np.ndarray):
+        _check_all_same_type([output_array], np.ndarray)
+        np.add.at(output_array, index, input_array)
```
Luthaf, line 148: This should check that index is a 1D array

python/src/equistore/operations/_dispatch.py
```diff
@@ -129,3 +129,49 @@ def vstack(arrays):
...
+    elif isinstance(input_array, TorchTensor):
+        _check_all_same_type([output_array], TorchTensor)
+        output_array.index_add_(0, torch.tensor(index), input_array)
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def zeros_like(array, shape=None, dtype=float, requires_grad=False):
```
Luthaf, line 156:
```suggestion
def zeros_like(array, shape=None):
```

python/src/equistore/operations/_dispatch.py
```diff
@@ -129,3 +129,49 @@ def vstack(arrays):
...
+    """Create an zeros_like with the same size of array.
+    if shape is not None it overrides the shape of the result.
+
+    It is equivalent of np.zeros_like(array, dtype=dtype, shape=shape).
+    requires_grad is used only in torch"""
+    if isinstance(array, np.ndarray):
+        result = np.zeros_like(array, dtype=dtype, shape=shape)
```
Luthaf, line 163:
```suggestion
        return np.zeros_like(array, dtype=dtype, shape=shape)
```

python/src/equistore/operations/_dispatch.py
```diff
@@ -129,3 +129,49 @@ def vstack(arrays):
...
+    elif isinstance(array, TorchTensor):
+        if shape is None:
+            shape = array.size()
+
+        result = torch.zeros(
```
Luthaf, line 168:
```suggestion
        return torch.zeros(
```

python/src/equistore/operations/_dispatch.py
```diff
@@ -129,3 +129,49 @@ def vstack(arrays):
...
+            shape,
+            dtype=array.dtype,
+            layout=array.layout,
+            device=array.device,
+            requires_grad=requires_grad,
+        )
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+    return result
```
Luthaf, line 177:
```suggestion
```

python/src/equistore/operations/reduce.py
```diff
@@ -0,0 +1,106 @@
...
+from typing import List
+
+# TODO maybe add a way to reduce only the rows
+#  with sample[sample_name] = idx
+# for example not summing over all the sample
+# that have the same 'structure'
+# but only those with 'structure' = 0
+
+
+def _reduction_sample_block(block: TensorBlock, sample_name: List[str]) -> TensorBlock:
```
Luthaf, line 16:
```suggestion
def _reduction_sample_block(block: TensorBlock, sample_names: List[str]) -> TensorBlock:
```

python/src/equistore/operations/reduce.py
```diff
@@ -0,0 +1,106 @@
...
+
+    For example if sample_name = ["structure"] it sum over
+    all the sample with the same structure
+    if sample_name = ["structure", "center"] it sums over all the rows with
+    the same values for the "structure" and the "center"
+
+    :param tensor:  input :py:class:`TensorMap`
+    :param sample_name: names of samples to sum
+    """
+
```
Luthaf, line 97: We could also check the sample names here using `tensor.sample_names`, and make the check in `_reduction_sample_block` an assertion.

python/src/equistore/operations/sum_samples.py
```diff
@@ -0,0 +1,94 @@
...
+
+    :param tensor:  input :py:class:`TensorMap`
+    :param sample_names: names of samples to sum
+    """
+
+    blocks = []
+    for key, block in tensor:
+        for sample in sample_names:
+            if sample not in block.samples.names:
+                raise ValueError(f"TensorBlock with key={key} has not sample={sample}")
```
Luthaf, line 92: Since the sample names are the same for all blocks, you can move this check outside of the block loop

python/tests/operations/sum_samples.py
```diff
@@ -0,0 +1,229 @@
...
+import equistore.io
+import numpy as np
+import equistore.operations as fn
+from equistore import Labels, TensorBlock, TensorMap
+
+
+DATA_ROOT = os.path.join(os.path.dirname(__file__), "..", "data")
+
+
+class TestSum_samples(unittest.TestCase):
```
Luthaf, line 13:
```suggestion
class TestSumSamples(unittest.TestCase):
```

docs/src/reference/python/operations/sum_samples.rst
```diff
@@ -0,0 +1,4 @@
+lstsq
+======
```
Luthaf, line 2:
```suggestion
sum_samples
===========
```

python/src/equistore/operations/sum_samples.py
```diff
@@ -0,0 +1,97 @@
...
+    if sample_name = ["structure", "center"] it sums over all the rows with
+    the same values for the "structure" and the "center"
+
+    :param tensor:  input :py:class:`TensorMap`
+    :param sample_names: names of samples to sum
+    """
+    # I check that the names in sample_names are indeed presents in the block sample
+    # I check only the first one because all the block have the same sample names
+    for sample in sample_names:
+        if sample not in tensor[0].samples.names:
```
Luthaf, line 92:
```suggestion
    # We check that the names in sample_names are indeed presents in the block sample
    for sample in sample_names:
        if sample not in tensor.sample_names:
```

python/src/equistore/operations/sum_samples.py
```diff
@@ -0,0 +1,97 @@
...
+def sum_samples(tensor: TensorMap, sample_names: List[str]) -> TensorMap:
+    """Create a new :py:class:`TensorMap` with the same keys as
+    as the input `tensor`, and each :py:class:`TensorBlock`
+    is obtained summing the corresponding input :py:class:`TensorBlock`
+    over the rows with the same sample_names.
+
+    For example if sample_name = ["structure"], the function sums over
+    all the sample with the same structure
+    if sample_name = ["structure", "center"] it sums over all the rows with
+    the same values for the "structure" and the "center"
```
Luthaf, line 84:
```suggestion
    For example if ``sample_name = ["structure"]``, the function sums over all
    the sample with the same structure, and if ``sample_name = ["structure",
    "center"]`` it sums over all the rows with the same values for both
    ``"structure"`` and ``"center"`` samples.
```

python/src/equistore/operations/sum_samples.py
```diff
@@ -0,0 +1,97 @@
...
+    The output :py:class:`TensorBlocks` have the same components of the input one.
+    :param block: -> input block
+    :param sample_names: -> names of samples to sum
+    """
+    for sample in sample_names:
+        assert (
+            sample in block.samples.names
+        ), "the values in sample_names should be contained in block.samples.names"
+
+    # get the indeces of the selected sample
```
Luthaf, line 22:
```suggestion
    # get the indices of the selected sample
```

Luthaf (commented):
I think this is good to go. Do you want to add the option to use "mean" instead of  "sum" now or leave for later?

docs/src/reference/python/operations/sum_samples.rst
DavideTisi: I was thinking of doing now, I thought it would not take much but I couldn't think an immediate way to change torch._index _add for the computation of mean. 
for numpy I think you can simply do np.mean.at

DavideTisi:
> I think this is good to go. Do you want to add the option to use "mean" instead of "sum" now or leave for later?

I was thinking of doing now, I thought it would not take much but I couldn't think an immediate way to change torch._index _add for the computation of mean. 
for numpy I think you can simply do np.mean.at

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -0,0 +1,168 @@
...
+    sample_selected = [block.samples.names.index(sample) for sample in sample_names]
+    # reshaping the samples in a 2D array
+    samples = block.samples.view(dtype=np.int32).reshape(block.samples.shape[0], -1)
+    s, index = np.unique(samples[:, sample_selected], return_inverse=True, axis=0)
+    values_result = _dispatch.zeros_like(
+        block.values, shape=(s.shape[0],) + block.values.shape[1:]
+    )
+
+    if reduction == "mean":
+        bincount = _dispatch.bincount(index)
```
DavideTisi, line 43: to take care of the mean in a way that works both with numpy and torch I used this bitcount thing.
I do not know if it is the best but there are no python loop

docs/src/reference/python/operations/sum_over_samples.rst
```diff
@@ -0,0 +1,4 @@
+sum_over_samples
+================
+
+.. autofunction:: equistore.operations.reduce_over_samples.sum_over_samples
```
Luthaf, line 4:
```suggestion
.. autofunction:: equistore.operations.sum_over_samples
```

python/src/equistore/operations/reduce_over_samples.py
Luthaf: that looks good!

Luthaf:
I rebased & force-pushed two small cleanup commit. Let me know if everything works for you & then we can merge this!

PR #79, opened 2022-11-24
Add join functions

### LINKED ISSUE #62: Join function for two `TensorMap`s
With LODE I have intensive calculations to perform. This requires pre-computing and saving the TensorMaps to quickly play around with some models based on the pre-computed values. To use as much as possible computer power I split my dataset into many chunks. After pre-computing these chunks, it would be nice to _join_ the individual TensorMaps into one for convenient model training.

However, AFAIK equistore is currently missing such a `join` function merging two TensorMaps into a new object. Below is a minimal example in Python of the user facing API.

```python
import ase.io
import equistore
from rascaline import SphericalExpansion

frames = ase.io.load("dataset.xyz")

calculator = SoapPowerSpectrum(
    cutoff=3.0,
    max_radial=6,
    max_angular=4,
    atomic_gaussian_width=0.3,
    center_atom_weight=1.0,
    radial_basis={"Gto": {},},
    cutoff_function={"ShiftedCosine": {"width": 0.5},}
)

descriptor_1 = calculator.compute(frames[:len(frames)//2])
descriptor_2 = calculator.compute(frames[len(frames)//2:])

# define a join function, that joins two TensorMaps
descriptor = equistore.join(descriptor_1, descriptor_2)

# Overload the `+` operator and allow
descriptor = descriptor_1 + descriptor_2
```

Pinging @DivyaSuman14  and @jwa7 since they were part of the discussion.

PicoCentauri:
Closes #62

This PR adds two new operations. First `join_along_samples` similar to `vstack`  and second `join_along_properties` similar to `hstack`.

# TODOs
- [x] fix`join_along_samples` with gradients
- [x] Adjust new property and sample names
- [x] Add tests
- [x] Update docs

### DIFF
--- a/python/src/equistore/operations/_utils.py
+++ b/python/src/equistore/operations/_utils.py
@@ -1,3 +1,5 @@
+from typing import List
+
 import numpy as np
 
 from ..block import TensorBlock
@@ -12,11 +14,49 @@ def _check_same_keys(a: TensorMap, b: TensorMap, fname: str):
         raise ValueError(f"inputs to {fname} should have the same keys")
 
 
-def _check_same_gradients(a: TensorBlock, b: TensorBlock, fname: str):
-    grad_a = a.gradients_list()
-    grad_b = b.gradients_list()
+def _check_blocks(a: TensorBlock, b: TensorBlock, props: List[str], fname: str):
+    """Check if a proprty is the same between two :py:class:`TensorBlock`s.
 
-    if len(grad_a) != len(grad_b) or (
-        not np.all([parameter in grad_b for parameter in grad_a])
-    ):
-        raise ValueError(f"inputs to {fname} should have the same gradients")
+    :param a: first :py:class:`TensorBlock` for check
+    :param b: second :py:class:`TensorBlock` for check
+    :param props: A list of strings containing the property to check.
+                 Allowed values are ``'properties'`` or ``'samples'``,
+                 ``'components'`` and ``'gradients'``.
+    """
+    for prop in props:
+        err_msg = f"inputs to {fname} should have the same {prop}"
+        if prop == "samples":
+            if not np.all(a.samples == b.samples):
+                raise ValueError(err_msg)
+        elif prop == "properties":
+            if not np.all(a.properties == b.properties):
+                raise ValueError(err_msg)
+        elif prop == "components":
+            if not np.all(a.components == b.components):
+                raise ValueError(err_msg)
+        elif prop == "gradients":
+            grads_a = a.gradients_list()
+            grads_b = b.gradients_list()
+
+            if len(grads_a) != len(grads_b) or (
+                not np.all([parameter in grads_b for parameter in grads_a])
+            ):
+                raise ValueError(err_msg)
+        else:
+            raise ValueError(
+                f"{prop} is not a valid property to check. "
+                "Choose from 'samples', 'properties', 'components' "
+                "or 'gradients'."
+            )
+
+
+def _check_same_gradients_components(a: TensorBlock, b: TensorBlock, fname: str):
+    for parameter, grad_a in a.gradients():
+        grad_b = b.gradient(parameter)
+        grad_comps_a = np.array(grad_a.components, dtype=object)
+        grad_comps_b = np.array(grad_b.components, dtype=object)
+
+        if not np.all(grad_comps_a == grad_comps_b):
+            raise ValueError(
+                f"input to {fname} should habe the same gradient components"
+            )
--- a/python/src/equistore/operations/allclose.py
+++ b/python/src/equistore/operations/allclose.py
@@ -3,7 +3,7 @@ import numpy as np
 from equistore import TensorBlock, TensorMap
 
 from . import _dispatch
-from ._utils import _check_same_gradients, _check_same_keys
+from ._utils import _check_blocks, _check_same_keys
 
 
 def allclose(
@@ -186,8 +186,7 @@ def allclose_block_raise(
             if not np.all(c1 == c2):
                 raise ValueError("components not the same or not in the same order")
 
-    if len(block1.gradients_list()) > 0:
-        _check_same_gradients(block1, block2, "allclose")
+    _check_blocks(block1, block2, ["gradients"], "allclose")
 
     for parameter, gradient1 in block1.gradients():
         gradient2 = block2.gradient(parameter)
--- a/python/src/equistore/operations/join.py
+++ b/python/src/equistore/operations/join.py
@@ -1,0 +1,213 @@
+from typing import List
+
+import numpy as np
+
+from ..block import TensorBlock
+from ..labels import Labels
+from ..tensor import TensorMap
+from . import _dispatch
+from ._utils import _check_blocks, _check_same_gradients_components, _check_same_keys
+
+
+def join(tensormaps: List[TensorMap], axis: str):
+    """Join a sequence of :py:class:`TensorMap` along an axis.
+
+    The ``axis`` parameter specifies the type join.
+    For example, if ``axis='properties'`` it will be the joined along the
+    properties of the TensorMaps and if ``axis='samples'`` it will be the
+    along the samples.
+
+    :param tensormaps: sequence of :py:class:`TensorMap` for join
+    :param axis: A string indicating how the tensormaps are stacked.
+                 Allowed values are ``'properties'`` or ``'samples'``.
+
+    :return: The stacked :py:class:`TensorMap` which has more properties or
+             samples than the input TensorMap.
+    """
+
+    if len(tensormaps) < 2 or type(tensormaps) not in [list, tuple]:
+        raise ValueError("provide at least two `TensorMap`s for joining")
+
+    for ts_to_join in tensormaps[1:]:
+        _check_same_keys(tensormaps[0], ts_to_join, "join")
+
+    blocks = []
+    for key in tensormaps[0].keys:
+        blocks_to_join = [ts.block(key) for ts in tensormaps]
+
+        if axis == "properties":
+            blocks.append(_join_blocks_along_properties(blocks_to_join))
+        elif axis == "samples":
+            blocks.append(_join_blocks_along_samples(blocks_to_join))
+        else:
+            raise ValueError(
+                "Only `'properties'` or `'samples'` are "
+                "valid values for the `axis` parameter."
+            )
+
+    return TensorMap(tensormaps[0].keys, blocks)
+
+
+def _join_labels(labels: List[Labels]) -> Labels:
+    """Join a sequence of :py:class:`Labels`
+
+    Possible name and value clashes are resolved by the following strategy:
+
+    1. If the names are the same and the values are unique we keep the names and
+       join the values vertically. In example::
+
+        labels_1 = ["n"], [0, 2, 3]
+        labels_2 = ["n"], [1, 4, 5]
+
+      will lead to::
+
+        properties = ["n"], [0, 2, 3, 1, 4, 5]
+
+    2. If the names are the same but the values are not unique a new variable
+       "tensor" is added to the names. In example::
+
+            properties_1 = ["n"], [0, 2, 3]
+            properties_2 = ["n"], [0, 2]
+
+        will lead to::
+            properties = ["tensor", "n"], [[0, 0], [0, 2], [0, 3], [1, 0], [1, 2]]
+
+    3. If names of the Labels are different we change the names two to
+       ("tensor", "property"). In example::
+
+            properties_1 = ["a"], [0, 2, 3]
+            properties_2 = ["b", "c], [[0, 0], [1, 2]]
+
+        will lead to::
+
+            properties = ["tensor", "q"], [[0, 0], [0, 1], [0, 2], [1, 0], [1, 1]]
+    """
+    names = [label.names for label in labels]
+    values = [label.tolist() for label in labels]
+    tensor_values = np.repeat(
+        a=np.arange(len(values)), repeats=[len(value) for value in values]
+    )
+
+    unique_names = np.unique(names)
+    unique_values = np.unique(np.vstack(values), axis=0)
+
+    names_are_same = np.all(
+        len(unique_names) == np.array([len(name) for name in names])
+    )
+    values_are_unique = len(unique_values) == len(np.vstack(values))
+
+    if names_are_same and values_are_unique:
+        new_names = names[0]
+        new_values = np.vstack(values)
+    elif names_are_same and not values_are_unique:
+        new_names = ["tensor"] + list(names[0])
+        new_values = np.hstack([tensor_values.reshape(-1, 1), np.vstack(values)])
+    else:
+        new_names = ["tensor", "property"]
+        property_values = np.hstack([np.arange(len(value)) for value in values])
+        new_values = np.vstack([tensor_values, property_values]).T
+
+    return Labels(names=new_names, values=new_values)
+
+
+def _join_blocks_along_properties(blocks: List[TensorBlock]) -> TensorBlock:
+    """Join a sequence of :py:class:`TensorBlock` along their properties."""
+    first_block = blocks[0]
+
+    fname = "_join_blocks_along_properties"
+    for block in blocks[1:]:
+        _check_blocks(first_block, block, ["components", "samples", "gradients"], fname)
+        _check_same_gradients_components(first_block, block, fname)
+
+    properties = _join_labels([block.properties for block in blocks])
+    result_block = TensorBlock(
+        values=_dispatch.hstack([block.values for block in blocks]),
+        samples=first_block.samples,
+        components=first_block.components,
+        properties=properties,
+    )
+
+    for parameter, first_gradient in first_block.gradients():
+
+        # Different blocks might contain different gradient samples
+        gradient_sample_values = np.unique(
+            np.hstack([block.gradient(parameter).samples for block in blocks])
+        ).tolist()
+        gradient_samples = Labels(
+            names=first_gradient.samples.names, values=np.array(gradient_sample_values)
+        )
+
+        gradient_data = np.zeros(
+            (len(gradient_samples),)
+            + first_gradient.data.shape[1:-1]
+            + (len(properties),)
+        )
+
+        properties_start = 0
+        for block in blocks:
+            gradient = block.gradient(parameter)
+
+            # find the correct sample position to put the data
+            for i_grad, sample_grad in enumerate(gradient.samples):
+                # TODO: once C-API is changed replace by:
+                # i_sample = gradient_samples.position(sample_grad)`
+                i_sample = np.where(gradient_samples == sample_grad)[0][0]
+                gradient_data[
+                    i_sample,
+                    ...,
+                    properties_start : properties_start + len(gradient.properties),
+                ] = gradient.data[i_grad]
+
+            properties_start += len(gradient.properties)
+
+        result_block.add_gradient(
+            parameter, gradient_data, gradient_samples, first_gradient.components
+        )
+
+    return result_block
+
+
+def _join_blocks_along_samples(blocks: List[TensorBlock]) -> TensorBlock:
+    """Join a sequence of :py:class:`TensorBlock` along their samples."""
+    first_block = blocks[0]
+
+    fname = "_join_blocks_along_samples"
+    for block in blocks[1:]:
+        _check_blocks(
+            first_block, block, ["components", "properties", "gradients"], fname
+        )
+        _check_same_gradients_components(first_block, block, fname)
+
+    result_block = TensorBlock(
+        values=_dispatch.vstack([block.values for block in blocks]),
+        samples=_join_labels([block.samples for block in blocks]),
+        components=first_block.components,
+        properties=first_block.properties,
+    )
+
+    for parameter, first_gradient in first_block.gradients():
+        gradient_data = [first_gradient.data]
+        gradient_samples = [np.array(first_gradient.samples.tolist())]
+
+        for block in blocks[1:]:
+            gradient = block.gradient(parameter)
+            gradient_data.append(gradient.data)
+
+            samples = np.array(gradient.samples.tolist())
+            # update the "sample" variable in gradient samples
+            samples[:, 0] += 1 + gradient_samples[-1][:, 0].max()
+            gradient_samples.append(samples)
+
+        gradient_data = np.vstack(gradient_data)
+        gradient_samples = np.vstack(gradient_samples)
+
+        gradients_samples = Labels(first_gradient.samples.names, gradient_samples)
+
+        result_block.add_gradient(
+            parameter,
+            gradient_data,
+            gradients_samples,
+            first_gradient.components,
+        )
+
+    return result_block
--- a/python/src/equistore/operations/lstsq.py
+++ b/python/src/equistore/operations/lstsq.py
@@ -5,7 +5,7 @@ import numpy as np
 from ..block import TensorBlock
 from ..tensor import TensorMap
 from . import _dispatch
-from ._utils import _check_same_gradients, _check_same_keys
+from ._utils import _check_blocks, _check_same_keys
 
 
 def lstsq(X: TensorMap, Y: TensorMap, rcond, driver=None) -> TensorMap:
@@ -75,16 +75,15 @@ def _lstsq_block(X: TensorBlock, Y: TensorBlock, rcond, driver) -> TensorBlock:
     Y_n_properties = Y.values.shape[-1]
     Y_values = Y.values.reshape(-1, Y_n_properties)
 
-    if len(X.gradients_list()) > 0:
-        _check_same_gradients(X, Y, "lstsq")
+    _check_blocks(X, Y, ["gradients"], "lstsq")
 
-        for parameter, X_gradient in X.gradients():
-            X_gradient_data = X_gradient.data.reshape(-1, X_n_properties)
-            X_values = _dispatch.vstack((X_values, X_gradient_data))
+    for parameter, X_gradient in X.gradients():
+        X_gradient_data = X_gradient.data.reshape(-1, X_n_properties)
+        X_values = _dispatch.vstack((X_values, X_gradient_data))
 
-            Y_gradient = Y.gradient(parameter)
-            Y_gradient_data = Y_gradient.data.reshape(-1, Y_n_properties)
-            Y_values = _dispatch.vstack((Y_values, Y_gradient_data))
+        Y_gradient = Y.gradient(parameter)
+        Y_gradient_data = Y_gradient.data.reshape(-1, Y_n_properties)
+        Y_values = _dispatch.vstack((Y_values, Y_gradient_data))
 
     weights = _dispatch.lstsq(X_values, Y_values, rcond=rcond, driver=driver)
 
--- a/python/src/equistore/operations/solve.py
+++ b/python/src/equistore/operations/solve.py
@@ -3,7 +3,7 @@ import numpy as np
 from ..block import TensorBlock
 from ..tensor import TensorMap
 from . import _dispatch
-from ._utils import _check_same_gradients, _check_same_keys
+from ._utils import _check_blocks, _check_same_keys
 
 
 def solve(X: TensorMap, Y: TensorMap) -> TensorMap:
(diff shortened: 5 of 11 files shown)


ceriottm:
how a bout having a single `join([list of tensorblocks], axis)` where axis can be `"properties", "samples"` etc? 
makes it similar to how concatenate works in numpy

Luthaf:
The join should take tensormap, not tensor blocks; but apart from that I'm fine with either API.

PicoCentauri:
Also fine. More important should we do it on the rust side?

Luthaf:
> More important should we do it on the rust side?

This specific function would be possible to implement on the rust side, but I'm not sure it is worth it. It will not bring more performance, since most of the cost should be in copying the arrays around. And it will increase the API surface that needs to be made available everywhere.

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,149 @@
...
+            values=np.arange(n_samples_gradient, dtype=np.int32).reshape(-1, 1),
+        )
+
+        # TODO: calls to add_gradient are failing...
+        result_block.add_gradient(
+            parameter,
+            gradient_data,
+            gradients_samples,
+            gradient1.components,
+        )
```
PicoCentauri, line 147: I am a bit puzzled here. This leads to

```
EquistoreError: invalid parameter: first variable in the gradients samples labels must be 'samples'
```

but I constructed the gradient sample labels in the same way as the other labels.

Luthaf: the names in the gradient samples must be `["sample", "something", "else"]`, e.g. positions gradients use `["sample", "structure", "atom"]`, and cell gradients use `["sample"]`. 

Here you are passing `["gradients_samples"]` as the names. You need instead to update the gradient samples (some of them will have to change, because they refer to the indices of the samples in the initial blocks)

---

BTW, the error message is misleading, it should be `sample` not `samples`, and it could include what the user gave for more context.

PicoCentauri: Ahh I see. So in principle we should update all sample and property names, right?

Luthaf: when joining along samples; you should not have to update sample names, but the joined tensormap should have the same sample names already. You will have to update the gradient sample **values**, and ensure the values samples are distinct.

---

When joining along properties, you can either try to keep the metdata from the previous properties (if they have the same names & are distinct, take the union) or re-create new properties with new names. Maybe something like 

```
properties_1 = ["n"], [0, 2, 3]   |=> properties = ["n"], [0, 2, 3, 1, 4, 5]
properties_2 = ["n"], [1, 4, 5]   |

properties_1 = ["n"], [0, 2, 3]   |=> properties = ["tensor", "n"], [[0, 0], [0, 2], [0, 3], [1, 0], [1, 2]]
properties_2 = ["n"], [0, 2]      |

properties_1 = ["a"], [0, 2, 3]             |=> properties = ["tensor", "q"], [[0, 0], [0, 1], [0, 2], [1, 0], [1, 1]]
properties_2 = ["b", "c], [[0, 0], [1, 2]]  |
```

in the last case, we loose some information, which could be preserved if we add arbitrary string metadata to tensormaps (https://github.com/lab-cosmo/equistore/blob/fa1db6410da7a1eba995e98e6b30843905bdeae6/src/tensor/mod.rs#L30)

PicoCentauri: I changed the behavior of setting the samples and properties. Let me know if you agree that the logic makes sense @Luthaf. If yes I will deal with the open todo and add tests.

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,202 @@
...
+                    "Only `'properties'` or `'samples'` are "
+                    "valid values for the `axis` parameter."
+                )
+
+        stacked = TensorMap(stacked.keys, blocks)
+
+    return stacked
+
+
+def _join_along_properties_block(
```
Luthaf, line 53:
```suggestion
def _join_blocks_along_properties(
```

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,202 @@
...
+            ]
+
+        result_block.add_gradient(
+            parameter, gradient_data, gradient1.samples, gradient1.components
+        )
+
+    return result_block
+
+
+def _join_along_samples_block(block1: TensorBlock, block2: TensorBlock) -> TensorBlock:
```
Luthaf, line 149:
```suggestion
def _join_blocks_along_samples(block1: TensorBlock, block2: TensorBlock) -> TensorBlock:
```

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,202 @@
...
+    for property1, property2 in zip(block1.properties, block2.properties):
+        if tuple(property1) != tuple(property2):
+            raise ValueError("`A` and `B` do not have the same properties.")
+
+    if len(block1.components) > 0:
+        raise ValueError("`A` should not have any components")
+
+    if len(block2.components) > 0:
+        raise ValueError("`B` should not have any components")
```
Luthaf, line 158: same as above, this should be fine if both blocks have the same components

python/src/equistore/operations/join.py
Luthaf: I'm not sure I can follow the logic without comments or running it on some examples to be honest =). I'll check this branch out locally

PicoCentauri: Yes it is a bit hard so see. While I am writing the tests it seems that there are still some things I have to sort out.

PicoCentauri: Okay now it should work and you can try some examples. I will do the tests tomorrow.

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,212 @@
...
+                    "Only `'properties'` or `'samples'` are "
+                    "valid values for the `axis` parameter."
+                )
+
+        stacked = TensorMap(stacked.keys, blocks)
+
+    return stacked
+
+
+def _join_labels(Labels1: Labels, Labels2: Labels):
```
Luthaf, line 53:
```suggestion
def _join_labels(labels_1: Labels, labels_2: Labels):
```

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,212 @@
...
+    """Join two of :py:class:`Labels`
+
+    Possible name and value clasehd are resolve by the following strategy:
```
Luthaf, line 56:
```suggestion
    Possible name and value clashes are resolve by the following strategy:
```

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,212 @@
...
+
+    1. If the names are the same and the values are unqiue we simply join them.
```
Luthaf, line 58:
```suggestion
    1. If the names are the same and the values are uniques we simply join them.
```

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,212 @@
...
+       In example::
+
+        properties_1 = ["n"], [0, 2, 3]   |=> properties = ["n"], [0, 2, 3, 1, 4, 5]
+        properties_2 = ["n"], [1, 4, 5]   |
+
+    2. If the names are the same but the values are not unique a new property
```
Luthaf, line 64: most of the code uses the name `variable` for this

```suggestion
    2. If the names are the same but the values are not unique a new variable
```

python/src/equistore/operations/join.py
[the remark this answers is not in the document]
Luthaf: so this function makes sense when joining `properties`where the names don't carry too much meaning, but this is not the case when joining samples/keys. For samples one should only be able use the case 1 below, and for properties either case 1/2 would make sense.

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,212 @@
...
+    properties = _join_labels(block1.properties, block2.properties)
+    result_block = TensorBlock(
+        values=_dispatch.hstack([block1.values, block2.values]),
+        samples=block2.samples,
+        components=[],
+        properties=properties,
+    )
+
+    for parameter, gradient1 in block1.gradients():
+        gradient2 = block2.gradient(parameter)
```
Luthaf, line 160: This does not handle the case where gradient samples for the two blocks differ

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,212 @@
...
+        n_samples_gradient = len(gradient1.samples) + len(gradient2.samples)
+        gradient_data = np.zeros((n_samples_gradient,) + gradient1.data.shape[1:])
+
+        gradient_data[: len(gradient1.samples)] = gradient1.data
+        gradient_data[len(gradient1.samples) :] = gradient2.data
+
+        result_block.add_gradient(
+            parameter,
+            gradient_data,
+            _join_labels(gradient1.samples, gradient2.samples),
```
Luthaf, line 208: This is not the right way to join gradient samples. They should be concatenated vertically, and then the `sample` column should be updated to reflect the changes in the joined **values** samples.

PicoCentauri: True I change it back to the previous version.

PicoCentauri:
Sorry for the delay @Luthaf. If you could take another look that would be nice!

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,236 @@
...
+        result_block.add_gradient(
+            parameter, gradient_data, gradient1.samples, gradient1.components
+        )
+
+    return result_block
+
+
+def _join_blocks_along_samples(block1: TensorBlock, block2: TensorBlock) -> TensorBlock:
+    for property1, property2 in zip(block1.properties, block2.properties):
+        if tuple(property1) != tuple(property2):
```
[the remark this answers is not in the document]
Luthaf: Here the test can pass with different properties (if they have different sizes, but match for the all the common ones)

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,236 @@
...
+    if block1.components != block2.components:
+        raise ValueError("`A` and `B` do not have the same components")
+
+    if sorted(block1.gradients_list()) != sorted(block2.gradients_list()):
+        raise ValueError("`A` and `B` do not have the same gradients")
+
+    result_block = TensorBlock(
+        values=_dispatch.vstack([block1.values, block2.values]),
+        samples=_join_labels(block1.samples, block2.samples),
+        components=[],
```
Luthaf, line 208:
```suggestion
        components=block1.components,
```

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,236 @@
...
+        raise ValueError("`A` and `B` do not have the same components")
+
+    if sorted(block1.gradients_list()) != sorted(block2.gradients_list()):
+        raise ValueError("`A` and `B` do not have the same gradients")
+
+    properties = _join_labels(block1.properties, block2.properties)
+    result_block = TensorBlock(
+        values=_dispatch.hstack([block1.values, block2.values]),
+        samples=block2.samples,
+        components=[],
```
Luthaf, line 168:
```suggestion
        components=block1.components,
```

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,236 @@
...
+    result_block = TensorBlock(
+        values=_dispatch.vstack([block1.values, block2.values]),
+        samples=_join_labels(block1.samples, block2.samples),
+        components=[],
+        properties=block1.properties,
+    )
+
+    for parameter, gradient1 in block1.gradients():
+        gradient2 = block2.gradient(parameter)
+
```
Luthaf, line 214: this should also check that the gradients components are the same (they should, but nothing requires that a gradient wrt "positions" always has a given set of components.)

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,236 @@
...
+    result_block = TensorBlock(
+        values=_dispatch.hstack([block1.values, block2.values]),
+        samples=block2.samples,
+        components=[],
+        properties=properties,
+    )
+
+    for parameter, gradient1 in block1.gradients():
+        gradient2 = block2.gradient(parameter)
+
```
Luthaf, line 174: this should also check that the gradients components are the same

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,236 @@
...
+        properties = ["n"], [0, 2, 3, 1, 4, 5]
+
+    2. If the names are the same but the values are not unique a new variable
+       "tensor" as added to the names. In example::
+
+            properties_1 = ["n"], [0, 2, 3]
+            properties_2 = ["n"], [0, 2]
+
+        will lead to::
+            properties = ["tensor", "n"], [[0, 0], [0, 2], [0, 3], [1, 0], [1, 2]]
```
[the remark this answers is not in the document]
PicoCentauri: I thought about this as well and it is indeed problematic. However, a helper function will not solve the problem if the use stacks them in 2 calls.

```python
labels = _join_labels([labels_1, labels_2, sample_3, labels_4, ...])
```

should be the same as 

```python
labels_tmp = _join_labels([labels_1, labels_2])
labels = _join_labels([labels_tmp, sample_3, sample_4, ...])
```

for the latter one we would still end up with the same problem.

python/src/equistore/operations/join.py
PicoCentauri: Yes!

python/src/equistore/operations/join.py
PicoCentauri: same as above

python/src/equistore/operations/join.py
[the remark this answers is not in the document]
Luthaf: Yes, if the users calls the join manually multiple time, there is not much we can do =/. I still think it is worth trying to do something nice if the user gives us everything at once, and document that calling join multiple time will result on worst metadata.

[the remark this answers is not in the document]
PicoCentauri: Okay, I agree. But then I would change all function to handle a list and not only two elements. With this we are consistent within.

python/src/equistore/operations/join.py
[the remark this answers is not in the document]
PicoCentauri: But this would require an additional logic step filter only the common ones?

python/tests/operations/join.py
```diff
@@ -0,0 +1,165 @@
...
+import numpy as np
+from numpy.testing import assert_equal
+
+from equistore import Labels
+from equistore.io import load
+from equistore.operations.allclose import allclose_block_raise
+from equistore.operations.join import (  # noqa
+    _join_blocks_along_properties,
+    _join_blocks_along_samples,
+    _join_labels,
```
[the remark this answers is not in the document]
PicoCentauri: Actually, I am asking why these are private, I did this because we are handling them privately in the other `operations` but equistore has 3 major objects: `TensorMap`, `Labels`, `TensorBlock` which are visible to the user. Why should we not provide public APIs for all of these three instead of only one for the `TensorMap `

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,236 @@
...
+        gradient2 = block2.gradient(parameter)
+
+        gradient_data = np.zeros(gradient1.data.shape[:-1] + (len(properties),))
+
+        # copy block1 gradient data over
+        gradient_data[..., : len(block1.properties)] = gradient1.data
+
+        # check where to put the block2 data
+        for i_grad, sample_grad in enumerate(gradient2.samples):
+            i_merged = gradient1.samples.position(sample_grad)
```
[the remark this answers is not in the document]
PicoCentauri: Hmm, I thought at least for exactly the  soap + lode case it was working. But, maybe only for `join(lode, soap)` and not `join(soap, lode)`...

python/tests/operations/join.py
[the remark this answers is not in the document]
Luthaf: In my mind, the operation module is only about `TensorMap`. This is the most common use case and allow us to minimize the API surface we need to carefully think about. For example, there could be other way to join labels that do not make sense in the context of joining tensor maps, so making _join_label private means that we don't have to implement these other ways.

I'm not fundamentally against also having functionalities to manipulate blocks & labels, but this will be quite a bit more work that I would rather leave to separate PR and discussions

python/src/equistore/operations/join.py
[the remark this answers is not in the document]
PicoCentauri: Also this joins along properties so samples should be same for both gradient samples are the same.

python/src/equistore/operations/join.py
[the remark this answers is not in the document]
Luthaf: With the current code, zip will stop after yielding n elements, where n=min(len(block1.properties), len(block2.properties)), meaning the test as written could pass even if the two blocks have different properties

python/tests/operations/join.py
```diff
@@ -0,0 +1,165 @@
...
+        block_joined = _join_blocks_along_samples(block1, block2)
+        allclose_block_raise(block_joined, self.block)
+
+    def test_join_samples_different_properties(self):
+        """Test error raise if `proprties` are not the same in the two blocks."""
+        block1 = slice_block(self.block, properties=self.block.properties[:1])
+
+        with self.assertRaises(ValueError):
+            _join_blocks_along_samples(self.block, block1)
+
```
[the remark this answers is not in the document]
PicoCentauri: But there are no gradient properties. I would only see that this make sense if we join along properties and check of the samples are the same.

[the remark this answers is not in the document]
Luthaf: I mean there is not test with gradient, even just for `test_join_samples`

python/src/equistore/operations/join.py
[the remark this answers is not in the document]
Luthaf: > Hmm, I thought at least for exactly the soap + lode case it was working. But, maybe only for join(lode, soap) and not join(soap, lode)...

yes, `join(lode, soap)` should be fine, since lode gradient samples are a superset of soap gradients

> Also this joins along properties so samples should be same for both gradients.

samples can be the same without gradient samples being the same. 

For example, the block for species_center=8, species_neighbor=1 for H2O with O-H1 distance larger than the cutoff. SOAP and LODE will have the same samples (O center), but LODE will have three gradient samples (w.r.t. the positions of O, H1, H2), while SOAP will only have two gradient samples (w.r.t. position of O and H2)

PicoCentauri:
@Luthaf I hope I adressed most of your comments on the main functions. If there is something else we can discuss it. Once we agree I will work on the test cases.

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,209 @@
...
+    if names_are_same and values_are_unqiue:
+        new_names = names[0]
+        new_values = np.vstack(values)
+    elif names_are_same and not values_are_unqiue:
+        new_names = ["tensor"] + list(names[0])
+        new_values = np.hstack([tensor_values.reshape(-1, 1), np.vstack(values)])
+    else:
+        new_names = ["tensor", "property"]
+        property_values = np.hstack([np.arange(len(value)) for value in values])
+        new_values = np.vstack([tensor_values, property_values]).T
```
Luthaf, line 107: I like the clarity of the logic here!

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,209 @@
...
+        a=np.arange(len(values)), repeats=[len(value) for value in values]
+    )
+
+    unique_names = np.unique(names)
+    unique_values = np.unique(np.vstack(values), axis=0)
+
+    names_are_same = np.all(
+        len(unique_names) == np.array([len(name) for name in names])
+    )
+    values_are_unqiue = len(unique_values) == len(np.vstack(values))
```
Luthaf, line 96:
```suggestion
    values_are_unique = len(unique_values) == len(np.vstack(values))
```

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,209 @@
...
+            names=gradient_ref.samples.names, values=np.array(gradient_sample_values)
+        )
+
+        gradient_data = np.zeros(
+            (len(gradient_samples),)
+            + gradient_ref.data.shape[1:-1]
+            + (len(properties),)
+        )
+
+        i_prop = 0
```
Luthaf, line 145:
```suggestion
        properties_start = 0
```

python/src/equistore/operations/join.py
```diff
@@ -0,0 +1,209 @@
...
+    for parameter, gradient_ref in block_ref.gradients():
+        gradient_data = [gradient_ref.data]
+        gradient_samples = [np.array(gradient_ref.samples.tolist())]
+
+        for block in blocks[1:]:
+            gradient = block.gradient(parameter)
+            gradient_data.append(gradient.data)
+
+            samples = np.array(gradient.samples.tolist())
+            samples += 1 + gradient_samples[-1][:, 0].max()
```
Luthaf, line 194:
```suggestion
            samples[:, 0] += 1 + gradient_samples[-1][:, 0].max()
```

I'm not sure why you don't need to specify which column you are adding 1+max to?

Luthaf, line 194:
```suggestion
            # update the "sample" variable in gradient samples
            samples += 1 + gradient_samples[-1][:, 0].max()
```

python/src/equistore/operations/lstsq.py
```diff
@@ -76,7 +76,7 @@ def _lstsq_block(X: TensorBlock, Y: TensorBlock, rcond, driver) -> TensorBlock:
     Y_values = Y.values.reshape(-1, Y_n_properties)
 
     if len(X.gradients_list()) > 0:
-        _check_same_gradients(X, Y, "lstsq")
+        _check_blocks(X, Y, ["gradients"], "lstsq")
```
Luthaf, line 79: same, the `if` might not be required

python/src/equistore/operations/solve.py
```diff
@@ -75,7 +75,7 @@ def _solve_block(X: TensorBlock, Y: TensorBlock) -> TensorBlock:
     Y_values = Y.values.reshape(-1, Y_n_properties)
 
     if len(X.gradients_list()) > 0:
-        _check_same_gradients(X, Y, "solve")
+        _check_blocks(X, Y, ["gradients"], "solve")
```
Luthaf, line 78: same, the if might not be required, and the loop below will do nothing if there are no gradients

Luthaf (commented):
I like this code much better, it is quite clear! I have a handful of naming suggestions

python/src/equistore/operations/_utils.py
```diff
@@ -12,11 +12,32 @@ def _check_same_keys(a: TensorMap, b: TensorMap, fname: str):
...
-    ):
-        raise ValueError(f"inputs to {fname} should have the same gradients")
+def _check_blocks(a: TensorBlock, b: TensorBlock, props: list, fname: str):
+    for prop in props:
+        err_msg = f"inputs to {fname} should have the same {prop}"
+        if prop == "samples" and not np.all(a.samples == b.samples):
+            raise ValueError(err_msg)
+        elif prop == "properties" and not np.all(a.properties == b.properties):
+            raise ValueError(err_msg)
+        elif prop == "components" and a.components != b.components:
```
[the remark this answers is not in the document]
PicoCentauri: Hmm only because the components are empty in our case. A `np.all` does not hurt in any case.

python/src/equistore/operations/allclose.py
```diff
@@ -187,7 +187,7 @@ def allclose_block_raise(
                 raise ValueError("components not the same or not in the same order")
 
     if len(block1.gradients_list()) > 0:
-        _check_same_gradients(block1, block2, "allclose")
+        _check_blocks(block1, block2, ["gradients"], "allclose")
```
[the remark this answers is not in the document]
PicoCentauri: yes true

python/src/equistore/operations/join.py
PicoCentauri: Indeed. Maybe numpy magic...

python/src/equistore/operations/join.py
PicoCentauri: Thanks. I tried my best that we still get this, if we don't look at the code for a while.

PicoCentauri:
Okay, I will work on the tests tomorrow :-)

python/src/equistore/operations/join.py
Luthaf: Did you try something with position gradients? With cell gradients it would work since there is a single gradient sample named "sample"

PicoCentauri: I tried position gradients and at least it did not throw an error. But I will do some additional tests today.

python/tests/operations/join.py
```diff
@@ -0,0 +1,272 @@
...
+        block_extra_grad.add_gradient(
+            "foo", gradient.data, gradient.samples, gradient.components
+        )
+        self.ps_first_block_extra_grad = TensorMap(keys_first_block, [block_extra_grad])
+
+    def test_single_tensormap(self):
+        """Test error raise if only one tensormap is provided."""
+        with self.assertRaises(ValueError) as err:
+            join(self.ps, axis="properties")
+        assert "provide at least two" in str(err.exception)
```
Luthaf, line 43:
```suggestion
        self.assertIn("provide at least two", str(err.exception))
```

unfortunately unittest does not use the default assert syntax

python/tests/operations/join.py
```diff
@@ -0,0 +1,272 @@
...
+    def test_join_properties(self):
+        """Test public join function with three tensormaps along `properties`.
+
+        We check for the values below."""
+
+        ps_joined = join([self.ps, self.ps, self.ps], axis="properties")
+
+        # test property names
+        names = self.ps.block(0).properties.names
+        assert ps_joined.block(0).properties.names == ("tensor",) + names
```
Luthaf, line 58:
```suggestion
        self.assertEqual(ps_joined.block(0).properties.names, ("tensor",) + names)
```

python/tests/operations/join.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # test property values
+        tensor_prop = np.unique(ps_joined.block(0).properties["tensor"])
+        assert set(tensor_prop) == set((0, 1, 2))
```
Luthaf, line 62:
```suggestion
        self.assertEqual(set(tensor_prop), set((0, 1, 2)))
```

python/tests/operations/join.py
```diff
@@ -0,0 +1,272 @@
...
+            join([self.ps, self.ps, self.ps], axis="foo")
+        assert "values for the `axis` parameter" in str(err.exception)
+
+    def test_join_properties_values(self):
+        """Test values for joining along `properties`."""
+        ts_1 = slice(self.ps, properties=self.first_block.properties[:1])
+        ts_2 = slice(self.ps, properties=self.first_block.properties[1:])
+
+        ts_joined = join([ts_1, ts_2], axis="properties")
+        allclose_raise(ts_joined, self.ps)
```
Luthaf, line 83:
```suggestion
        self.assertTrue(allclose(ts_joined, self.ps))
```

PicoCentauri:
Thanks for the comments. Now the error messages of `unittest` also might be helpful.

python/tests/operations/join.py
```diff
@@ -0,0 +1,275 @@
...
+        block_extra_grad.add_gradient(
+            "foo", gradient.data, gradient.samples, gradient.components
+        )
+        self.ps_first_block_extra_grad = TensorMap(keys_first_block, [block_extra_grad])
+
+    def test_single_tensormap(self):
+        """Test error raise if only one tensormap is provided."""
+        with self.assertRaises(ValueError) as err:
+            join(self.ps, axis="properties")
+        self.assertIn("provide at least two", str(err.exception))
```
Luthaf, line 42: why check only part of the error message? It is very long?

python/tests/operations/join.py
```diff
@@ -0,0 +1,275 @@
...
+
+        self.assertEqual(joined_labels.names, ("structure", "prop_1"))
+        self.assertTrue(
+            np.equal(
+                joined_labels.tolist(), np.vstack([np.arange(10), np.arange(10)]).T
+            ).all()
+        )
+
+    def test_different_names(self):
+        """Test Label joining using labens with different names."""
```
Luthaf, line 228:
```suggestion
        """Test Label joining using labels with different names."""
```

python/tests/operations/join.py
```diff
@@ -0,0 +1,275 @@
...
+        values = np.vstack([np.arange(5), np.arange(5)]).T
+        property_labels_1 = Labels(("structure", "prop_1"), values)
```
Luthaf, line 230:
```suggestion
        property_labels_1 = Labels(("structure", "prop_1"), -values)
```

this way we check that the new values for `"property"` are actually re-created

Luthaf (commented):
We still need some tests for the joined gradient values / joined gradient samples and it should be good to go!

python/tests/operations/join.py
PicoCentauri: No but I always check a part which representative. If you fix a typo or change the structure slightly we don't have to change the test and even with this part it is clear that the correct error is raised.

python/tests/operations/join.py
```diff
@@ -0,0 +1,275 @@
+import unittest
+from os import path
+
+import numpy as np
+
+from equistore import Labels, TensorBlock, TensorMap
+from equistore.io import load
+from equistore.operations.allclose import allclose
+from equistore.operations.join import join
+from equistore.operations.slice import slice
```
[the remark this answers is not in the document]
PicoCentauri: I like this more!

python/tests/operations/join.py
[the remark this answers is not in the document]
PicoCentauri: But there are gradients in the power spectrum which are split and joind. And even though `allclose` is checking also gradients or not?


But I can do a handwritten example for gradients.

PicoCentauri:
Gradients should be checked actually? But we discuss this in the comment.

python/tests/operations/join.py
Luthaf: I tend to prefer checking everything, but I can understand your point of view here. Let's leave it like this for now.

python/tests/operations/join.py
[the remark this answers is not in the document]
Luthaf: right, I forgot allclose also deals with gradients ^^ 

maybe add a `self.assertIn("positions", ts.gradients_list())` to make prevent us from unintentionally dropping part of that test in the future?

PicoCentauri:
Lets see if the tests go through. I squashed also the commits.

Luthaf:
I fixed some formatting, should be good to go now!

PR #82, opened 2022-11-30
Cleanup tests and fix a bug for slice operation

Luthaf:
The issue was that the new gradient samples where just picked out of the old ones, but they need to be updated to refer to the new samples. For example, a block with gradient samples `[0, 1, 1, 2, 2, 3]` where samples 1 and 3 where picked need to be updated to read `[0, 0, 1]`. This only applies to the first column (`name = "sample"`) of the  gradient samples.

### DIFF
--- a/python/tests/operations/slice.py
+++ b/python/tests/operations/slice.py
@@ -27,37 +27,37 @@ class TestSlice(unittest.TestCase):
 
     # TEST BLOCK 1: SLICING SAMPLES WITHOUT GRADIENTS
 
-    def test_slice_samples_tensorblock_no_gradients(self):
-        tensormap = equistore.io.load(
+    def test_slice_samples_block_no_gradients(self):
+        tensor = equistore.io.load(
             os.path.join(DATA_ROOT, TEST_FILE),
             use_numpy=True,
         )
         # Remove the gradients to simplify test
-        tensormap = fn.remove_gradients(tensormap)
+        tensor = fn.remove_gradients(tensor)
         # Define a single block to test
-        tensorblock = tensormap.block(0)
+        block = tensor.block(0)
         # Slice only 'structures' 2, 4, 6, 8
         structures_to_keep = np.arange(2, 10, 2).reshape(-1, 1)
         samples_to_slice = Labels(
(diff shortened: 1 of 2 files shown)


python/tests/operations/slice.py
```diff
@@ -1,589 +1,319 @@
...
-        6) Testing the type handling of the user-facing slice function.
-        7) Checking issuance of user warnings when empty blocks are
-            created.
-    """
-
-    # TEST BLOCK 1: SLICING SAMPLES WITHOUT GRADIENTS
-
-    def test_slice_samples_tensorblock_no_gradients(self):
-        tensormap = equistore.io.load(
+class TestSliceSamples(unittest.TestCase):
```
Luthaf, line 15: FYI, you can have multiple classes inheriting from `TestCase` in the same file, that's usually how I structure multiple test blocks in the same file

jwa7: Ok that is much cleaner, thanks!

jwa7 (commented):
All fine

python/tests/operations/slice.py
```diff
@@ -54,26 +54,26 @@ def _check_sliced_block(self, block, sliced_block, structures_to_keep):
             # no slicing of properties has occurred
             self.assertTrue(np.all(sliced_gradient.properties == gradient.properties))
 
-            # samples have been sliced to the correct dimension
+            # samples have been updated to refer to the new samples
```
jwa7, line 57: This only checks that the max sample index isn't greater than the number of samples in the parent. Should we also check that the mapping of samples indices in the Gradient is correct too?

Luthaf: I could not find a nice way to test it without re-implementing the function ... Maybe the best kind of test would be to check the gradients with finite differences? Should not be too hard to do, and we can re-use it for all operations

jwa7:
Commit 7abe577ab3d8b906e8ef463b5ce91495dc501324 all fine too. Thanks @Luthaf for finding the bug in re-indexing of Gradient samples, I missed that! I'm happy for you to squash your commits and then I'll merge :)

PR #83, opened 2022-12-05
Fix incorrect code example in docs

jwa7:
Minor changes to docs:

1. Add return type to the ``slice`` function
2. Change erroneous arg names ``samples`` and ``properties`` in docstring code examples of ``slice`` and ``slice_block`` fxns to ``samples_to_slice`` and ``properties_to_slice``.

### DIFF
--- a/python/src/equistore/operations/slice.py
+++ b/python/src/equistore/operations/slice.py
@@ -7,15 +7,15 @@ from ..labels import Labels
 from ..tensor import TensorMap
 
 
-def slice(tensor: TensorMap, samples_to_slice=None, properties_to_slice=None):
+def slice(tensor: TensorMap, samples=None, properties=None) -> TensorMap:
     """
     Slices an input :py:class:`TensorMap` along the samples and/or properties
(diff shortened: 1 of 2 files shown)


python/src/equistore/operations/slice.py
```diff
@@ -22,13 +24,13 @@ def slice(tensor: TensorMap, samples_to_slice=None, properties_to_slice=None):
 
         sliced_tensor = slice(
             tensor,
-            samples=Labels(
+            samples_to_slice=Labels(
```
Luthaf, line 27: should this really be named `samples_to_slice`? Given we are calling the `slice` function, this feels redundant :smiley:

jwa7: Good point! I can do the reverse, but it will be a breaking change (i.e. for the equistore examples thing that uses it)

Luthaf: I'm fine with breaking equistore-example, that's kind of the point =) (figuring out an API and changing it until it is great!)

jwa7:
Changes made and commits squashed

PR #85, opened 2022-12-08
Split equistore in a C-API core crate and a Rust wrapper

### LINKED ISSUE #102: equistore.__version__
Currently, we have a version info when installing equistore via pip etc. However, there is no `__version__` attribute when using equistore from the Python interpreter.

We should add this attribute. `__version__` helps a lot with debugging production code, especially when using rascaline together equistore which could lead to some weird version clash behavior.

A quick and dirty solution is the following. We are pulling the version from the rust code using 

https://github.com/lab-cosmo/equistore/blob/e65cd07d2f0d5aefa41a7b2beade835b9b1dc03b/setup.py#L91-L98

in `setup.py`. During this step we can write a `version.py` file to the source and import this in our `__init__.py`.

Another maybe cleaner  would be providing a version string from `rust` over `c` to `Python`.

### LINKED ISSUE #140: Reusing TensorBlock for creating a TensorMap causes segmentation fault
When I reuse the TensorBlock twice for creating a TensorMap (that happens quite easily when playing around with TensorMaps in a notebook, then my notebook crashes.
```python
# block_1 = some Tensor definition ...
TensorMap(Labels.single(), [block_1])
TensorMap(Labels.single(), [block_1]) <--- causes segmentation fault
```

I think it is unrelated to the TensorBlock definition, but here the whole script
```python
from equistore import TensorMap, TensorBlock, Labels
import numpy as np

block_1 = TensorBlock(
    values=np.random.rand(4, 2),
    samples=Labels(
        ["sample", "structure"], np.array([[0,0], [1,1], [2,2], [3,3]], dtype=np.int32)
    ),
    components=[],
    properties=Labels(["properties"], np.array([[0], [1]], dtype=np.int32)),
)
block_1.add_gradient(
    "positions",
    data=np.random.rand(7, 3, 2),
    samples=Labels(
        ["sample", "structure", "center"],
        np.array([[0, 0, 1], [0, 0, 2], [1, 1, 0], [1, 1, 1], [1, 1, 2], [2, 2, 0], [3, 3, 0]], dtype=np.int32)
    ),
    components=[Labels(["direction"], np.array([[0], [1], [2]], dtype=np.int32))],
)

TensorMap(Labels.single(), [block_1])
TensorMap(Labels.single(), [block_1])
```

Luthaf:
Closes #102 
Closes #140 

This is phase I of making equistore compatible with TorchScript, by having a separate C API we can load the code from a single shared library regardless if we are using it from Python/Torch/rascaline/…

- split equistore crate into a core, containing the C API and a Rust wrapper API
- the core no longer depends on the ndarray crate
    - loading files now takes a callback to create arrays in which the data will
      be placed
    - serialization is no longer optional, there is a single way the core can be
      built (it had multiple optional feature before)
- ~Python bindings explicitly manage the lifetime of array using `Py_IncRef` and
  `Py_DecRef`~ This is now #86 
    - ~we no longer have to explicitly keep references to arrays around, this
      simplifies the code quite a bit~

Still TODO:
- [x] update the CI for the new configuration
- [x] make sure all tests have been updated, and none is left commented out
- [x] integrate with rascaline
  - [x] check that the performance hit of not inlining equistore functions is not too bad
- [x] write API documentation
- [x] write something on the code organization for new developers
- [x] remove all TODO in the code
- [x] add soname to the shared library

----
:books: Documentation preview :books:: https://equistore--85.org.readthedocs.build/en/85/


docs/src/devdoc/architecture.rst
```diff
@@ -0,0 +1,80 @@
...
+  tests and examples
+
+Finally, ``docs/`` contains the documentation for everything related to
+equistore.
+
+------------------------
+
+.. figure:: ../../static/images/API-organization.*
+    :width: 600px
+    :align: center
```
PicoCentauri, line 23: Could you change the color of the arrows to gray like (0.5,0.5,0.5) in rgb. with this the diagramm also works in darkmode quite well.

Luthaf:
And merged!

PR #86, opened 2022-12-10
Make Tensormap immutable

### LINKED ISSUE #58: Segmentation fault when accessing TensorBlock that doesn't exist
This bug was discovered when working through the notebook `tutorial-equivariant-models.ipynb` in the lab-cosmo/equistore-examples repo. I have included some MWE code below to illustrate the bug. This code should be executed in the relative directory `equistore-examples/` as it depends on an example water dataset found in the `equistore-examples/data/` directory.

**Set Up**

The dataset is loaded, hypers defined, and descriptors computed. Then, we can access a block of desired angular momentum, species center, and species neighbour as follows:
```
import ase.io
import equistore
from rascaline import SphericalExpansion

frames = ase.io.read("data/chemrev_nuprime-theta-grid_computed.xyz", ":")

hypers = {
    "cutoff": 2.0,
    "max_radial": 6,
    "max_angular": 4,
    "atomic_gaussian_width": 0.2,
    "radial_basis": {"Gto": {}},
    "cutoff_function": {"ShiftedCosine": {"width": 0.5}},
    "center_atom_weight": 1.0,
}

calculator = SphericalExpansion(**hypers)
descriptor = calculator.compute(frames, gradients=['positions'])

# Define block with l=2, center=O, neighbour=H
block1 = descriptor.block(spherical_harmonics_l=2,species_center=8,species_neighbor=1)
```

If we want to move the `species_neighbour` index from keys to properties, we can do so with the `key_to_properties()` function:

```
# Move the index from keys to properties
descriptor.keys_to_properties('species_neighbor')
```

**Expected Behaviour**

Because `species_neighbour` is no longer a valid key, we cannot define a block as above with the specification of `species_neighbour`. Doing so throws an error, as expected and desired:
```
# Define block with l=2, center=O, and invalid parameter neighbour=H
block2 = descriptor.block(spherical_harmonics_l=2,species_center=8,species_neighbor=1)

>>> EquistoreError: invalid parameter: 'species_neighbor' is not part of the keys for this tensor
```

Instead, a block must be accessed only by specifying the remaining valid keys:
```
block2 = descriptor.block(spherical_harmonics_l=2,species_center=8)
```

**Unexpected Behaviour**

If a block is defined, i.e. `block1`, and then its index is moved from keys to properties (as shown above), the block previously assigned to the variable `block1` no longer exists. If attempts are made to access this block, instead of an error being thrown a Segmentation Fault occurs at runtime and, if executing the code in a notebook, this result in the kernel dying.
```
# Define block with l=2, center=O, neighbour=H
block1 = descriptor.block(spherical_harmonics_l=2,species_center=8,species_neighbor=1)

# Move the index from keys to properties
descriptor.keys_to_properties('species_neighbor')

# Try to access block
block1

# >>> Kernel dies due to segmentation fault
```

Ideally, such user behaviour should result in a well-reported error as opposed to a runtime fault.

Luthaf:
The first commit is extracted from #85, and required to make the second one work.

The second commit fixes #58 and is a API breaking change. It will make #85 easier to implement in full.

```py
# before
tensor = ...
tensor.keys_to_properties("species_neighbor")

# now
tensor = ...
tensor = tensor.keys_to_properties("species_neighbor")
```

TODO before merging:
- [x] update all examples

### DIFF
--- a/python/src/equistore/data/array.py
+++ b/python/src/equistore/data/array.py
@@ -61,7 +61,6 @@ class ArrayWrapper:
         self.array = array
         self._shape = ctypes.ARRAY(c_uintptr_t, len(array.shape))(*array.shape)
 
-        self._children = []
         if _is_numpy_array(array):
             array_origin = _origin_numpy()
         elif _is_torch_array(array):
@@ -69,10 +68,9 @@ class ArrayWrapper:
         else:
             raise ValueError(f"unknown array type: {type(array)}")
 
-        self.eqs_array = eqs_array_t()
-
-        # `eqs_data.eqs_array.ptr` is a pointer to the PyObject `self`
-        self.eqs_array.ptr = ctypes.cast(
+        eqs_array = eqs_array_t()
+        # `eqs_array_t::ptr` is a pointer to the PyObject `self`
+        eqs_array.ptr = ctypes.cast(
             ctypes.pointer(ctypes.py_object(self)), ctypes.c_void_p
         )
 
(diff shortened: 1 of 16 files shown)


python/src/equistore/tensor.py
```diff
@@ -371,9 +330,54 @@ def components_to_properties(self, variables: Union[str, List[str]]):
...
             self._ptr, c_variables, c_variables._length_
         )
+        return TensorMap._from_ptr(ptr)
+
+    def keys_to_properties(
+        self,
+        keys_to_move: Union[str, List[str], Labels],
+        *,
+        sort_samples=True,
+    ) -> "TensorMap":
```
jwa7, line 343: Why is this ``"TensorMap"`` and not just ``TensorMap``?

jwa7: Also, I was just wondering out of curiosity if there is a good reason why we are going for ``new_tm = tm.keys_to_properties(keys_to_move)`` over ``new_tm = keys_to_properties(tm, keys_to_move)``, i.e. why keep as a class method if it returns a new ``TensorMap`` object?

python/src/equistore/data/array.py
```diff
@@ -80,24 +78,39 @@ def eqs_array_origin(this, origin):
...
+        eqs_array.create = eqs_array.create.__class__(_eqs_array_create)
+        eqs_array.copy = eqs_array.copy.__class__(_eqs_array_copy)
+        eqs_array.destroy = eqs_array.destroy.__class__(_eqs_array_destroy)
 
-        self.eqs_array.move_samples_from = self.eqs_array.move_samples_from.__class__(
+        eqs_array.move_samples_from = eqs_array.move_samples_from.__class__(
             _eqs_array_move_samples_from
         )
 
+        self._eqs_array = eqs_array
```
jwa7, line 97: From line 70-97 it looks like a simple change of the order in which attributes are assigned. Before you were assigning the object ``eqs_array_t()`` to the ``ArrayWrapper.eqs_array`` attribute and then modifying it, but now you are creating the object, modifying it, then assigning it. It's quite possible I'm missing some deeper implications here - what is the reason for these changes?

python/src/equistore/tensor.py
Luthaf: > Why is this "TensorMap" and not just TensorMap?

I remember either mypy or the Python parser chocking if we use the current class as a return type, I'll have to check again. The string syntax has the same meaning.

> why keep as a class method if it returns a new TensorMap object?

because the function still acts on a TensorMap instance. This is similar to `array.contiguous()` or `array.sum()`, which returns new arrays.

python/src/equistore/data/array.py
Luthaf: no deep meaning here =) Assigning a fully constructed object reduces the rightward drift of the code by removing a bunch of `self.` =)

jwa7: Ok I see thanks!

python/src/equistore/tensor.py
jwa7: Ok makes sense :)

PR #69, opened 2022-11-10
First version of user documentation.

PicoCentauri:
Based on the work by @curiosity54 and @HowWeiBin this PR adds a first version of a user documentation. The structure is similar to rascalines but of course with different content.

If you review take a close look at the general structure since the content is in a rough state.

PR #87, opened 2023-01-04
Refactor our tox usage

Luthaf:
This was extracted out of #69, and should bring equistore closer to how rascaline uses tox.


.github/workflows/docs.yml
```diff
@@ -23,10 +23,11 @@ jobs:
           python-version: "3.10"
       - name: install dependencies
         run: |
-          pip install -r docs/requirements.txt
+          pip install tox
```
PicoCentauri, line 26:
```suggestion
          python -m pip install tox
```

setup.cfg
```diff
@@ -1,7 +1,7 @@
 [metadata]
```
PicoCentauri, line 1: I think we can even get rid of this `setup.cfg` and move everything inside the `pyproject.toml`. I can do this in a separate PR.

PicoCentauri (changes requested):
Thanks for translating the skeleton.

I am not sure why the test was failing. We might only need to rerun things.

PicoCentauri:
I quickly tried to address the comments by myself :-)

PicoCentauri (approved):

PR #88, opened 2023-01-05
Translate setup.cfg in pyproject.toml

PicoCentauri:
Moved all setup options from setup.cfg into pyproject.toml. This required to increase the build version for the python wheels from 3.8 to 3.10. However, I think this will not be a problem and now we consistently use Python 3.10 in all our CI jobs.


pyproject.toml
```diff
@@ -6,10 +6,24 @@ requires = [
...
+[project]
+name = "equistore"
+description = "A specialized data storage format suited to atomistic machine learning needs and more."
+readme = "README.rst"
+requires-python = ">=3.6"
+
+dependencies = [
+    "numpy>=1.20.0",
+]
+dynamic = ["version"]
```
Luthaf, line 18: what does this do?

.github/workflows/build-wheels.yml
```diff
@@ -98,7 +98,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.8
+          python-version: "3.10"
```
Luthaf, line 101: Do we have any test left that the code builds fine with Python 3.7 -> 3.9? Or should we add one?

Luthaf: Sorry, I missed the top comment. Does this also mean that I can not install from a local checkout with an older Python version? If so, this might be a bit early to do this, given that most users still install from source for now, and most HPC centers do not come with Python 3.10

pyproject.toml
PicoCentauri: This tells the build system the our version is dynamically set in the `setup.py`.

.github/workflows/build-wheels.yml
Luthaf: Yes, Python 3.7 is not EOL until June this year, and we might want to continue supporting it for a bit longer after that

PicoCentauri: Okay lets see what the new tests are saying. I am also fine if we keep this PR stale until we drop the versions.

PicoCentauri:
Okay as expected this needs a higher python version for the external build to work. So we leave this for now.

.github/workflows/build-wheels.yml
Luthaf: I would rather close than keep a stale PR, but we can open an issue to track this, pointing to the changes in this PR

PR #91, opened 2023-01-06
First pass for the user documentation

Luthaf:
This is contains the same changes as #69, squashed into a single commit.

---

Here is a list of remaining things to do:

- [x] home page: explain what equistore is and why one would want to use it instead of the different doc sections (moved from `get-started/equistore`)
- [ ] get-started/concepts: rewrite to not require previous knowledge of ACDC & friends, adding figures and other visual aids
- [x] get-started/installation: add some explanation and text around the commands

We also have to add how-to & friends, but It would be better of as separate PR to ensure we can merge this one quickly!


Luthaf:
I removed the figures from the git history, to prevent polluting it too much until we converge to the final version. Here are the files for reference.

<details>

collection-blocks

single-block

</details>

docs/src/tutorials/linear-model.rst
```diff
@@ -0,0 +1,4 @@
+Equivariant Linear Model
+------------------------
+
+TBD
```
PicoCentauri, line 4: Since I am currently doing a Ridge implementation in equisolve, I am asking myself if an example really belongs here or do we just link to equisolve on this page?

Luthaf: My idea was more to provide a simple example using `equistore.operations` directly. Then the doc should 100% link to equisolve as a more complete implementation of this!

PicoCentauri: Okay make sense.

PicoCentauri:
As agreed today with @curiosity54 and @HowWeiBin today we should try to keep this branch going. The user base is increasing and we need proper tutorials to clearly state what equistore can do and where to find things.

One example is that it may seems that equistore objects are immutable since python like operations (i.e. `+=` etc) are not possible (yet). However, we already have a lot operations, buy we have to emphasize them. Maybe even with an example in the `README.rst`

```python
from equistore import Tensormap
import equistore.operations as eqst

tensor = Tensormap(xxx)
tensor = eqst. add(tensor, 5)
```

PicoCentauri (changes requested):
I talked to @Luthaf yesterday about the the user documentation and we think it is already in a really good shape!

I like the tutorials. But, they require a bit more explanation of what is going on between the code cells. Also, please split the one big tutorial into a second at the point where you start training an actual model.

Luthaf:
There have been a lot of improvement on the documentation, but this PR still contains some other tutorials that should be integrated. Keeping this open for now, until we have to time to finish the tutorial work.

Luthaf:
After checking again, not much to be salvaged anymore

PR #92, opened 2023-01-06
naming scheme in reduction operations

DavideTisi:
in reduction operation changed the variable name from ``sample_names`` to ``group_by``.
Now there is a new private function ``_reduce_over_samples`` which is called by the functions ``sum_over_samples`` and ``mean_over_samples`` with a different value for the input variable reduction. ``_reduce_over_samples`` does the necessary inputs check and dispatch to ``_reduce_over_samples_block``

### DIFF
--- a/python/src/equistore/operations/reduce_over_samples.py
+++ b/python/src/equistore/operations/reduce_over_samples.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Union
 
 import numpy as np
 
@@ -8,23 +8,23 @@ from . import _dispatch
 
 
 def _reduce_over_samples_block(
-    block: TensorBlock, sample_names: List[str], reduction: str
+    block: TensorBlock, group_by: List[str], reduction: str
 ) -> TensorBlock:
     """Create a new :py:class:`TensorBlock` summing the ``properties`` among
     the selected ``samples``.
     The output :py:class:`TensorBlocks` have the same components of the input one.
     :param block: -> input block
-    :param sample_names: -> names of samples to sum
+    :param group_by: -> names of samples to sum
     """
 
     block_samples = block.samples
-    for sample in sample_names:
+    for sample in group_by:
         assert sample in block_samples.names
 
     assert reduction in ["sum", "mean"]
 
     # get the indices of the selected sample
-    sample_selected = [block_samples.names.index(sample) for sample in sample_names]
+    sample_selected = [block_samples.names.index(sample) for sample in group_by]
     # reshaping the samples in a 2D array
     samples = block_samples.view(dtype=np.int32).reshape(block_samples.shape[0], -1)
     # get which samples will still be there after reduction
@@ -53,7 +53,7 @@ def _reduce_over_samples_block(
     result_block = TensorBlock(
         values=values_result,
         samples=Labels(
-            sample_names,
+            group_by,
             new_samples,
         ),
         components=block.components,
@@ -101,22 +101,29 @@ def _reduce_over_samples_block(
     return result_block
 
 
-def sum_over_samples(tensor: TensorMap, sample_names: List[str]) -> TensorMap:
+def _reduce_over_samples(
+    tensor: TensorMap, group_by: Union[List[str], str], reduction: str
+) -> TensorMap:
     """Create a new :py:class:`TensorMap` with the same keys as
     as the input ``tensor``, and each :py:class:`TensorBlock` is obtained
     summing the corresponding input :py:class:`TensorBlock` over the rows with
-    the same ``sample_names``.
+    the same ``group_by`` sample names.
 
-    For example if ``sample_name = ["structure"]``, the function sums over all
-    the sample with the same structure, and if ``sample_name = ["structure",
+    Both "sum" and "mean" reductions can be performed.
+
+    For example if ``group_by = "structure"``, the function reduces over all
+    the sample with the same structure, and if ``group_by = ["structure",
     "center"]`` it sums over all the rows with the same values for both
     ``"structure"`` and ``"center"`` samples.
 
     :param tensor: input :py:class:`TensorMap`
-    :param sample_names: names of samples to sum over
+    :param group_by: names of samples to reduce over
+    :param reduction: how to reduce, only available values are "mean" or "sum"
     """
-    # We check that the names in sample_names are indeed presents in the block
-    for sample in sample_names:
+    if isinstance(group_by, str):
+        group_by = [group_by]
+
+    for sample in group_by:
         if sample not in tensor.sample_names:
             raise ValueError(
                 f"one of the requested sample name ({sample}) is not part of "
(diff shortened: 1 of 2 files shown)


ceriottm:
I wonder if `gather_by` would be more telling? It's not grouping, it's reducing.
Would it be much more complicated to specify what's it's summing over / averaging over? The behavior is exactly the same as np.sum(axis=xxx), only that the axes are sparse.

DavideTisi:
Not really sure I understand what you mean. The parameter ``group_by`` (or what ever will be its name) is telling you, reducing (sum or mean) over all the lines that have the same values corresponding to the sample with name ``group_by``.
while ``np.sum(axis=xxx)`` does something slightly different if I am no mistaken. Taken ``A`` a matrix ``NxM`` with  ``np.sum(A,axis=0)`` I end up with a vector of dimension M with each element is the sum over all the N lines. 
our function does not do that Suppose now ``A`` is a tensorBlock the first dimension corresponds to the samples. The sample name are [“structure“, "atom"] and you use
``group_by`` = [“structure“]. The first N1 samples you have “structure“=0, the other N-N1 “structure“=1, the final result of sum_by_samples will be an array of dimension (2,N) equivalent of the staking of ``np.sum(A[:N1],axis=0)`` and ``np.sum(A[N1:],axis=0)``.

ceriottm:
What I mean is to turn around the logic, keeping `sample_names` as parameter and specifying what indices you sum over. Then it becomes similar to what `axis`  does. If you imagine that the `samples` labels specify sparse directions in an array, then the `group_by` names are the axes that "are left" after the summation. But this is not how one usually write reduction operations in numpy, or on paper. You write `A_i=\sum_{j1,j2} A_{ij1j2}`. I'm suggesting to specify j1,j2 rather than (as one currently does) i.

DavideTisi:
Sorry but I still do not get it. What you are suggesting it is not what the operations does or is suppose to do. 
In your example if `A` has 3 indeces 
`B_i=\sum_{j1,j2} A_{i j1 j2}` makes B an array with only one indeces summing over all the rest. What our operation does is 
`C_{i1,j1,j2}=\sum_{i if i=i1} A_{i j1 j2}`.
C still has 3 indeces.
The parameter `sample_names` basically gives only the condition that must be satisfied in the sum, but you are always summing over the first index of the actual array.

DavideTisi:
basically you are reducing on the “names“ of the samples, and with `sample_names` or `group_by` you are telling the result must only have these sample_names, which are the only that i care about but the same number of features or components.

ceriottm:
I think there is some confusion in what you write. 

image

The result of the operation is exactly what I had written above - the indices indicated in sample_names are kept, and the others are summed over. I think if you say `array.sum` it is more intuitive to specify which indices you _sum over_  than those that are kept.

Luthaf:
I agree it makes sense to inverse the meaning, and specify what you want to sum over. The main use case for specifying what you want to keep instead would be some code that tries to be able to process arbitrary samples, but we can always built it from the new operation. 

In this case, I would not name the parameter `group_by` anymore, but maybe something like `samples`? Do you have other ideas?

DavideTisi:
Hi all, 
sorry I didn't write but in last commit, I have already inverted the logic of the input variable. 
For the naming I do not have any strong feelings, now it is still `group_by` just because I left the one I used, but I agree that we need to change it now.
`samples` or `sample_names` is ok for me, but I have no ideas. As soon we have decided the name I finish the PR

ceriottm:
I think that with the reverse logic `sample_names` works well. Alternatively, if we want to emphasize the link with numpy.sum, `sample_axis` would also be a good name.

Luthaf:
I think `sample_axis` will be more confusing, since these are not directly axes of the data array (even if they work in a similar fashion to sparse indexes for axes). I'd rather do `samples_names` or `samples`, I tend to prefer the simpler `samples`, but I'm fine with either.

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -101,69 +101,94 @@ def _reduce_over_samples_block(
...
-def sum_over_samples(tensor: TensorMap, sample_names: List[str]) -> TensorMap:
+def _reduce_over_samples(
+    tensor: TensorMap, samples_names: Union[List[str], str], reduction: str
+) -> TensorMap:
     """Create a new :py:class:`TensorMap` with the same keys as
     as the input ``tensor``, and each :py:class:`TensorBlock` is obtained
-    summing the corresponding input :py:class:`TensorBlock` over the rows with
-    the same ``sample_names``.
+    summing the corresponding input :py:class:`TensorBlock` over the ``samples_names``
+    indeces.
```
Luthaf, line 110:
```suggestion
    indices.
```

Here and everywhere else =)

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -101,69 +101,94 @@ def _reduce_over_samples_block(
...
-    ``"structure"`` and ``"center"`` samples.
+    Both "sum" and "mean" reductions can be performed.
+
+    ``samples_names`` tells over which indeces of the sparse matrix the reduction
+    is performed. It accept both a single string or a list of the string
+    with the sample_names corresponding to the directions along which the
+    reduction is performed.
+    Moreover if only one sample name is given both a single string or a list with
+    a string is allowed: ``samples_names = ["center"]`` or equivalently
+    ``samples_names = "center"``.
```
Luthaf, line 120: I would not duplicate the documentation here, since this is a private function, and I would rather have fewer places where we need to update documentation

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -8,23 +8,23 @@
...
 def _reduce_over_samples_block(
-    block: TensorBlock, sample_names: List[str], reduction: str
+    block: TensorBlock, samples_names: List[str], reduction: str
 ) -> TensorBlock:
     """Create a new :py:class:`TensorBlock` summing the ``properties`` among
     the selected ``samples``.
     The output :py:class:`TensorBlocks` have the same components of the input one.
     :param block: -> input block
-    :param sample_names: -> names of samples to sum
+    :param samples_names: -> names of samples to sum
```
Luthaf, line 17: this is missing doc for `reduction`

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -8,23 +8,23 @@
 
 
 def _reduce_over_samples_block(
-    block: TensorBlock, sample_names: List[str], reduction: str
+    block: TensorBlock, samples_names: List[str], reduction: str
```
Luthaf, line 11:
```suggestion
    block: TensorBlock, samples_names_to_reduce: List[str], reduction: str
```

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -8,23 +8,27 @@
 
 
 def _reduce_over_samples_block(
-    block: TensorBlock, sample_names: List[str], reduction: str
+    block: TensorBlock, samples_names_to_reduce: List[str], reduction: str
```
Luthaf, line 11:
```suggestion
    block: TensorBlock, remaining_samples: List[str], reduction: str
```

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -101,69 +105,84 @@ def _reduce_over_samples_block(
...
+        samples_names = [samples_names]
+
+    for sample in samples_names:
         if sample not in tensor.sample_names:
             raise ValueError(
                 f"one of the requested sample name ({sample}) is not part of "
                 "this TensorMap"
             )
 
+    samples_names_to_reduce = [
```
Luthaf, line 132:
```suggestion
    remaining_samples = [
```

PR #93, opened 2023-01-12
added the printing of shapes in the print function

DavideTisi:
following @ceriottm suggestion, now the print fuction prints also the shapes of the different direction of the tensorblock, basically the block.value.shape .

This first commit is a draft ( do not pass the test becouse i haven't modified them).

i think that there are different possible ways to show the shapes. The codes I have here does that:

```
>> tensor[1]
TensorBlock
     samples (len: 52): ['structure', 'center']
     components (len: [3]): ['spherical_harmonics_m']
     properties (len: 4): ['n']
     gradients: ['cell', 'positions']

>>tensor[1].gradient('positions')
Gradient TensorBlock 
 parameter: 'positions'
 samples (len: 128): ['sample', 'structure', 'atom']
 components (len: [3, 3]): ['direction', 'spherical_harmonics_m']
 properties (len: 4): ['n']
```

Now at each different dimension (sample, components, properties) it has in () the lenght of that dimension in the block.value array. The components is a list of all the lengths (if the are no components it is an empy array like the one after the :)

For the gradient it writes nothing in the TensorBlock (no idea what to write), but when you print one specific gradient it prints all the shapes, with the same spirit.

One other idea is just to add one line that prints the values.shape like:

```
>> tensor[1]
TensorBlock
     values.shape : (52, 3, 4)
     samples: ['structure', 'center']
     components: ['spherical_harmonics_m']
     properties: ['n']
     gradients: ['cell', 'positions']

>>tensor[1].gradient('positions')
Gradient TensorBlock 
 data.shape: (128, 3, 3, 4)
 parameter: 'positions'
 samples: ['sample', 'structure', 'atom']
 components: ['direction', 'spherical_harmonics_m']
 properties: ['n']
```
I think that putting the lenght next to each axis name is clearer. 

What do you think?

### DIFF
--- a/python/src/equistore/block.py
+++ b/python/src/equistore/block.py
@@ -95,19 +95,19 @@ class TensorBlock:
 
     def __repr__(self) -> str:
         s = "TensorBlock\n"
-        s += "    samples: ["
+        s += f"    samples (len: {len(self.samples)}): ["
         for sname in self.samples.names[:-1]:
             s += "'" + sname + "', "
         s += "'" + self.samples.names[-1] + "']"
         s += "\n"
-        s += "    component: ["
+        s += f"    components (len: {[len(c) for c in self.components]}): ["
         for ic in self.components:
             for name in ic.names[:]:
                 s += "'" + name + "', "
         if len(self.components) > 0:
             s = s[:-2]
         s += "]\n"
-        s += "    properties: ["
+        s += f"    properties (len: {len(self.properties)}): ["
         for name in self.properties.names[:-1]:
             s += "'" + name + "', "
         s += "'" + self.properties.names[-1] + "']\n"
@@ -272,19 +272,19 @@ class Gradient:
     def __repr__(self) -> str:
         s = "Gradient TensorBlock \n"
         s += "parameter: '{}'\n".format(self._name)
-        s += "samples: ["
+        s += f"samples (len: {len(self.samples)}): ["
         for sname in self.samples.names[:-1]:
             s += "'" + sname + "', "
         s += "'" + self.samples.names[-1] + "']"
         s += "\n"
-        s += "component: ["
+        s += f"components (len: {[len(c) for c in self.components]}): ["
         for ic in self.components:
             for name in ic.names[:]:
                 s += "'" + name + "', "
         if len(self.components) > 0:
             s = s[:-2]
         s += "]\n"
-        s += "properties: ["
+        s += f"properties (len: {len(self.properties)}): ["
         for name in self.properties.names[:-1]:
             s += "'" + name + "', "
         s += "'" + self.properties.names[-1] + "']"
--- a/python/src/equistore/block.py
+++ b/python/src/equistore/block.py
@@ -101,7 +101,7 @@ class TensorBlock:
         s += "'" + self.samples.names[-1] + "']"
         s += "\n"
         s += "    components ("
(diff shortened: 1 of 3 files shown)


DavideTisi:
I agree that = is better now the output is:

```
>> tensor[1]
TensorBlock
     samples (len = 52): ['structure', 'center']
     components (len = [3]): ['spherical_harmonics_m']
     properties (len = 4): ['n']
     gradients: ['cell', 'positions']

>> tensor[1].gradient('positions')
Gradient TensorBlock 
 parameter: 'positions'
 samples (len = 128): ['sample', 'structure', 'atom']
 components (len = [3, 3]): ['direction', 'spherical_harmonics_m']
 properties (len = 4): ['n']
```

ceriottm:
Thanks Davide. No strong feelings either but given that components can have
a shape, and not just a single scalar, how about we just use a tuple
notation, e.g.

```

TensorBlock
     samples (52): ['structure', 'center']
     components (3,3): ['spherical_harmonics_m', 'another_axis']
     properties (4): ['n']
     gradients: ['cell', 'positions']

```

DavideTisi:
Yes I like it, I think it is still clear even without the explicit len=
Only note now the code produce something exactly like @ceriottm said:

```
TensorBlock
     samples (52): ['structure', 'center']
     components (3,3): ['spherical_harmonics_m', 'another_axis']
     properties (4): ['n']
     gradients: ['cell', 'positions']
``` 


Which makes the thing in () not technically a tuple since tuple with only one element would end with a "," giving something like:

```
TensorBlock
     samples (52,): ['structure', 'center']
     components (3,3): ['spherical_harmonics_m', 'another_axis']
     properties (4,): ['n']
     gradients: ['cell', 'positions']
``` 
which honestly I do not like, but it is the exact output you would get if those were tuples.
I do not thing it is necessary to stick with the exact tuple format since it is just a visualization thing

python/src/equistore/block.py
```diff
@@ -95,19 +95,21 @@ def copy(self) -> "TensorBlock":
...
         s = "TensorBlock\n"
-        s += "    samples: ["
+        s += f"    samples ({len(self.samples)}): ["
         for sname in self.samples.names[:-1]:
             s += "'" + sname + "', "
         s += "'" + self.samples.names[-1] + "']"
         s += "\n"
-        s += "    component: ["
+        s += "    components ("
+        s += f"{[len(c) for c in self.components]}"[1:-1]
```
Luthaf, line 104:
```suggestion
        s += ", ".join([str(len(c)) for c in self.components])
```

python/src/equistore/block.py
```diff
@@ -270,21 +272,23 @@ def __init__(self, block: TensorBlock, name: str):
...
         s += "parameter: '{}'\n".format(self._name)
-        s += "samples: ["
+        s += f"samples ({len(self.samples)}): ["
         for sname in self.samples.names[:-1]:
             s += "'" + sname + "', "
         s += "'" + self.samples.names[-1] + "']"
         s += "\n"
-        s += "component: ["
+        s += "components ("
+        s += f"{[len(c) for c in self.components]}"[1:-1]
```
Luthaf, line 283:
```suggestion
        s += ", ".join([str(len(c)) for c in self.components])
```

Luthaf (approved):

PR #95, opened 2023-01-19
Add LICENSE and AUTHORS files

PicoCentauri:
As the other projects of our ecosystem (chemiscope, chemfiles) equistore should also use the BSD-3 license. This PRs adds the corresponding LICENSE file.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,28 @@
+BSD 3-Clause License
+
+Copyright (c) 2023, equistore developers
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its
+   contributors may be used to endorse or promote products derived from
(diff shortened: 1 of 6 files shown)


LICENSE
```diff
@@ -0,0 +1,28 @@
+BSD 3-Clause License
+
+Copyright (c) 2023, COSMO
```
Luthaf, line 3:
```suggestion
Copyright (c) 2023, equistore developers
```

and we could add an AUTHORS file if we want (or just use git to track authorship)

PicoCentauri: Yes an AUTHOR file is a good idea.

setup.cfg
```diff
@@ -1,22 +1,40 @@
 [metadata]
 name = equistore
-description = "A specialized data storage format suited to atomistic machine learning needs and more."
-long_description = file: README.rst
+url = https://github.com/lab-cosmo/equistore
+description = A specialized data storage format suited to atomistic machine learning needs and more.
+long_description = file: README.rst, LICENSE.rst
```
Luthaf, line 5:
```suggestion
long_description = file: README.rst, LICENSE.rst
```

I've never seen the license included in the long description. There is no way to manually include the license file in a separate field?

PicoCentauri: Me neither. But I took a look at the setuptools docs and they suggest this.

https://setuptools.pypa.io/en/latest/userguide/declarative_config.html

I think we can go with the `license_files` attribute though.

Luthaf (approved):

PR #103, opened 2023-01-23
added add function

DavideTisi:
A first draft of a sum function, so far it has no test.
it sums a TensorMap + a constant value/array or TensorMap + TensorMap. 
In the first case, the metadata of the result is the one of th; input TensorMap, in the second the two input TensorMaps must have the same metadata that will be the same as the input.

### DIFF
--- a/docs/src/reference/python/operations/add.rst
+++ b/docs/src/reference/python/operations/add.rst
@@ -1,0 +1,4 @@
+add
+===
+
+.. autofunction:: equistore.operations.add
--- a/docs/src/reference/python/operations/multiply.rst
+++ b/docs/src/reference/python/operations/multiply.rst
@@ -1,0 +1,4 @@
+multiply
+========
+
+.. autofunction:: equistore.operations.multiply
--- a/python/src/equistore/operations/add.py
+++ b/python/src/equistore/operations/add.py
@@ -1,0 +1,89 @@
+from typing import Union
+
+from ..block import TensorBlock
+from ..tensor import TensorMap
+from ._utils import _check_blocks, _check_same_gradients_components, _check_same_keys
+
+
+def add(A: TensorMap, B: Union[float, TensorMap]) -> TensorMap:
+    """Return a new :class:`TensorMap` with the values being the sum of ``A`` and ``B``.
+
+    If ``B`` is a :py:class:`TensorMap` it has to have the same metadata as ``A``.
+
+    If gradients are present in ``A`` a sum is only performed if ``B`` is
+    a :py:class:`TensorMap` as well.
+
+    :param A: First :py:class:`TensorMap` for the addition.
+    :param B: Second instance for the addition. Parameter can be a scalar a array or a
+              :py:class:`TensorMap`. In the latter case ``B`` must have the same
+              metadata of ``A``.
+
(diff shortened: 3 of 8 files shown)


DavideTisi:
the sum with array should be not possible

python/src/equistore/operations/multiply.py
```diff
@@ -0,0 +1,91 @@
+from ..block import TensorBlock
+from ..tensor import TensorMap
+from ._utils import _check_blocks, _check_same_gradients_components, _check_same_keys
+
+
+def add(A: TensorMap, B) -> TensorMap:
```
PicoCentauri, line 6:
```suggestion
def multiply(A: TensorMap, B) -> TensorMap:
```

PicoCentauri (commented):
Just one minor comment and no real reveiw so far. Most importantly, I think we should also at least add the [magic functions](https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types) 
`__add__()` and `__mul__()`   for the `TensorMap` and `TensorBlock` class.

But, I can also do this in a upcoming PR...

DavideTisi:
I added the sum and multiplication for scalar and among `TensorMaps` with the same exact metadata. This last bit is not general but I would start merging this PR given the need for addition/multiplication with scalar. An to complete the general operations later.

PicoCentauri:
@DavideTisi I did some changes to the docstring and to the code to make a bit clearer. If you agree you can rebase/squash and merge.

PicoCentauri (approved):

docs/src/reference/python/operations/add.rst
```diff
@@ -0,0 +1,4 @@
+join
+====
```
Luthaf, line 2:
```suggestion
add
===
```

docs/src/reference/python/operations/multiply.rst
```diff
@@ -0,0 +1,4 @@
+join
+====
```
Luthaf, line 2:
```suggestion
multiply
========
```

python/src/equistore/operations/add.py
```diff
@@ -0,0 +1,93 @@
...
+            raise TypeError("B must be a TensorMap or a float.") from e
+        for blockA in A.blocks():
+            blocks.append(_add_block_constant(block=blockA, constant=B))
+
+    return TensorMap(A.keys, blocks)
+
+
+def _add_block_constant(block: TensorBlock, constant: float) -> TensorBlock:
+
+    # values = block1.values @ block2.values.T
```
Luthaf, line 50:
```suggestion
```

python/src/equistore/operations/multiply.py
```diff
@@ -0,0 +1,110 @@
...
+    """Return a new :class:`TensorMap` with the values being the element-wise
+    multiplication of ``A`` and ``B``.
+
+    If ``B`` is a :py:class:`TensorMap` it has to have the same metadata as ``A``.
+
+    If gradients are present in ``A`` a multiplication is only performed if ``B`` is
+    a :py:class:`TensorMap` as well. The multiplication is performed with the 
+    rule of the derivatives::
+
+      TODO
```
Luthaf, line 21: ?

PR #105, opened 2023-01-23
zeros and zeros_like

DavideTisi:
Add the function ``zeros`` and ``zeros_like``. 
The first takes the metadata as input and gives a ``TensorMap`` with values full of zeros.
The second takes a ``TensorMap``  and gives a ``TensorMap``  with the same metadata but the values matrices full zero.

Not sure how much these functions can be usefull

### DIFF
--- a/docs/src/reference/python/operations/index.rst
+++ b/docs/src/reference/python/operations/index.rst
@@ -13,10 +13,12 @@ Operations
 
     allclose() <allclose>
     dot() <dot>
+    empty_like() <empty_like>
     join() <join>
     lstsq() <lstsq>
+    ones_like() <ones_like>
     slice() <slice>
     solve() <solve>
-
     sum_over_samples() and mean_over_samples() <samples-reduction>
     remove_gradients() <remove-gradients>
+    zeros_like() <zeros_like>
--- a/python/src/equistore/operations/_dispatch.py
+++ b/python/src/equistore/operations/_dispatch.py
@@ -206,7 +206,20 @@ def index_add(output_array, input_array, index):
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
-def zeros_like(array, shape=None):
+def zeros(shape, array_type="numpy"):
+    """Create a `zeros_like` with the given input shape.
+
+    `array_type` can be `Numpy` or `torch`.
+    """
+    if array_type == "numpy":
+        return np.zeros(shape)
+    elif array_type == "torch":
+        return torch.zeros(shape)
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def zeros_like(array, shape=None, requires_grad=False):
     """Create an zeros_like with the same size of array.
     if shape is not None it overrides the shape of the result.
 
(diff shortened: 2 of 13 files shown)


PicoCentauri (commented):
Super nice @DavideTisi. This makes our lifes easier. I would add another option to the `zeros_like` function controlling which gradients should be "copied".

DavideTisi:
hello @PicoCentauri, I did what you asked but I have a doubt. All the blocks in a `TensorMap` must have the same gradient info? I do not remember now 
 
 If not, it is not that easy and you give as input a number of parameters equal the number of keys

DavideTisi:
> Why no use the already existing `remove_gradients` for this? i.e.
> 
> ```python
> tensor  = ... # has "positions" and "cell"
> 
> # current
> zero = zeros_like(tensor, parameters=["positions"])
> 
> # alternative
> zero = zeros_like(remove_gradients(tensor, ["cell"]))
> ```

I do not know I did it in very few lines of code, in this way you do not have to allocate memory and then deallocated. I know it is a very week point, my main point is doing even without ``remove_gradients`` took me few lines of code

python/src/equistore/operations/_dispatch.py
```diff
@@ -206,7 +206,19 @@ def index_add(output_array, input_array, index):
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
-def zeros_like(array, shape=None):
+def zeros(shape, array_type="numpy"):
+    """Create an zeros_like with the input shape.
+    array_type can be "numpy" or "torch".it create a numpy or torch array
+    """
```
PicoCentauri, line 212:
```suggestion
    """Create a `zeros_like` with the given input shape.
    
    `array_type` can be `Numpy` or `torch`.
    """
```

docs/src/reference/python/operations/index.rst
```diff
@@ -13,10 +13,14 @@ Operations
...
     join() <join>
     lstsq() <lstsq>
+    ones_like() <ones_like>
     slice() <slice>
     solve() <solve>
 
     sum_over_samples() and mean_over_samples() <samples-reduction>
     remove_gradients() <remove-gradients>
+
+    zeros_like() <zeros_like>
```
PicoCentauri, line 26:
```suggestion
    solve() <solve>
    sum_over_samples() and mean_over_samples() <samples-reduction>
    remove_gradients() <remove-gradients>
    zeros_like() <zeros_like>
```

python/src/equistore/operations/_dispatch.py
```diff
@@ -219,7 +231,57 @@ def zeros_like(array, shape=None):
...
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def ones_like(array, shape=None, requires_grad=False):
+    """Create an ones_like with the same size of array.
+    if shape is not None it overrides the shape of the result.
+
+    It is equivalent of np.ones_like(array, shape=shape).
+    requires_grad is used only in torch"""
```
PicoCentauri, line 249:
```suggestion
    """Create an ones_like with the same size of array.
   
    :param shape: If not ``None`` override the shape with the given one
    :param requires_grad: used only in torch
    """
```

python/src/equistore/operations/ones_like.py
```diff
@@ -0,0 +1,82 @@
...
+    :param tensor: input tensor from which you get the metadata.
+    :param parameters: which gradient parameter to copy. Default ``None``
+                      means copy all of them if present.
+                      Empty list for no gradients information.
+    :param requires_grad: If autograd should record operations on the returned tensor.
+                          Only for torch, default False.
+    """
+
+    blocks = []
+    for _k, block in tensor:
```
PicoCentauri, line 28: Since you don't need k can you not loop over the blocks directly?

```suggestion
    for block in tensor.blocks():
```

python/src/equistore/operations/ones_like.py
```diff
@@ -0,0 +1,82 @@
...
+        blocks.append(
+            ones_like_block(
+                block=block, parameters=parameters, requires_grad=requires_grad
+            )
+        )
+    return TensorMap(tensor.keys, blocks)
+
+
+def ones_like_block(
```
PicoCentauri, line 37: I think we should do this as a private function since we usually only provide operations for TensorMaps to the user.

PicoCentauri (changes requested):
Why do we want to provide a `zeros` function which takes `Labels` as parameters and `zeros_like` which takes a TensorMap as parameter but only a `ones_like` and a `empty_like` and no `ones` and `empty`?

DavideTisi:
> Why do we want to provide a `zeros` function which takes `Labels` as parameters and `zeros_like` which takes a TensorMap as parameter but only a `ones_like` and a `empty_like` and no `ones` and `empty`?

Because I was thinking of removing `zeros` and leave only the `*_like` function. The reason is that it you give to the function already all the metadata and the matrix, you are better off just to give them to the `TensorMap` creator with as values the `np.zeros()` with the size of your choice

PicoCentauri:
Okay, I am fine with this but at leasr we should try to be consistent.Sent from my phone, forgive brevity and typosOn 1 Feb 2023, at 21:30, Davide Tisi ***@***.***> wrote:﻿

Why do we want to provide a zeros function which takes Labels as parameters and zeros_like which takes a TensorMap as parameter but only a ones_like and a empty_like and no ones and empty?

Because I was thinking of removing zeros and leave only the *_like function. The reason is that it you give to the function already all the metadata and the matrix, you are better off just to give them to the TensorMap creator with as values the np.zeros() with the size of your choice

python/src/equistore/operations/ones_like.py
DavideTisi: I left it public because i thought that it might be useful if want for example add to some TensorMap a dummy Block with all ones (or whatever). Or he wants to build a TensorMap with a  block with zeros and one with ones, etc.
These are very corner and "stupid" cases but they might be useful. 
Of course I can put everything private it is preferred

python/src/equistore/operations/empty_like.py
```diff
@@ -0,0 +1,82 @@
...
+        gradient_data = _dispatch.empty_like(gradient.data)
+
+        result_block.add_gradient(
+            parameter,
+            gradient_data,
+            gradient.samples,
+            gradient.components,
+        )
+
+    return result_block
```
PicoCentauri, line 82: I like that we know have a couple of helper functions that make these functions quite clear.

python/src/equistore/operations/ones_like.py
PicoCentauri: We can also keep them.

PicoCentauri:
I am happy and rebased the branch. If you agree @DavideTisi and the CI is through you can merge.

DavideTisi:
for me it is ok but the CI still cannot compile some things in Rust, for some reason 😂😭

PR #106, opened 2023-01-25
RFC: Utility Functions for converting arrays

PicoCentauri:
While writing tests for the linear model I needed functions for creating equistore objects from numpy arrays. I distilled two 
functions 

1. `matrix_to_block`, taking a **2d** numpy array and create a new block
2. `tensor_to_tensormap`, taking a **3d** numpy array and create a new tensormap

I think they should live in equistore, but before writing tests and polishing them I would like to have some feedback if this actually useful here. Also the names of the functions and the name of the file are to be discussed.

### DIFF
--- a/python/src/equistore/operations/from_arr.py
+++ b/python/src/equistore/operations/from_arr.py
@@ -1,0 +1,68 @@
+import numpy as np
+
+from ..block import TensorBlock
+from ..labels import Labels
+from ..tensor import TensorMap
+
+
+def matrix_to_block(a: np.ndarray) -> TensorBlock:
+    """Create a :class:`equistore.TensorBlock` from a 2d :class`numpy.ndarray`.
+
+    The values of the block are the same as ``a``. The name of the properties label
+    is ``'property'`` and name of the samples label is ``'sample'``. The block has
+    no components.
+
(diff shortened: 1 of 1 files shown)


Luthaf:
I'm not sure I see the use case for these, could you give some examples? For me, most of the value in equistore comes from the metadata handling. 

I do see a point for these inside internal equistore/equisolve tests, but this we can have such function inside the internal test code and not publicly available.

PicoCentauri:
> I'm not sure I see the use case for these, could you give some examples? For me, most of the value in equistore comes from the metadata handling.

The application is of course on the one hand for testing. But on the other could it also be interesting for users. In equisolve's Ridge we define the regulerizer as a `TensorMap`, which in the core is only a 1d array with the length of the number of features. You can do it with a the `ones_like` function and multiplying with an array of the length

```python
alpha_arr = np.array([1,2,3])
alpha = alpha_arr * ones_like(tensor)
```

or directly use a functions like

```python
alpha_arr = np.array([1,2,3])
alpha = tensor_to_tensormap(alpha_arr.reshape(1,-1, 1))
```

I think could be valid options for users.

PR #107, opened 2023-01-26
Fix joining labels with different number of names

PicoCentauri:
When try to join labels of different number of names like

```python
property_labels_1 = Labels(
    ("structure", "prop_1"), np.vstack(2 * [np.arange(5)]).T
)
property_labels_2 = Labels(
    ("structure", "prop_2", "prop_3"), np.vstack(3 * [np.arange(5)]).T
)
```

the `_join_labels` label function raises an error because `np.unique` doesn't work with arrays of inhomogeneous shape. This issue is fixed in this PR.

### DIFF
--- a/python/src/equistore/operations/join.py
+++ b/python/src/equistore/operations/join.py
@@ -1,3 +1,5 @@
+import functools
+import operator
 from typing import List
 
 import numpy as np
@@ -9,31 +11,70 @@ from . import _dispatch
 from ._utils import _check_blocks, _check_same_gradients_components, _check_same_keys
 
 
-def join(tensormaps: List[TensorMap], axis: str):
+def join(tensor_maps: List[TensorMap], axis: str):
     """Join a sequence of :py:class:`TensorMap` along an axis.
 
-    The ``axis`` parameter specifies the type join.
-    For example, if ``axis='properties'`` it will be the joined along the
-    properties of the TensorMaps and if ``axis='samples'`` it will be the
-    along the samples.
+    The ``axis`` parameter specifies the type join. For example, if
+    ``axis='properties'`` it will be the joined along the properties of the
+    TensorMaps and if ``axis='samples'`` it will be the along the samples.
+
+    Possible clashes of the meta data like the property names are resolved by
+    one of the three following strategies:
+
+    1. If the names are the same and the values are unique we keep the names and
+       join the values vertically::
+
+        labels_1 = ["n"], [0, 2, 3] labels_2 = ["n"], [1, 4, 5]
+
+       will lead to::
+
+        joined_labels = ["n"],
+                       [0, 2, 3, 1, 4, 5]
+
+    2. If the names are the same but the values are not unique a new variable
+       ``"tensor"`` is added to the names::
+
+            labels_1 = ["n"], [0, 2, 3]
+            labels_2 = ["n"], [0, 2]
+
+       will lead to:
+
+            joined_labels = ["tensor", "n"],
+                            [[0, 0], [0, 2], [0, 3], [1, 0], [1, 2]]
+
+       In ``joined_labels`` the label values ``[0, 0], [0, 2], [0, 3]``
+       correspond to ``properties_1`` ``[1, 0], [1, 2]`` correspond to
+       ``properties_2``.
+
+    3. If names of the labels are different we change the names to ("tensor",
+       "property"). This case is only supposed to happen when joining in the
+       property dimension, hence the choice of names. For example::
+
+            properties_1 = ["a"], [0, 2, 3]
+            properties_2 = ["b", "c"], [[0, 0], [1, 2]]
+
+       will lead to:
+
+            joined_properties = ["tensor", "property"],
+                         [[0, 0], [0, 1], [0, 2], [1, 0], [1, 1]]
 
     :param tensormaps: sequence of :py:class:`TensorMap` for join
-    :param axis: A string indicating how the tensormaps are stacked.
-                 Allowed values are ``'properties'`` or ``'samples'``.
+    :param axis: A string indicating how the tensormaps are stacked. Allowed
+                 values are ``'properties'`` or ``'samples'``.
 
-    :return: The stacked :py:class:`TensorMap` which has more properties or
-             samples than the input TensorMap.
+    :return: The stacked :py:class:`TensorMap` with more properties or samples
+             than the input TensorMap.
     """
 
-    if len(tensormaps) < 2 or type(tensormaps) not in [list, tuple]:
+    if len(tensor_maps) < 2 or not isinstance(tensor_maps, (list, tuple)):
         raise ValueError("provide at least two `TensorMap`s for joining")
 
-    for ts_to_join in tensormaps[1:]:
-        _check_same_keys(tensormaps[0], ts_to_join, "join")
+    for ts_to_join in tensor_maps[1:]:
+        _check_same_keys(tensor_maps[0], ts_to_join, "join")
 
     blocks = []
-    for key in tensormaps[0].keys:
-        blocks_to_join = [ts.block(key) for ts in tensormaps]
+    for key in tensor_maps[0].keys:
+        blocks_to_join = [ts.block(key) for ts in tensor_maps]
 
         if axis == "properties":
             blocks.append(_join_blocks_along_properties(blocks_to_join))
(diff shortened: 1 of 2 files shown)


python/src/equistore/operations/join.py
```diff
@@ -82,29 +82,33 @@ def _join_labels(labels: List[Labels]) -> Labels:
...
-        a=np.arange(len(values)), repeats=[len(value) for value in values]
+        a=np.arange(len(values_list)), repeats=[len(value) for value in values_list]
     )
 
-    unique_names = np.unique(names)
-    unique_values = np.unique(np.vstack(values), axis=0)
-
+    # We have to flatten the list because the number of elements in each sublist
+    # could be diferent which causes a problem for Numpy.
+    unique_names = np.unique([name for names in names_list for name in names])
```
Luthaf, line 93: I don't get what this list comprehension is doing. Could you add a short example? Why does `for name in names` appears twice here?

python/src/equistore/operations/join.py
```diff
@@ -82,29 +82,33 @@ def _join_labels(labels: List[Labels]) -> Labels:
...
+            new_values = np.vstack(values_list)
+        else:
+            new_names = ["tensor"] + list(names_list[0])
+            new_values = np.hstack(
+                [tensor_values.reshape(-1, 1), np.vstack(values_list)]
+            )
     else:
         new_names = ["tensor", "property"]
-        property_values = np.hstack([np.arange(len(value)) for value in values])
+        property_values = np.hstack([np.arange(len(value)) for value in values_list])
```
agoscinski, line 111:
```suggestion
        property_values = np.hstack([np.arange(len(values)) for values in values_list])
```

python/tests/operations/join.py
```diff
@@ -275,6 +275,55 @@ def test_different_names(self):
...
+        )
+
+        block_2 = TensorBlock(
+            values=np.zeros([2, 5]),
+            samples=self.sample_labels,
+            components=[],
+            properties=property_labels_2,
+        )
+
+        joined_ts = eop.join(
```
agoscinski, line 301: what does `joined_ts` mean? shouldn't be `joined_tm`

agoscinski: would prefer for readability `equi_ops`, but not important for this PR

python/tests/operations/join.py
```diff
@@ -275,6 +275,55 @@ def test_different_names(self):
...
+
+        block_1 = TensorBlock(
+            values=np.zeros([2, 5]),
+            samples=self.sample_labels,
+            components=[],
+            properties=property_labels_1,
+        )
+
+        block_2 = TensorBlock(
+            values=np.zeros([2, 5]),
```
agoscinski, line 295: shouldn't this be `np.zeros([5,2])`?  Since these are `property_labels_2`

python/tests/operations/join.py
```diff
@@ -275,6 +275,55 @@ def test_different_names(self):
...
+        )
+
+        joined_ts = eop.join(
+            [TensorMap(self.keys, [block_1]), TensorMap(self.keys, [block_2])],
+            axis="properties",
+        )
+
+        joined_labels = joined_ts.block(0).properties
+
+        self.assertEqual(joined_labels.names, ("tensor", "property"))
```
agoscinski, line 308: where are these names coming from? I don't see them anywhere defined. The property labels before are ("structure", "prop_1") and ("structure", "prop_2", "prop_3"). I really have problems to understand the join function. Can we improve the doc of it.

python/tests/operations/join.py
```diff
@@ -275,6 +275,55 @@ def test_different_names(self):
...
+
+        ref = np.array(
```
agoscinski, line 310: why do you get 2 properties when you merge 3 properties with 2 properties where only 1 property has the same name?

agoscinski (commented):
Looks good, but I really have problems to understand the join function in the tests. Maybe I am missing something simple, but I dont see it.

python/src/equistore/operations/join.py
PicoCentauri: It flattens a list of lists

```python
names_list = [('prop_a', 'prop_b', 'prop_c'), ('prop_a', 'prop_b')]
[name for names in names_list for name in names]
```

results in 

```
['prop_a', 'prop_b', 'prop_c', 'prop_a', 'prop_b']
```

we can also use `functools.reduce(operator.concat, names_list)` as alex suggested if this is clearer for you.

python/src/equistore/operations/join.py
```diff
@@ -82,29 +82,33 @@ def _join_labels(labels: List[Labels]) -> Labels:
...
+        a=np.arange(len(values_list)), repeats=[len(value) for value in values_list]
     )
 
-    unique_names = np.unique(names)
-    unique_values = np.unique(np.vstack(values), axis=0)
-
+    # We have to flatten the list because the number of elements in each sublist
+    # could be diferent which causes a problem for Numpy.
+    unique_names = np.unique([name for names in names_list for name in names])
     names_are_same = np.all(
```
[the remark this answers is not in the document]
PicoCentauri: No it is a check if they are the same. If the number of unique names is the same as in each label

python/tests/operations/join.py
PicoCentauri: Yes it is quite complicated. The problem is joining meta data that is not unique. Thus problem is exactly what the `join_labels` function is trying to solve. I  move the docstring into the join function.

python/tests/operations/join.py
PicoCentauri: There is a way to improve sure. We could try to only replace those properties that are not the same, but I resolving name clashes is already a bit messy. If you have ideas I am open for suggestions.

python/tests/operations/join.py
PicoCentauri: Both fine for me. I change this in a next commit.

python/tests/operations/join.py
PicoCentauri: Why. These are the values of the Block. It has two samples and 5 properties.

python/src/equistore/operations/join.py
Luthaf: the functools version is not much clearer for me. But adding the small example you give here in a comment would be nice!

python/tests/operations/join.py
Luthaf: > would prefer for readability equi_ops, but not important for this PR

at this point, why not optimize for reading instead of writing and go for `equistore.operations.join`? Any of you feeling strongly about this? I don't mind much

python/src/equistore/operations/join.py
```diff
@@ -17,6 +17,40 @@ def join(tensormaps: List[TensorMap], axis: str):
...
+        properties = ["n"], [0, 2, 3, 1, 4, 5]
+
+    2. If the names are the same but the values are not unique a new variable
+       "tensor" is added to the names. In example::
+
+            properties_1 = ["n"], [0, 2, 3]
+            properties_2 = ["n"], [0, 2]
+
+        will lead to::
+            properties = ["tensor", "n"], [[0, 0], [0, 2], [0, 3], [1, 0], [1, 2]]
```
agoscinski, line 42: wouldn't it be better if the property is called `tensor_n`, so
```suggestion
            properties = ["tensor_n", "n"], [[0, 0], [0, 2], [2, 0], [2, 2], [1, 2]]
```
to differ it from other properties that are resolved the same way? Especially because you cannot have multiple times the same name

python/src/equistore/operations/join.py
```diff
@@ -17,6 +17,40 @@ def join(tensormaps: List[TensorMap], axis: str):
...
+
+    3. If names of the Labels are different we change the names two to
+       ("tensor", "property"). In example::
+
+            properties_1 = ["a"], [0, 2, 3]
+            properties_2 = ["b", "c], [[0, 0], [1, 2]]
+
+        will lead to::
+
+            properties = ["tensor", "q"], [[0, 0], [0, 1], [0, 2], [1, 0], [1, 1]]
```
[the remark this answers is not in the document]
agoscinski: I still have problems to understand this case. It would help me if it is explained in detail how the new labels correspond to the old ones. Let me give the second case as example
```
2. If the names are the same but the values are not unique a new variable
       "tensor" is added to the names. In example::
            properties_1 = ["n"], [0, 2, 3]
            properties_2 = ["n"], [0, 2]
        will lead to::
            joined_properties = ["tensor", "n"], [[0, 0], [0, 2], [0, 3], [1, 0], [1, 2]]

       in joined_properties the labels
       [0, 0], [0, 2], [0, 3] correspond to properties_1
       [1, 0], [1, 2] correspond to properties_2
```

python/tests/operations/join.py
agoscinski: thanks the doc helps a lot, put some comments there

agoscinski (commented):
Thanks, changes help me a lot. For me, only case 3 of the join operation in the doc is unclear

python/tests/operations/join.py
agoscinski: No strong feelings, only eop is hard to interpret for me without reading the import

python/tests/operations/join.py
agoscinski: okay seems like a discussion for another PR

python/src/equistore/operations/join.py
agoscinski: If this is more a core issue about the join function, we can also move the discussion to an issue to merge this fix.

python/tests/operations/join.py
PicoCentauri: I am fine with using the extended version.

python/src/equistore/operations/join.py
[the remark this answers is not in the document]
PicoCentauri: Yes! also good comment. Thanks @agoscinski for reading this so carefully. It took also @Luthaf and me a while to understand what is going on.

python/src/equistore/operations/join.py
PicoCentauri: I will change this to the functools version and extend the comment.

python/src/equistore/operations/join.py
PicoCentauri: Yes, I agree. I think we should merge this fix asap. We should however try to improve the joining of labels further.

python/src/equistore/operations/join.py
```diff
@@ -17,6 +17,40 @@ def join(tensormaps: List[TensorMap], axis: str):
...
+        properties = ["n"], [0, 2, 3, 1, 4, 5]
+
+    2. If the names are the same but the values are not unique a new variable
+       "tensor" is added to the names. In example::
+
+            properties_1 = ["n"], [0, 2, 3]
+            properties_2 = ["n"], [0, 2]
+
+        will lead to::
+            properties = ["tensor_n", "n"], [[0, 0], [0, 2], [2, 0], [2, 2], [1, 2]]
```
PicoCentauri, line 42: @agoscinski I think the extra n is a good idea but this only works nice if you have only one name like `n`. If you have two like (`n`, `l`) I think it doesn't make sense to call the attribute `tensor_n_l` because it might become super lengthy. I would stay with `tensor` for know because this is what we implemented.

PicoCentauri:
I hope I addressed all your comments and already squash the commits.

python/src/equistore/operations/join.py
```diff
@@ -17,6 +19,43 @@ def join(tensormaps: List[TensorMap], axis: str):
     properties of the TensorMaps and if ``axis='samples'`` it will be the
     along the samples.
 
+    Possible clashes of the meta data like the property names
+    are resolved by the following strategy:
+
+    1. If the names (i.e property names) are the same and the
+       values (i.e property values)
+       are unique we keep the names and
+       join the values vertically. In example::
```
Luthaf, line 28:
```suggestion
    1. If the names are the same and the values are unique we keep the names and
       join the values vertically:
```

python/src/equistore/operations/join.py
```diff
@@ -17,6 +19,43 @@ def join(tensormaps: List[TensorMap], axis: str):
...
+
+        labels_1 = ["n"], [0, 2, 3]
+        labels_2 = ["n"], [1, 4, 5]
+
+      will lead to::
+
+        properties = ["n"],
```
Luthaf, line 35:
```suggestion
        joned_labels = ["n"],
```

python/src/equistore/operations/join.py
```diff
@@ -17,6 +19,43 @@ def join(tensormaps: List[TensorMap], axis: str):
...
+                     [0, 2, 3, 1, 4, 5]
+
+    2. If the names are the same but the values are not unique a new variable
+       "tensor" is added to the names. In example::
```
Luthaf, line 39:
```suggestion
       "tensor" is added to the names:
```

python/src/equistore/operations/join.py
```diff
@@ -17,6 +19,43 @@ def join(tensormaps: List[TensorMap], axis: str):
...
+
+            properties_1 = ["n"], [0, 2, 3]
+            properties_2 = ["n"], [0, 2]
+
+        will lead to::
+            properties = ["tensor", "n"],
+                         [[0, 0], [0, 2], [0, 3], [1, 0], [1, 2]]
```
Luthaf, line 46:
```suggestion
            labels_1 = ["n"], [0, 2, 3]
            labels_2 = ["n"], [0, 2]

        will lead to::
            joined_labels = ["tensor", "n"],
                         [[0, 0], [0, 2], [0, 3], [1, 0], [1, 2]]
```

python/src/equistore/operations/join.py
```diff
@@ -17,6 +19,43 @@ def join(tensormaps: List[TensorMap], axis: str):
...
+    3. If names of the Labels are different we change the names two to
+       ("tensor", "property"). In example::
+
+            properties_1 = ["a"], [0, 2, 3]
+            properties_2 = ["b", "c"], [[0, 0], [1, 2]]
+
+        will lead to::
+
+            properties = ["tensor", "property"],
+                         [[0, 0], [0, 1], [0, 2], [1, 0], [1, 1]]
```
Luthaf, line 57:
```suggestion
    3. If names of the labels are different we change the names to
       ("tensor", "property"). This case is only supposed to happen when joining 
       in the property dimension, hence the choice of names. For example:

            properties_1 = ["a"], [0, 2, 3]
            properties_2 = ["b", "c"], [[0, 0], [1, 2]]

        will lead to::

            joined_properties = ["tensor", "property"],
                         [[0, 0], [0, 1], [0, 2], [1, 0], [1, 1]]
```

Luthaf (commented):
Some small suggestions for the doc, the code & tests look good to me

python/src/equistore/operations/join.py
agoscinski: I see, I thought this can happen to multiple properties separately, but even it happens to all or none

python/src/equistore/operations/join.py
```diff
@@ -17,6 +19,43 @@ def join(tensormaps: List[TensorMap], axis: str):
...
+    2. If the names are the same but the values are not unique a new variable
+       "tensor" is added to the names. In example::
+
+            properties_1 = ["n"], [0, 2, 3]
+            properties_2 = ["n"], [0, 2]
+
+        will lead to::
+            properties = ["tensor", "n"],
+                         [[0, 0], [0, 2], [0, 3], [1, 0], [1, 2]]
+
```
agoscinski, line 47:
```suggestion
       in joined_labels the labels values
       [0, 0], [0, 2], [0, 3] correspond to properties_1
       [1, 0], [1, 2] correspond to properties_2
```

agoscinski: you resolved this last time, I am not sure why it was resolved

agoscinski (commented):
I still dont really understand case 3 of the join operation, but I am okay to do this in a separate PR, might be good to add examples in this separate PR that make it clear. From my side approve.

PicoCentauri:
We can squash and merge if everybody is okay...

Luthaf:
I did a tiny bit more reformatting of the docstring, I'll merge once CI is happy!

PR #111, opened 2023-02-03
Use glibc==2.28 and rust==1.59

PicoCentauri:
This PRs updates a number of dependencies to allow building the equistore.

1. All Github actions are upgraded to the latest version
2. Use glibc with version 2.28 of the python wheels. Required by rustup
3. Update the mininal rust version to 1.59. Required by zip.

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -39,16 +39,16 @@ jobs:
           default: true
           target: ${{ matrix.rust-target }}
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: "3.10"
       - name: build wheel
         env:
           RUST_BUILD_TARGET: ${{ matrix.rust-target }}
         run: |
           python -m pip install wheel
           python setup.py bdist_wheel --plat-name ${{ matrix.platform-name }}
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v3
         with:
           name: wheels
           path: dist/*.whl
@@ -63,20 +63,20 @@ jobs:
   build-manylinux-wheels:
     runs-on: ubuntu-20.04
     # TODO: add other arch for linux?
-    name: x86_64 manylinux2010
+    name: x86_64 manylinux
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - name: build manylinux2010 with rust docker image
-        run: docker build -t manylinux2010-with-rust python/scripts/build-wheels
+      - name: build manylinux with rust docker image
+        run: docker build -t manylinux-with-rust python/scripts/build-wheels
       - name: build wheel in docker
-        run: docker run --rm -v $(pwd):/code manylinux2010-with-rust bash -c "cd /code && /opt/python/cp38-cp38/bin/python setup.py bdist_wheel"
+        run: docker run --rm -v $(pwd):/code manylinux-with-rust bash -c "cd /code && /opt/python/cp38-cp38/bin/python setup.py bdist_wheel"
       - name: run auditwheel in docker
-        run: docker run --rm -v $(pwd):/code manylinux2010-with-rust bash -c "auditwheel repair /code/dist/*.whl -w /code/dist"
+        run: docker run --rm -v $(pwd):/code manylinux-with-rust bash -c "auditwheel repair /code/dist/*.whl -w /code/dist"
       - name: remove wheel with wrong tag
         run: sudo rm dist/*linux_x86_64.whl
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v3
         with:
           name: wheels
           path: dist/*.whl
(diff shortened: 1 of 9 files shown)


python/scripts/build-wheels/Dockerfile
```diff
@@ -3,7 +3,18 @@ FROM quay.io/pypa/manylinux2010_x86_64
...
+# update glibc to 2.17 for CentOS 6
+RUN curl -OJ http://copr-be.cloud.fedoraproject.org/results/mosquito/myrepo-el6/epel-6-x86_64/glibc-2.17-55.fc20/glibc-2.17-55.el6.x86_64.rpm
+RUN curl -OJ http://copr-be.cloud.fedoraproject.org/results/mosquito/myrepo-el6/epel-6-x86_64/glibc-2.17-55.fc20/glibc-common-2.17-55.el6.x86_64.rpm
+RUN curl -OJ http://copr-be.cloud.fedoraproject.org/results/mosquito/myrepo-el6/epel-6-x86_64/glibc-2.17-55.fc20/glibc-devel-2.17-55.el6.x86_64.rpm
+RUN curl -OJ http://copr-be.cloud.fedoraproject.org/results/mosquito/myrepo-el6/epel-6-x86_64/glibc-2.17-55.fc20/glibc-headers-2.17-55.el6.x86_64.rpm
+
+RUN rpm -Uvh glibc-2.17-55.el6.x86_64.rpm
+RUN glibc-common-2.17-55.el6.x86_64.rpm
+RUN glibc-devel-2.17-55.el6.x86_64.rpm
+RUN glibc-headers-2.17-55.el6.x86_64.rpm
```
Luthaf, line 15: Doesn't this make us no longer compatible with the manylinux wheels? The absence of a modern glibc is why we where using 1.63 toolchain here before.

PicoCentauri:
@Luthaf No as you can also see in #103 and #105 the wheel for manylinux fails with an error

```
/tmp/tmp.Sm5JVJAbnX/rustup-init: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by /tmp/tmp.Sm5JVJAbnX/rustup-init)
[235](https://github.com/lab-cosmo/equistore/actions/runs/4082325647/jobs/7036557282#step:3:236)
/tmp/tmp.Sm5JVJAbnX/rustup-init: /lib64/libc.so.6: version `GLIBC_2.15' not found (required by /tmp/tmp.Sm5JVJAbnX/rustup-init)
[236](https://github.com/lab-cosmo/equistore/actions/runs/4082325647/jobs/7036557282#step:3:237)
/tmp/tmp.Sm5JVJAbnX/rustup-init: /lib64/libc.so.6: version `GLIBC_2.17' not found (required by /tmp/tmp.Sm5JVJAbnX/rustup-init)
[237](https://github.com/lab-cosmo/equistore/actions/runs/4082325647/jobs/7036557282#step:3:238)
The command '/bin/sh -c curl https://sh.rustup.rs/ -sSf | sh -s -- -y --profile minimal --default-toolchain 1.63' returned a non-zero code: 1
```

Luthaf:
Right, looks like rustup (in addition to rustc/cargo) now needs a modern glibc =/. 

In this case, this is fine for a temporary fix, and we can look into the implications for manylinux1 wheels later. We might have to drop manylinux1, and support manylinux2014 instead. Maybe the official manylinux docker images now have a rust compiler installed as well.

PicoCentauri:
Okay, so yes lets go for this fix, There is something additionally going wrong with the zip library which I will check as well.

PicoCentauri:
It seems even with hardcore updating the glibc version we later run into issue. maybe we try manylinux2014. For manylinux2010 also the support for python 3.6 does not exist anymore....

DavideTisi (approved):

CONTRIBUTING.rst
```diff
@@ -56,7 +56,7 @@ on equistore:
 - **the rust compiler**: you will need both ``rustc`` (the compiler) and
   ``cargo`` (associated build tool). You can install both using `rustup`_, or
   use a version provided by your operating system. We need at least Rust version
-  1.56 to build equistore.
+  1.59 to build equistore.
 - **Python**: you can install ``Python`` and ``pip`` from your operating system.
   We require a Python version of at least 3.6.
```
jwa7, line 61: Should the "We require a Python version of at least 3.x" be updated?

.github/workflows/build-wheels.yml
```diff
@@ -96,14 +96,14 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: "3.8"
```
jwa7, line 101: Here you use python 3.8 but in other files, such as .github/workflows/tests.yml, it says python 3.10. Should python versioning be consistent everywhere?

agoscinski:
>It seems even with hardcore updating the glibc version we later run into issue. maybe we try manylinux2014. For manylinux2010 also the support for python 3.6 does not exist anymore....

Is this something for a new issue for a different PR? (Also a bit related to @jwa7 comment)

Seems to me fine if the information in PR's first message is in the rebased/squashed commit.

PicoCentauri:
Yes we can discuss this in another issue.

CONTRIBUTING.rst
PicoCentauri: But 3.x would also mean 3.5 which we do not support.

.github/workflows/build-wheels.yml
PicoCentauri: Yes we can try it. At least the manylinux we use here should work for 3.10

CONTRIBUTING.rst
jwa7: Apologies, perhaps I was too ambiguous with what I meant here - I wasn't suggesting a change to "3.x", just to the specific and appropriate version (i.e. "3.10"), consistent with the named versions in all other files.

jwa7: Ah I see. Thanks! Looks good to me.

PR #112, opened 2023-02-06
Restructure operations documentation pages

PicoCentauri:
Our operations routines are growing and growing 🚀. We reached a number of operations where the list in the documentation becomes a bit crowded. Therefore, this PR introduces 5 subsections for our `TensorMap` and `TensorBlock` operations:

1. Creation operations
2. Linear algebra
3. Logic function
4. Manipulation operations
5. Mathematical functions

So far this only applies to the documentation. Question is if we should also introduces this structure for the source files themselves?


python/src/equistore/operations/multiply.py
```diff
@@ -9,14 +9,17 @@
...
-    rule of the derivatives.
+    If gradients are present in ``A`` a multiplication (:math:`*`) is only performed if
+    ``B`` is a :py:class:`TensorMap` as well. The multiplication is performed with the
+    rule of the derivatives:
+
+    .. math::
+        \nabla(A * B) = B * \nabla A + A * \nabla B
 
     :param A: First :py:class:`TensorMap` for the addition.
     :param B: Second instance for the multiplication. Parameter can be a scalar a array
```
jwa7, line 25: Here it says that "Parameter can be a scalar **a array**..." - can ``B`` be an array? It looks like there are only methods for scalars (``_multiply_block_constant``) and blocks with the same dimensions (``_multiply_block_block``). 

If not, just edit so it reads "Parameter can be a scalar or a :py:class"`TensorMap`...". Apart from that everything looks good :)

jwa7 (changes requested):

python/src/equistore/operations/multiply.py
PicoCentauri: Ohh yes. I think this is a leftover from our draft.

PR #113, opened 2023-02-06
added std and variance

DavideTisi:
I add this for the variance and the std over the samples. These are used in the Ridge regression.

For the values in the block it is straightforward, and it is already implemented. The problems are the derivatives. Also because I do not know how much can be used the true derivative of the variance:

`Grad[Var[X]] = 2(E[X Grad[X]] - E[X]E[Grad[X]])`

for the std:
`Grad[Std(X)]=0.5(Grad[Var[X]])/Std(X)= (E[X Grad[X]] - E[X]E[Grad[X]])/Std(X)`

what do you think?

### DIFF
--- a/python/src/equistore/operations/reduce_over_samples.py
+++ b/python/src/equistore/operations/reduce_over_samples.py
@@ -140,20 +140,32 @@ def _reduce_over_samples_block(
                 values_grad_result = values_grad_result / bincount.reshape(
                     (-1,) + (1,) * len(other_shape)
                 )
-                for i, s in enumerate(new_gradient_samples):
-                    data_result[i] = data_result[i] * values_mean[s[0]]
-                data_result = 2 * (values_grad_result - data_result)
-                if reduction == "std":
-                    v_shape = values_result.shape
-                    d_shape = data_result.shape
-                    v_shape_broadcast = (
-                        (v_shape[0],)
-                        + (1,) * len(d_shape[1 : -(len(v_shape) - 1)])
-                        + v_shape[1:]
-                    )
-                    data_result = (
-                        0.5 * data_result / values_result.reshape(v_shape_broadcast)
-                    )
+
+                if reduction == "variance":
+                    for i, s in enumerate(new_gradient_samples):
+                        data_result[i] = data_result[i] * values_mean[s[0]]
+                    data_result = 2 * (values_grad_result - data_result)
+                else:  # std
+                    for i, s in enumerate(new_gradient_samples):
+                        data_result[i] = (
+                            values_grad_result[i] - (data_result[i] * values_mean[s[0]])
+                        ) / values_result[s[0]]
+                # if reduction == "std":
+                #     v_shape = values_result.shape
+                #     d_shape = data_result.shape
+                #     v_shape_broadcast = (
+                #         (v_shape[0],)
+                #         + (1,) * len(d_shape[1 : -(len(v_shape) - 1)])
+                #         + v_shape[1:]
+                #     )
+                #     # values_grad = []
+                #     # print("l", data_result.shape, len(new_gradient_samples))
+                #     for i, s in enumerate(new_gradient_samples):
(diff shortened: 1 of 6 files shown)


DavideTisi:
it should pass the test but i think that they are quite difficult to understand.

PicoCentauri (changes requested):
Nice @Davide. Maybe adding a unittest that tests if only one sample is present there is no error raised and we return either nan or 0.

Also didnt we said we want to raise a warning for gradients if there is only one sample?

DavideTisi:
> Nice @davide. Maybe adding a unittest that tests if only one sample is present there is no error raised and we return either NaN or 0.
> 
> Also didnt we said we want to raise a warning for gradients if there is only one sample?

So what happens now is that if there is only one sample variance e std are zero, (there is a test for that I will put a comment maybe). 
For the gradients: the gradient of the variance is correctly zero; the problems appear in the gradient of std which has a term at the denominator. Now "it raises a warning" if an element of std is zero, but is the standard "division by zero" warning of division of NumPy arrays. Maybe you meant something custom.

But more important I thought a bit more about that and the behaviour is wrong: 
As I said the code in the present state leaves numpy (or any underline system in which you store the arrays to handle the exception of division by zero). I think that numpy put a None for that element, but this is wrong in our case.

Suppose you have only one value to compute the std; then the std of one number is the function identically zero, which derivative is again the function identically zero, not matter of the formula that simply do not apply for that case. 

All this treatment is ok but it must also be considered that the derivative of std probably has not a big meaning and it is not super defined.

DavideTisi:
Conclusion yes we for sure have at leat to catch the error and replace that values with 0

DavideTisi:
Thanks, @PicoCentauri for the help.
The last commits should handle adequately the case with only one value for the std.
I added tests for that case.

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -51,11 +52,29 @@ def _reduce_over_samples_block(
...
+            values_result2 = _dispatch.zeros_like(
+                block_values, shape=(new_samples.shape[0],) + other_shape
+            )
+            _dispatch.index_add(
+                values_result2,
+                block_values**2,
+                index,
+            )
+            values_result2 = values_result2 / bincount.reshape(
+                (-1,) + (1,) * len(other_shape)
```
PicoCentauri, line 70: I like this trick :)

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -245,3 +303,60 @@ def mean_over_samples(tensor: TensorMap, samples_names: List[str]) -> TensorMap:
     return _reduce_over_samples(
         tensor=tensor, samples_names=samples_names, reduction="mean"
     )
+
+
+def std_over_samples(tensor: TensorMap, samples_names: List[str]) -> TensorMap:
+    r"""Create a new :py:class:`TensorMap` with the same keys as
+    as the input ``tensor``, and each :py:class:`TensorBlock` is obtained
+    performing the std deviation of the corresponding input :py:class:`TensorBlock`
+    over the ``samples_names`` indices.
```
PicoCentauri, line 312: We can leave this for now, but should try to shorten the "short" description into a line.

python/tests/operations/reduce_over_samples.py
```diff
@@ -537,13 +539,402 @@ def test_reduction_allsamples(self):
...
+        )
+        self.assertTrue(
+            np.all(
+                np.allclose(
+                    np.std(bl1.values[46:], axis=0),
+                    reduce_tensor_ps.block(0).values[9],
+                    rtol=1e-13,
+                )
+            )
+        )
```
PicoCentauri, line 614: Also lets try to split this into individual tests in the future. This makes debugging in case of errors easier.

PicoCentauri (approved):
Looks, good. I have some general comments that we can address after this PR.

PR #114, opened 2023-02-07
Enhance TensorMap check function

PicoCentauri:
The current check function for TensorMaps was only checking for keys and giving not clear error messages. I enhances this function by also checking that number of blocks (same as number of keys) is the same.


DavideTisi (approved):
I like it, for me it is fine and can be merged. 
The only thing that I noticed is that there are no tests for these functions. But honestly are very straightforward.

PR #115, opened 2023-02-08
Check components in a propper way

PicoCentauri:
As @curiosity54 found out: The `_check_blocks` function is not testing correctly for components of two blocks. Since components are of type `List[Labels]` a simple comparison with `==`  is not suitable. Here, `_check_blocks` is extended to test for the length and for the equality using the same code as for gradients.

### DIFF
--- a/python/src/equistore/operations/_utils.py
+++ b/python/src/equistore/operations/_utils.py
@@ -50,16 +50,30 @@ def _check_blocks(a: TensorBlock, b: TensorBlock, props: List[str], fname: str):
     for prop in props:
         err_msg = f"Inputs to {fname} should have the same {prop}."
         if prop == "samples":
+            if not a.samples.names == b.samples.names:
+                raise ValueError(err_msg)
             if not np.all(a.samples == b.samples):
                 raise ValueError(err_msg)
         elif prop == "properties":
+            if not a.properties.names == b.properties.names:
+                raise ValueError(err_msg)
             if not np.all(a.properties == b.properties):
                 raise ValueError(err_msg)
         elif prop == "components":
-            if len(a.components) != len(b.components) or (
-                not np.all([labels in b.components for labels in a.components])
-            ):
+            if len(a.components) != len(b.components):
                 raise ValueError(err_msg)
+
+            for ca in a.components:
+                flag = False
+                for cb in b.components:
+                    if ca.names == cb.names:
+                        flag = True
+                        if not np.all(ca == cb):
+                            raise ValueError(err_msg)
+                # if flag is False ca.names == cb.names has never been satisfied
+                if not flag:
+                    raise ValueError(err_msg)
+
         elif prop == "gradients":
             grads_a = a.gradients_list()
             grads_b = b.gradients_list()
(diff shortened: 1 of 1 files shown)


DavideTisi:
yes for components, you need a loop. 
for example, look at the `all_close` function:

https://github.com/lab-cosmo/equistore/blob/467baf039d6349996545b65aff74c05f4002521e/python/src/equistore/operations/allclose.py#L179-L188

The changes should  fix the issue, maybe we want to add to `_check_blocks` also other check for the names of the Labels: `samples`, `components` and `properties`. So far it seems to me it checks only the values.

DavideTisi:
In particular if you have two labels with different names the comparison fails. for example:
Given 
```
a = TensorBlock(
            values=np.array([[1], [2]]),
            samples=Labels(["samples"], np.array([[0], [6]], dtype=np.int32)),
            components=[],
            properties=Labels(["properties"], np.array([[0]], dtype=np.int32)),
        )

b = TensorBlock(
            values=np.array([[[1], [4]], [[44], [2]]]),
            samples=Labels(["samples"], np.array([[0], [2]], dtype=np.int32)),
            components=[
                Labels(["component"], np.array([[0], [6]], dtype=np.int32)),
            ],
            properties=Labels(["properties"], np.array([[0]], dtype=np.int32)),
        )
```

The command:
`np.all(a.samples==b.samples)`
gives the following error
`TypeError: Cannot compare structured arrays unless they have a common dtype.  I.e. `np.result_type(arr1, arr2)` must be defined.`

PicoCentauri:
Thanks @DavideTisi for the addition. I will merge this now since this is something which is blocking usage. I open an issue discussing the tests.

curiosity54:
Thanks @PicoCentauri and @DavideTisi!

jwa7:
Just to add to this, in allclose.py fxn ``allclose_block_raise()`` there are the lines:
https://github.com/lab-cosmo/equistore/blob/cb4148e556a9bba2b57a4042b8b7e0e11fb1bdcc/python/src/equistore/operations/allclose.py#L168-L175

I think that there are some redundant checks here, the properties are checked like they are components (i.e. potentially multiple dimensions), but in fact should be checked analogous to samples. i.e. a) the length of the properties don't need to be checked, as these are already checked in the ``np.all(block1.values.shape == block2.values.shape)`` comparison and b) the properties don't need to be looped over, as they are just a single Labels object and not a list of them:
```py
if not block1.properties.names == block2.properties.names:
        raise ValueError("properties names are not the same or not in the same order")

if not np.all(block1.properties == block2.properties):
    raise ValueError("properties not the same or not in the same order")
```

Furthermore, there are some redundant uses of if/else in this function, for instance the ``else`` can be removed from:
```py
if not (len(block1.components) == len(block2.components)):
        raise ValueError("different number of components")
else:
    for c1, c2 in zip(block1.components, block2.components):
        if not (c1.names == c2.names):
            raise ValueError(
                "components names are not the same or not in the same order"
            )

        if not np.all(c1 == c2):
            raise ValueError("components not the same or not in the same order")
```

DavideTisi:
@jwa7 I pin you in #121 where we all close to work with the inner functions.

PR #116, opened 2023-02-08
Split function and more numpy-esque slice

jwa7:
**1. Split**

This PR introduces the ``split`` and ``split_block`` functions, which are essentially wrappers for the private function ``_slice_block`` found in operations/slice.py. They allow tensors (TensorMap or TensorBlock, respectively) to be split into multiple tensors according to a specified index name along either the "samples" or "properties" axis, and returned in a list.

How they work:
```py
import numpy as np
from equistore.operations import split

# Load a tensor - i.e. SOAP for 11 training structures (idxs 0 -> 11 inc.)
tensor = io.load("path/to/tensor.npz")

split_tensor = split(
    tensor,
    axis="samples",
    grouped_idxs=[
        Labels(
            names=["structure"],
            values=np.array([[0], [6], [7]])
        ),
        Labels(
            names=["structure"],
            values=np.array([[2], [3], [4]])
        ),
        Labels(
            names=["structure"],
            values=np.array([[1], [5], [8], [9], [10]])
        ),
    ]
)
```
In this case, there are 3 groups of structure indices passed as a list of Labels. Inclusion of the ``axis`` argument means that splitting can only occur along one axis with a single function call, in contrast to the slice function where one can specify both the ``samples`` and ``properties`` Labels to slice with. This is intended, so as to mirror the ``join`` function, for which this operation is essentially the inverse.

The ``split*`` functions wrap the private ``_slice_block`` fxn as opposed to the public ``slice*`` fxns as all appropriate input checks are performed in ``split*`` and so would be run twice by calling ``slice*``. To split a tensor into N tensors, N slice operations have to be performed. This is not the ideal under-the-hood implementation, but no better solution could be found at present.

The ``split`` function provides a basis for a ``split_tensors`` function in equisolve that can be used to perform splitting of multiple TensorMaps at once, i.e. for a train-test(-validation) split.

**2. Slice**

Also included in the same PR are changes to slice and appropriate updates to unit tests. The ``numpy.slice`` function does not raise warnings when slicing arrays to emptiness, so after a discussion with @PicoCentauri it was decided that such warnings should be removed from ``equistore.slice`` too. Checks for the arguments to ``slice`` and ``slice_block`` functions were unified, and the ``_slice_tensormap`` function was removed. Unit tests related to error messages were updated with new messages, and those related to warnings were removed.

### DIFF
--- a/python/src/equistore/operations/_utils.py
+++ b/python/src/equistore/operations/_utils.py
@@ -3,6 +3,7 @@ from typing import List
 import numpy as np
 
 from ..block import TensorBlock
+from ..labels import Labels
 from ..tensor import TensorMap
 
 
@@ -111,3 +112,22 @@ def _check_parameters_in_gradient_block(
                 f"The requested parameter '{p}' in {fname} is not a valid parameter"
                 "for the TensorBlock"
             )
+
+
+def _labels_equal(a: Labels, b: Labels, exact_order: bool):
+    """
+    For 2 :py:class:`Labels` objects ``a`` and ``b``, returns true if they are
+    exactly equivalent in names, values, and elemental positions. Assumes that
+    the Labels are already searchable, i.e. they belong to a parent TensorBlock
+    or TensorMap.
+    """
+    # They can only be equivalent if the same length
+    if len(a) != len(b):
+        return False
+    # Check the names
+    if not np.all(a.names == b.names):
+        return False
+    if exact_order:
+        return np.all(np.array(a == b))
+    else:
+        return np.all([a_i in b for a_i in a])
--- a/python/src/equistore/operations/_utils.py
+++ b/python/src/equistore/operations/_utils.py
@@ -190,5 +190,4 @@ def _labels_equal(a: Labels, b: Labels, exact_order: bool):
         return False
     if exact_order:
         return np.all(np.array(a == b))
-    else:
-        return np.all([a_i in b for a_i in a])
+    return np.all([a_i in b for a_i in a])
--- a/python/src/equistore/operations/slice.py
+++ b/python/src/equistore/operations/slice.py
@@ -1,4 +1,4 @@
-import warnings
+from typing import Optional, Union
 
 import numpy as np
 
@@ -7,7 +7,11 @@ from ..labels import Labels
 from ..tensor import TensorMap
 
 
-def slice(tensor: TensorMap, samples=None, properties=None) -> TensorMap:
+def slice(
+    tensor: TensorMap,
+    samples: Optional[Labels] = None,
+    properties: Optional[Labels] = None,
+) -> TensorMap:
     """
     Slices an input :py:class:`TensorMap` along the samples and/or properties
     dimension(s). ``samples`` and ``properties`` are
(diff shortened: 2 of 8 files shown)


python/src/equistore/operations/split.py
```diff
@@ -0,0 +1,211 @@
...
+    Checks the arguments passed to :py:func:`split` and :py:func:`split_block`.
+    """
+    # Check types
+    if not isinstance(axis, str):
+        raise TypeError("``axis`` should be passed as a ``str``")
+    if axis not in ["samples", "properties"]:
+        raise ValueError("must pass ``axis`` as either 'samples' or 'properties'")
+    if not isinstance(grouped_idxs, list):
+        raise TypeError(
+            "``axis`` should be passed as a ``list`` of equistore ``Labels``"
```
DavideTisi, line 182:
```suggestion
            "``grouped_idxs`` should be passed as a ``list`` of equistore ``Labels``"
```

python/src/equistore/operations/split.py
```diff
@@ -0,0 +1,211 @@
...
+                    "the names of all ``Labels`` passed in ``grouped_idxs``"
+                    + " must be equivalent"
+                )
+    # Get a single block
+    block = tensor.block(0) if isinstance(tensor, TensorMap) else tensor
+
+    # Check the names in grouped_idxs Labels are contained within the names for
+    # the block
+    names = block.samples.names if axis == "samples" else block.properties.names
+    for ref_i, ref_name in enumerate(ref_names):
```
DavideTisi, line 205:
```suggestion
    for ref_name in ref_names:
```

python/tests/operations/split.py
```diff
@@ -0,0 +1,524 @@
...
+    """
+    return TensorBlock(
+        values=np.full((len(labels), 1), 0.0),
+        samples=labels,
+        components=[],
+        properties=Labels(["p"], np.array([[0]], dtype=np.int32)),
+    ).samples
+
+
+def _labels_equal(a: Labels, b: Labels, exact_order: bool):
```
DavideTisi, line 520: this function is very nice, I would put it in some _utilis file in src not hidden here

DavideTisi: but it should also check the label.names

DavideTisi (commented):
General very good just I would move the _label_equal in a more easy to find place. The function is quite general but here it is a bit hidden.

python/tests/operations/split.py
jwa7: Do you mean in ``equistore/python/src/equistore/operations/_utils.py``?

DavideTisi (approved):
for me it is ok

jwa7:
Thanks for the review @DavideTisi, I think all is good with tests now, right?  If so I'll wait for approval/changes from @PicoCentauri then squash

DavideTisi:
Yes for me everything is ok

Il giorno dom 12 feb 2023 alle 19:02 Joseph W. Abbott <
***@***.***> ha scritto:

python/src/equistore/operations/_utils.py
```diff
@@ -111,3 +112,22 @@ def _check_parameters_in_gradient_block(
...
+    # They can only be equivalent if the same length
+    if len(a) != len(b):
+        return False
+    # Check the names
+    if not np.all(a.names == b.names):
+        return False
+    if exact_order:
+        return np.all(np.array(a == b))
+    else:
+        return np.all([a_i in b for a_i in a])
```
PicoCentauri, line 133: How does this function go with @DavideTisi  changes in #121? We maybe can include `_labels_equal ` into #121...

PicoCentauri, line 133: How does this function go with @DavideTisi  changes in #121? We maybe can include `_labels_equal ` into #121...

python/src/equistore/operations/slice.py
```diff
@@ -358,35 +280,42 @@ def _slice_block(block: TensorBlock, samples=None, properties=None) -> TensorBlo
...
-    # Iterate over, and slice, each block (+ gradients) of the input TensorMap in turn.
-    return TensorMap(
-        keys=tensormap.keys,
-        blocks=[_slice_block(block, samples, properties) for _, block in tensormap],
-    )
+    # Must pass either samples or properties or both
+    if samples is None and properties is None:
+        raise ValueError(
+            "you must specify either samples or properties (or both) to slice by"
+        )
```
PicoCentauri, line 295: What is happening if we don't have this check? Are the slice and split function only returning a copy of the original Block? If yes we might think about removing this test.

python/src/equistore/operations/split.py
```diff
@@ -0,0 +1,211 @@
...
+        if axis == "samples":
+            new_block = _slice_block(block, samples=indices)
+        else:  # properties
+            new_block = _slice_block(block, properties=indices)
+        new_blocks.append(new_block)
+
+    return new_blocks
+
+
+def _check_args(
```
PicoCentauri, line 169: Why is there another check_args function here? The name might be confusing with the one from the slice functions even of they are in different files.

PicoCentauri (changes requested):
I am also almost happy. We should only avoid naming functions equally even if they are in different files. 


Some ideas that go in the spirit of these `_check_args` args functions. So far we only have them here, right?
It might be worth to write a general decorator `@check_args` that checks the arguments of a decorated function based on the type hints. This is a bit of work but could be on the one hand really helpful for the users. On the other it destroys a bit the idea of Python to be vague with input types...

python/src/equistore/operations/slice.py
jwa7: Currently, yes if we take out this function ``slice(tensor, samples=None, properties=None)`` will return a copy of the input tensor. Is this the behaviour we want? Or do we want to throw an error if neither samples or properties are specified?

python/src/equistore/operations/split.py
jwa7: They are called the same thing but do different things in ``slice.py`` and ``split.py``. I can change the names but don't see much point as they just check arguments in each respective file and should be hidden

python/src/equistore/operations/slice.py
jwa7: Similarly if I call ``split(tensor, axis="samples", grouped_idxs=[])`` should this also return the original tensor, or throw an error?

@DavideTisi thoughts?

python/src/equistore/operations/_utils.py
jwa7: I don't think @DavideTisi included it in #121, which has now been merged. Shall I keep it in ``_utils.py`` for this PR?

PicoCentauri (approved):
Thanks for the changes @jwa7!

jwa7:
Thanks for the reviews @DavideTisi @PicoCentauri !

PR #117, opened 2023-02-10
new meta-operations stack, starting with first function: ``unique``

jwa7:
**Summary**

Function to get the sorted unique indices of a TensorMap (or TensorBlock), along either the "samples" or "properties" ``axis`` and according to specified ``names``. the ``unique`` function returns a Labels object with the unique indices found across all of the TensorBlocks in a given input ``tensor``. 

Passing the name of a gradient parameter too, i.e. using ``gradient_param="cell"`` will return the unique indices only for the gradient blocks in the input tensor. Along the properties axis, the unique indices of a given name of the TensorBlocks in a given TensorMap will be exactly equivalent to those of the gradient blocks. However, as the samples may be different, this function needs to handle gradient and parent blocks separately.

Because these functions do not act on the actual values of the TensorMap, but instead the metadata (i.e. Labels), we have introduced a new section of documentation and functions to equistore: "metaoperations". This can also be where useful functions such as ``equal`` (where 2 TensorMaps can be compared for exact equivalence in their metadata, i.e. keys, and samples/components/properties/ of their blocks and gradients) and other logic operations such as ``intersection`` ``union``.

Note that while these functions acts on metadata and not values, it is analogous to numpy in the sense that sorted unique indices (which themselves are integers) are returned in a Labels object. If no unique indices are found along a given axis, i.e. in the case of a block whose samples dimension has been sliced to zero, an empty Labels object with the correct names will be returned.

**Usage**

For instance, to find the unique structure indices of the normal blocks of a TensorMap:

```py
from equistore import io
from equistore.metaoperations import unique

# Load a TensorMap - i.e. SOAP with 10 structures (indices 0 -> 9 inc.)
tensor = io.load("/path/to/tensor.npz")
unique_idxs = unique(tensor, axis="samples", names=["structure"])
unique_idxs
>>> Labels([(0,), (1,), (2,), (3,), (4,), (5,), (6,), (7,), (8,), (9,)], dtype=[('structure', '<i4')])
```

and for the gradient blocks to find, for instance, the samples named "atom" for the "positions" gradients:

```py
unique_grad_idxs = unique(tensor, axis="samples", names=["atom"], gradient_param="positions")
unique_grad_idxs
>>> Labels([( 0,), ( 1,), ( 2,), ( 3,), ( 4,), ( 5,), ( 6,), ( 7,), ( 8,), ( 9,), (10,)], dtype=[('atom', '<i4')])
```


Luthaf:
Since this operate on `Labels` exclusively, I don't think it makes sense to be an operation on a TensorMap/TensorBlock. It should instead work on a `Labels` (I can still be a Python `operation`)

PicoCentauri:
Why not, and you could then pass the labels to the `slice` function.

jwa7:
I'm not sure I completely agree with the above. Finding the unique values and slicing a single ``Labels`` object is straightforward in equistore already. You can take a Labels object and return a Labels object in a single line of code, as follows:

```py
import numpy as np

samples_labels = io.load("/path/to/tensor.npz").block(0).samples

# a 'slice' fxn on Labels
samples_labels[:5]
>>> Labels([(9, 19), (9, 20), (9, 21), (9, 22), (9, 23)], dtype=[('structure', '<i4'), ('center', '<i4')])

# a 'slice' fxn on Labels
samples_labels["structure"][:5]
>>> Labels([9, 9, 9, 9, 9], dtype=int32)

# a 'unique' fxn on Labels
np.unique(samples_labels["structure"])
>>> Labels([9], dtype=int32)
```

but this is not the point of the proposed ``unique`` function. ``unique`` is intended to be a convenience function that takes as input a *TensorMap* and finds the unique indices across *all* blocks (normal or gradient), returning a ``Labels`` object. It is not intended to act on a Labels object. 

I have found myself needing and using the ``unique`` function as implemented in this PR in multiple places in my work. The code is non-trivial enough (i.e. cannot be reduced to a single line of code as in the example shown above), in my opinion, to warrant having a public convenience function that does this job.

Perhaps the question is where, and under what name, should this function live? 

And re: issue #120, I agree that perhaps only ``unique`` (and not ``unique_block``) should be public, but maybe this should be addressed in a single overhaul of equistore operations, where we decide on a consistent philosophy moving forward.

jwa7:
After the discussion last week with @PicoCentauri, and the reason for introducing the new class of operations ``metaoperations`` in this PR, it seemed like there needed to be a distinction between operations that act on metadata, and those that act on data.

At the lower-level, I think this distinction is clear:

* ``operations``: acting on the data (numpy arrays or torch tensors). Examples, i.e. equal: ``np.all(block1.values == block2.values)``, intersection (``np.intersect1d``), unique (``np.unique``) and their torch counterparts

* ``metaoperations``: acting on metadata (Labels objects). Examples: i.e. equal (``labels_obj_a == labels_obs_b``), intersection (``labels_obj_a.intersection(labels_object_b)`` or ``intersection(a, b)``), unique (``unique(labels_obj_a, names)``, which is equivalent to ``np.unique(labels_obj_a[names])`` and is essentially your suggestion @Luthaf, acting on and returning a Labels object)

But on the higher level, with functions that act on TensorMaps, the distinction is less clear. For instance, the ``allclose`` module makes comparisons of both data and metadata. A higher level function like the proposed ``equal()`` and discussed in issue #119 can compare both data and metadata, or just metadata, of 2 TensorMaps. The higher-level ``unique()`` fxn proposed in this PR works on only the metadata of a TensorMap. 

Should all of these higher level functions, that take as input and return as output different things, all just live in ``equistore.operations``? 

If not, where should ``unique()`` live? I'm not convinced that it belongs in any of the new categories of operations introduced in d20a5745f1415c3435747df02859a10512d40878, i.e. creation, linear_algebra, logic, manipulation or math.

Also, I can see why 'unique' might not be an appropriate name here, as it clashes with the numpy intuition of what ``np.unique`` does. Any suggestions for another name?

PicoCentauri:
Thanks for the detailed explanation. I can see application of applying unique to a `TensorMap` and if you say that these are useful for your work it might also for other users. 

To your question: With this PR we have two classes now functions:

1. Functions that **take** `TensorMap` instance and **return** `TensorMap`instance  and 
2. functions that **take** a `TensorMap` instance and **return** `Labels` instance. 

`allclose` belongs to the first category which we call `operations` and `unique` to the second.

jwa7:
Closed by PR #125 . Creating a new ``metaoperations`` stack of operations seems a bit drastic for now, so instead these functions in #125 are renamed to be descriptive, and the documentation organised into the operations/set sub-division.

Issue #119, opened 2023-02-10
Move`_check_maps` and `_check_blocks` into public `equal` and `equal_block` functions


PicoCentauri:
In #115 we discussed that the two functions `_check_maps` and `_check_blocks` which check if meta data between two TensorMaps/TensorBlocks is the same. Since @Luthaf has a valid point of not testing private function I propose we make them public but change their scope a bit.

The function should also have the option to check the values/gradients and not only the meta data. With this we can create a functions similar to `allclose`. I think of something like:

```python
def equal(
    tensor1: TensorMap,
    tensor2: TensorMap,
    test_data: bool=True,
) -> bool:
    ...
   
def equal_raise(
    tensor1: TensorMap,
    tensor2: TensorMap,
    test_data: bool=True,
) -> bool:
    ...

def equal_block(
    tensor1: TensorMap,
    tensor2: TensorMap,
    test_data: bool=True,
) -> bool:
    ...

def equal_block_raise(
    tensor1: TensorMap,
    tensor2: TensorMap,
    test_data: bool=True,
) -> bool:
    ...
```

jwa7:
An argument of semantics / UX philosophy here, but should the out-of-the-box behaviour of a function like ``equal`` be to act on the whole TensorMap, i.e. data and metadata? I know you have default behaviour ``test_data=True`` here, but I think negating the argument to enforce explicit specification of metadata-only comparison makes more sense.

For instance change from:
```py
def equal(
    tensor1: TensorMap,
    tensor2: TensorMap,
    test_data: bool = True,
) -> bool:
    ...
```
to:
```py
def equal(
    tensor1: TensorMap,
    tensor2: TensorMap,
    metadata_only: bool = False,
) -> bool:
    ...
```

Such that:
```py
# Compare whole TensorMaps, i.e. an X and Y
equal(tensor1=X, tensor2=Y)
>>> False

# Compare just metadata, i.e. an X and Y
equal(tensor1=X, tensor2=Y, metadata_only=True)
>>> True
```

jwa7:
But to add I think having these functions as public is a great idea, I find myself using them (or variations of them) a lot

DavideTisi:
i close this since we merged #121 which implements what is requested here

PR #121, opened 2023-02-11
rewrite of _utils._check_blocks _utils._check_same_gradients,

DavideTisi:
In the spirit of #119  I rewrote them in a more stringent way and allclose now use them to do all the checks for the metadata . This implies that all the test of allclose now are on these functions. I think everything has a test (maybe not all the gradient possibilities).

What do you think? 
I, of course, can create the equal functions, being the same as allclose except for the more stringent numerical checks, but I do not know if it really usefull.

### DIFF
--- a/python/src/equistore/operations/_utils.py
+++ b/python/src/equistore/operations/_utils.py
@@ -89,7 +89,7 @@ def _check_same_gradients(a: TensorBlock, b: TensorBlock, props: List[str], fnam
      for an operation.
 
     The functions verifies that the metadata of the given props is the same
-    (length, names, values).
+    (length, names, values) and in the same order.
 
     If props is None it only checks if the ``'parameters'`` are consistent.
 
--- a/python/src/equistore/operations/_utils.py
+++ b/python/src/equistore/operations/_utils.py
@@ -48,58 +48,121 @@ def _check_blocks(a: TensorBlock, b: TensorBlock, props: List[str], fname: str):
                  ``'components'`` and ``'gradients'``.
     """
     for prop in props:
-        err_msg = f"Inputs to {fname} should have the same {prop}."
+        err_msg = f"Inputs to '{fname}' should have the same {prop}:\n"
+        err_msg_len = f"{prop} of the two `TensorBlock` have not the same lenght."
+        err_msg_1 = f"{prop} are not the same or not in the same order."
+        err_msg_names = f"{prop} names are not the same or not in the same order."
         if prop == "samples":
+            if not len(a.samples) == len(b.samples):
+                raise ValueError(err_msg + err_msg_len)
             if not a.samples.names == b.samples.names:
-                raise ValueError(err_msg)
+                raise ValueError(err_msg + err_msg_names)
             if not np.all(a.samples == b.samples):
-                raise ValueError(err_msg)
+                raise ValueError(err_msg + err_msg_1)
         elif prop == "properties":
+            if not len(a.properties) == len(b.properties):
+                raise ValueError(err_msg + err_msg_len)
             if not a.properties.names == b.properties.names:
-                raise ValueError(err_msg)
+                raise ValueError(err_msg + err_msg_names)
             if not np.all(a.properties == b.properties):
-                raise ValueError(err_msg)
+                raise ValueError(err_msg + err_msg_1)
         elif prop == "components":
             if len(a.components) != len(b.components):
-                raise ValueError(err_msg)
+                raise ValueError(err_msg + err_msg_len)
 
-            for ca in a.components:
-                flag = False
-                for cb in b.components:
-                    if ca.names == cb.names:
-                        flag = True
-                        if not np.all(ca == cb):
-                            raise ValueError(err_msg)
-                # if flag is False ca.names == cb.names has never been satisfied
-                if not flag:
-                    raise ValueError(err_msg)
+            for c1, c2 in zip(a.components, b.components):
+                if not (c1.names == c2.names):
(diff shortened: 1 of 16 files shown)


jwa7:
``_check_maps`` and ``_check_blocks`` seem like such fundamental functions that form the basis of most of our ``equistore.operations`` modules that I feel like they deserve their own dedicated and self-contained unit tests. Of course there would be some repetition in the unit tests of the public functions that use them, like ``add`` and ``allclose`` etc. 

Is it just a rule we follow that we don't test private functions at all? That all testing of them come from the public functions that wrap them?

python/src/equistore/operations/_utils.py
```diff
@@ -48,58 +48,121 @@ def _check_blocks(a: TensorBlock, b: TensorBlock, props: List[str], fname: str):
...
+
+                err_msg_names = (
+                    f'gradient ("{parameter}") {prop} names are not the same '
+                    "or not in the same order."
+                )
+
+                if prop == "samples":
+                    if not len(grad_a.samples) == len(grad_b.samples):
+                        raise ValueError(err_msg + err_msg_len)
+                    if not grad_a.samples.names == grad_b.samples.names:
```
PicoCentauri, line 135: Do I see this correct that we check for exactly the same gradients. If the values are shuffled in one gradients wrt to the second we throw an error. For me this is fine, but we should state this in the docstring.

python/src/equistore/operations/_utils.py
```diff
@@ -48,58 +48,121 @@ def _check_blocks(a: TensorBlock, b: TensorBlock, props: List[str], fname: str):
...
+                        if not (c1.names == c2.names):
+                            raise ValueError(err_msg + err_msg_names)
+
+                        if not np.all(c1 == c2):
+                            raise ValueError(err_msg + err_msg_1)
+                elif prop == "parameters":
+                    # parameters are already checked at the begining but i want to give
+                    # the opportunity to the 'user' to use it, without problems
+                    continue
+                else:
```
PicoCentauri, line 160: I think we can go for an `!=` here and leave the docstring.

```suggestion
                elif prop != "parameters":
                    # parameters are already checked at the begining but i want to give
                    # the opportunity to the 'user' to use it, without problems
```

PicoCentauri (changes requested):
Good simplification here. I have some comments for the documentation and a possible simplification.


I would definetly vote for adding the public equal functions which will test these new functions properly.

python/src/equistore/operations/_utils.py
DavideTisi: Now these functions always check for the values and the order. All the operations that uses this check so far are implemented that need the order. Make them work non with the order is the next step

DavideTisi: i added a comment in the docstring

PicoCentauri (approved):
Fine for me now. Can you rebase @DavideTisi

DavideTisi:
did the rebase added the equal function

DavideTisi:
with that we surpassed 100 tests 🎉🎉

jwa7 (approved):
Looks good!

python/src/equistore/operations/_utils.py
```diff
@@ -48,58 +48,119 @@ def _check_blocks(a: TensorBlock, b: TensorBlock, props: List[str], fname: str):
...
+
+    If props is None it only checks if the ``'parameters'`` are consistent.
+
+    :param a: first :py:class:`TensorBlock` for check
+    :param b: second :py:class:`TensorBlock` for check
+    :param props: A list of strings containing the property to check.
+                 Allowed values are ``'samples'`` or ``'properties'``,
+                 ``'components'``. To check only if the ``'parameters'`` are consistent
+                 use ``props=None``, using ``'parameters'`` is allowed
+                  even though deprecated.
```
Luthaf, line 102: Why are we keeping around deprecated stuff? We don't have a release yet, we can break everything =)

python/src/equistore/operations/equal.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def equal(tensor1: TensorMap, tensor2: TensorMap, only_metadata: bool = False) -> bool:
+    """Compare two :py:class:`TensorMap`.
+
+    This function returns ``True`` if the two tensors have the same keys
+    (potentially in different order) and all the :py:class:`TensorBlock` have
+    the same (and in the same order) samples, components, properties,
+    and their values matrices pass the == test with the provided
+    ``rtol``, and ``atol``.
```
Luthaf, line 16: There is no `atol`/`rtol` in this function parameters

python/src/equistore/operations/_utils.py
DavideTisi: deprecated is maybe not the correct word, the idea is that I wanted to have both behavior: `props=None` -> checks the ``parameters`` and  the possibility to not break the code for someone that writes: `props=['samples',... whatever...,'parameters']`, thinking that `parameters` is needed to check the parameters, which are actually always checked.

python/src/equistore/operations/equal.py
DavideTisi: it is a copy-paste error

DavideTisi: i will correct it thanks

python/src/equistore/operations/_utils.py
Luthaf: I would raise an error if someone passes `parameters`, and make it clear in the doc that they are always checked

Issue #101, opened 2023-01-20
Missing operations transpose, swapaxis and basic arithmetic operation


agoscinski:

agoscinski:
I think we can generalize all functions that reduce the shape and functions that keep the shape fixed. We already have reduce_over_samples_block, but I think this can be generalized even more.

- *operation_name*: str 
- *parameters* : list of strings,  allows an application of the operation on a subset of the TB + GTBs, by default it does the operation on  all of them, so it is by default ['values'] + [parameter for paramater, _ in tensor_block.gradients()]
- *axis*: list of str, can contain the strings 'samples', 'components', 'properties' , by default ['samples', 'components', 'properties']
- *axis_names*: list of str, the names within the axis over which the reduction should be applied on, by default all names in the axis are used

and operations like *add* that do not change the shape do not need the last two paramaters

Luthaf:
> allows an application of the operation on a subset of the TB + GTBs, by default it does the operation on all of them, so it is by default ['values'] + [parameter for paramater, _ in tensor_block.gradients()]

We should never apply operations only on values or only on gradients, otherwise the values and gradients are no longer in sync (and the "gradients" are not actually gradients of the values). Doing this inside a function is fine, as long as when we give back the data to the user it is fully consistent.

>  but I think this can be generalized even more.

I don't think it is worth generalizing too much in the user-facing API, this would be very easily confusing. You mention `reduce_over_samples_block`, but this is an internal function that end users are not supposed to see. I prefer adding functions as we need them, with an API that make sense for this function. 

For example, `transpose` would be fine, and invert samples with properties and components within themself. But `swapaxis` only makes sense if applied between two components, or between samples and properties.

In general, not all numpy operations have a direct equivalent in equistore, because we are imposing more structure and meaning than a pure numpy array.

agoscinski:
Right, the consistency is important. I think I posted here too much of a brain dump here, sorry for that. I thought about functions that might be used for standardizing features, but not all are needed. I think basic arithmetic would already make this possible with the current reduce function.

> For example, transpose would be fine, and invert samples with properties and components within themself

Doesn't it result in the loss of the meaning of the gradient, because the gradient is the derivative of the properties wrt. the parameter?

ceriottm:
> Doesn't it result in the loss of the meaning of the gradient, because the gradient is the derivative of the properties wrt. the parameter?

[...]

PR #122, opened 2023-02-11
Ops/div min pow

DavideTisi:
added function for divide, subtract and pow #101 .
All of them compute the gradient with the relative formulas.
pow works only for numerical powers, it has test for positive integers and  power 0.5 (which gives results consistent with np.sqrt)

### DIFF
--- a/python/src/equistore/operations/add.py
+++ b/python/src/equistore/operations/add.py
@@ -13,12 +13,12 @@ def add(A: TensorMap, B: Union[float, TensorMap]) -> TensorMap:
 
     If gradients are present in ``A``:
 
-    #.  ``B`` is a scalar:
+    *  ``B`` is a scalar:
 
        .. math::
             \nabla(A + B) = \nabla A
 
-    #. ``B`` is a :py:class:`TensorMap` with the same metadata of ``A``:
+    * ``B`` is a :py:class:`TensorMap` with the same metadata of ``A``:
 
        .. math::
             \nabla(A + B) = \nabla A + \nabla B
--- a/python/src/equistore/operations/add.py
+++ b/python/src/equistore/operations/add.py
@@ -6,12 +6,21 @@ from ._utils import _check_blocks, _check_maps, _check_same_gradients_components
 
 
 def add(A: TensorMap, B: Union[float, TensorMap]) -> TensorMap:
-    """Return a new :class:`TensorMap` with the values being the sum of ``A`` and ``B``.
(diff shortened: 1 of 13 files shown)


python/src/equistore/operations/add.py
```diff
@@ -6,12 +6,22 @@
...
+    If gradients are present in ``A``:
+
+    *  ``B`` is a scalar:
+
+       .. math::
+            \nabla(A + B) = \nabla A
+
+    * ``B`` is a :py:class:`TensorMap` with the same metadata of ``A``:
+
+       .. math::
```
PicoCentauri, line 23: I always like explanation in docstrings :-)!

python/src/equistore/operations/subtract.py
```diff
@@ -0,0 +1,42 @@
...
+                    -B.block(key).gradient(parameter).data[:]
+                )
+        tensor_result = add(A=A, B=B)
+    else:
+        # check if can be converted in float (so if it is a constant value)
+        try:
+            float(B)
+        except TypeError as e:
+            raise TypeError("B should be a TensorMap or a scalar value. ") from e
+        tensor_result = add(A=A, B=-B)
```
PicoCentauri, line 40: Might be a bit simpler..

```suggestion
    else:
        # check if can be converted in float (so if it is a constant value)
        try:
            B = -float(B)
        except TypeError as e:
            raise TypeError("B should be a TensorMap or a scalar value. ") from e
     tensor_result = add(A=A, B=B)
```

python/src/equistore/operations/subtract.py
```diff
@@ -0,0 +1,42 @@
...
+    """Return a new :class:`TensorMap` with the values being the subtract
+    of ``A`` and ``B``.
+
+    If ``B`` is a :py:class:`TensorMap` it has to have the same metadata as ``A``.
+
+    If gradients are present in ``A`` a subtract is only performed if ``B`` is
+    a :py:class:`TensorMap` as well.
+
+    :param A: First :py:class:`TensorMap` for the subtraction.
+    :param B: Second instance for the subtraction. Parameter can be a scalar or a
```
PicoCentauri, line 18: Maybe add here the same math explanation as for `add`. It is trivial but for consistency we should do it.

PicoCentauri (changes requested):
Thanks @DavideTisi! More operations 🥳

PicoCentauri (approved):
Thanks for the changes @DavideTisi

Issue #118, opened 2023-02-10
Broadcasted TensorMap/TensorBlock x scalar operations (add, minus, multiplication, division)


agoscinski:
Because issue #101 is too large as one feature request, I split it up into chunks. This is the first chunk.

So we have this use case
```python
# A is TensorMap
a = 10
A = add(A, a)
```

lets say we use an add function for this, then it would look like

```python
def add(A, a):
    _A = copy(A)
    for key, block in _A.blocks():
        block.values += a
        for key, grad in block.gradients():
            grad.data += a
    return _A
```

### QUESTION/PROBLEM I think one conceptual problem is how should we differ between a operation considering gradient and just broadcasting operation as shown above

So the above example would be fully broadcasting operation and the below one would recognize the gradients
```python
def add(A, a):
    _A = copy(A)
    for key, block in _A.blocks():
        block.values += a
        # no operation applied on gradients since ∇(A+a) = ∇A
    return _A
```

Maybe this can be solved with an additional argument `gradient_operation`

```python
def add(A, a, gradient_operation=False):
    _A = copy(A)
    for key, block in A.blocks():
        block.values += a
        if not(gradient_operation):
            for key, grad in block.gradients():
                grad.data += a
    return _A
```

### QUESTION/PROBLEM how to do inplace operations?

The above ignores inplace operations, which would be required at some point, but I don't fully understand how numpy's inplace operation work
```python
A += 10 # seems to be inplace
B = A + 10 # not this inplace
A = A + 10 # is this inplace? I think not
np.add(A, 1, out= A) # seems to also not be inplace, but works different than the above ones
```
in the end we would like something as `add(A, scalar, out=A)`, to be inplace.

Luthaf:
The "just broadcasting operation" makes no sense, because it result in a TensorMap where the `gradients` are no longer the actual gradients of the `values`, breaking an invariant of TensorMap. I don't think we should provide these, except if you have a really compelling case where you need them?

> The above ignores inplace operations

All of your examples are doing inplace operations, modifying the input tensormap instead of making a copy.

DavideTisi:
So, I think that breaking the relation between values and gradient is a bit against the main idea of equistore Which keep together some quantities and its gradient. 
To answer your first question the idea for doing that operation would be to create a new TensorBlock equal to the gradient block of `A` (let's call it `gA`) and then do the add operation on `gA` in this way you have 2 blocks which are not dependent to each other anymore.

[...]

PR #123, opened 2023-02-13
added gradient_to_block

DavideTisi:
In order to address the conversation of #118. One idea is to create a function like ``gradient_to_block``:

Return a tuple containing a``TensorMap`` and a ``dictionary``. The first ``TensorMap`` is the input ``tensor`` where you removed the selected gradients; the second is a dictionary with as keys the ``parameters`` of the removed gradients and as values as ``TensorMap`` with the same keys as ``tensor`` and ``TensorBlock`` equal to the removed ``Gradients``.
In this way you separate the gradients and the values in different ``TensorMap`` and then you can apply all the operations you want separately clearly understanding that the relation values->gradient do not hold anymore.
The function is a small modification of remove_gradient.

The PR is not complete but what do you think?

### DIFF
--- a/python/src/equistore/operations/gradient_to_block.py
+++ b/python/src/equistore/operations/gradient_to_block.py
@@ -1,0 +1,62 @@
+from typing import List, Optional, Tuple
+
+from equistore import TensorBlock, TensorMap
+
+
+def gradient_to_block(
+    tensor: TensorMap,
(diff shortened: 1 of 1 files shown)


Luthaf:
Also see #219 which would solve this issue by making gradients actual blocks

Luthaf:
This was solved in a different way as part of #253

PR #125, opened 2023-02-15
New operation ``unique_metadata``

jwa7:
Re-do of PR #117, will close branch 'unique'. This PR introduces the operations ``unique_metadata`` and ``unique_metadata_block`` with a more descriptive name than in PR #117 and while staying inside ``equistore.operations``, with documentation living inside the 'operations/set' sub-class of operations.

**Usage**

To find the unique structure indices of the samples metadata of the blocks of a TensorMap:
```py
from equistore import io
from equistore.operations import unique_metadata

# Load a TensorMap - i.e. SOAP with 10 structures (indices 0 -> 9 inc.)
tensor = io.load("/path/to/tensor.npz")
unique_structures = unique_metadata(tensor, axis="samples", names=["structure"])
unique_structures
>>> Labels([(0,), (1,), (2,), (3,), (4,), (5,), (6,), (7,), (8,), (9,)], dtype=[('structure', '<i4')])
```

For the gradient blocks to find, for instance, the samples named "atom" for the "positions" gradients:

```py
unique_grad_atoms = unique_metadata(tensor, axis="samples", names=["atom"], gradient_param="positions")
unique_grad_atoms
>>> Labels([( 0,), ( 1,), ( 2,), ( 3,), ( 4,), ( 5,), ( 6,), ( 7,), ( 8,), ( 9,), (10,)], dtype=[('atom', '<i4')])
```

### DIFF
--- a/python/src/equistore/operations/unique_metadata.py
+++ b/python/src/equistore/operations/unique_metadata.py
@@ -38,17 +38,44 @@ def unique_metadata(
 
     .. code-block:: python
 
+        from equistore.operations import unique_metadata
+
         unique_structures = unique_metadata(
             tensor, axis="samples", names=["structure"],
         )
 
-    To find the unique ``"atom"`` indices in the ``"samples"`` metadata present
+    Or, to find the unique ``"atom"`` indices in the ``"samples"`` metadata present
     in the ``"positions"`` gradient blocks of a given :py:class:`TensorMap`:
 
     .. code-block:: python
 
         unique_grad_atoms = unique_metadata(
             tensor, axis="samples", names=["atom"], gradient_param="positions",
+        )
+
+    The unique indices can then be used to split the :py:class:`TensorMap`
+    into several smaller :py:class:`TensorMap` objects. Say, for example, that
+    the ``unique_structures`` from the example above are:
+
+    .. code-block:: python
+
+        Labels(
+            [(0,), (1,), (2,), (3,), (4,), (5,), (6,), (7,), (8,), (9,)],
+            dtype=[('structure', '<i4')],
+        )
+
+    Then, the following code will split the :py:class:`TensorMap` into 2
+    :py:class:`TensorMap` objects, with first containing structure indices 0-3
+    and the second containing structure indices 4-9:
+
+    .. code-block:: python
+
+        from equistore.operations import split
+
+        [tensor_1, tensor_2] = split(
+            tensor,
+            axis="samples",
+            grouped_idxs=[unique_structures[:4], unique_structures[4:]]
         )
 
     :param tensor: the :py:class:`TensorMap` to find unique indices for.
@@ -170,21 +197,27 @@ def _unique_from_blocks(
     Finds the unique metadata of a list of blocks along the given ``axis`` and
     for the specified ``names``. If ``gradient_param`` is specified, only finds
     the unique indices for gradient blocks under the specified parameter name.
+    If the full list of names is not present in the metadata of any block, an
+    empty :py:class:`Labels` object is returned.
     """
     # Extract indices from each block
     all_idxs = []
     for block in blocks:
-        if axis == "samples":
-            idxs = block.samples[names]
-        else:  # "properties"
-            idxs = block.properties[names]
-        all_idxs += idxs.tolist()
+        try:
+            if axis == "samples":
+                idxs = block.samples[names]
+            else:  # "properties"
+                idxs = block.properties[names]
+            all_idxs += idxs.tolist()
+        except KeyError:
+            # If the block does not have the specified metadata, skip it
+            continue
 
     # If no matching indices across all blocks return a empty Labels w/ the
     # correct names
     if len(all_idxs) == 0:
         # Create Labels with single entry
-        labels = Labels(names=names, values=np.arange(len(names)).reshape(-1, 1))
+        labels = Labels(names=names, values=np.arange(len(names)).reshape(1, -1))
         return labels[:0]  # Slice to zero length
 
     # Define the unique and sorted indices
@@ -250,14 +283,19 @@ def _check_args(
         )
     # Check names
     if not isinstance(names, list):
-        raise TypeError("`names` must be a `list` of `str`")
+        raise TypeError(f"`names` must be a `list` of `str`, received {type(names)}")
+    if not np.all([isinstance(name, str) for name in names]):
+        raise TypeError(
+            "`names` must be a `list` of `str`, received"
+            + f" {[type(name) for name in names]}"
+        )
     # Assumes samples/properties names are the same for every block in blocks
-    block_names = (
-        blocks[0].samples.names if axis == "samples" else blocks[0].properties.names
-    )
-    for name in names:
-        if name not in block_names:
-            raise ValueError(
-                f"the block(s) passed must have {axis} names that match those"
-                + f" passed in `names`. {names} were passed, {block_names} found."
-            )
+    # block_names = (
+    #     blocks[0].samples.names if axis == "samples" else blocks[0].properties.names
+    # )
+    # for name in names:
+    #     if name not in block_names:
+    #         raise ValueError(
+    #             f"the block(s) passed must have {axis} names that match those"
+    #             + f" passed in `names`. {names} were passed, {block_names} found."
+    #         )
--- a/python/src/equistore/operations/unique_metadata.py
+++ b/python/src/equistore/operations/unique_metadata.py
@@ -42,8 +42,8 @@ def unique_metadata(
             tensor, axis="samples", names=["structure"],
         )
 
-    To find the unique "atom" indices in the ``"samples"`` metadata present in
-    the ``"positions"`` gradient blocks of a given :py:class:`TensorMap`:
+    To find the unique ``"atom"`` indices in the ``"samples"`` metadata present
+    in the ``"positions"`` gradient blocks of a given :py:class:`TensorMap`:
 
     .. code-block:: python
 
@@ -117,8 +117,8 @@ def unique_metadata_block(
             block, axis="samples", names=["structure"],
         )
 
-    To find the unique ``"atom"`` indices along the "samples" axis present in
-    the ``"positions"`` gradient block of a given :py:class:`TensorBlock`:
+    To find the unique ``"atom"`` indices along the ``"samples"`` axis present
+    in the ``"positions"`` gradient block of a given :py:class:`TensorBlock`:
 
     .. code-block:: python
 
--- a/python/src/equistore/operations/unique_metadata.py
+++ b/python/src/equistore/operations/unique_metadata.py
@@ -18,22 +18,23 @@ def unique_metadata(
 ) -> Labels:
     """
     Returns a :py:class:`Labels` object containing the unique metadata across
-    all blocks of the input :py:class:`TensorMap`  ``tensor``, for the specified
-    ``axis`` (either "samples" or "properties") and metatdata ``names``.
+    all blocks of the input :py:class:`TensorMap`  ``tensor``. Unique Labels are
+    returned for the specified ``axis`` (either ``"samples"`` or
+    ``"properties"``) and metatdata ``names``.
 
-    Passing ``gradient_param`` as a str corresponding to a gradient parameter
-    (for instance "cell" or "positions") returns the unique indices only for the
-    gradient blocks associated with each block in the input ``tensor``,
-    according to the specified ``axis`` and ``names``. Note that gradient blocks
-    by definition have the same properties metadata as their parent
+    Passing ``gradient_param`` as a ``str`` corresponding to a gradient
+    parameter (for instance ``"cell"`` or ``"positions"``) returns the unique
+    indices only for the gradient blocks. Note that gradient blocks by
+    definition have the same properties metadata as their parent
     :py:class:`TensorBlock`.
 
-    If there are no indices in the (gradient) blocks of ``tensor`` corresponding
-    to the specified ``axis`` and ``names``, an empty Labels object with the
-    correct names as in the passed ``names`` is returned.
+    An empty :py:class:`Labels` object is returned if there are no indices in
+    the (gradient) blocks of ``tensor`` corresponding to the specified ``axis``
+    and ``names``. This will have length zero but the names will be the same as
+    passed in ``names``.
 
-    For example, to find the unique "structure" indices in the "samples"
-    metadata present in a given TensorMap:
+    For example, to find the unique ``"structure"`` indices in the ``"samples"``
+    metadata present in a given :py:class:`TensorMap`:
 
     .. code-block:: python
 
@@ -41,8 +42,8 @@ def unique_metadata(
             tensor, axis="samples", names=["structure"],
         )
 
-    To find the unique "atom" indices in the "samples" metadata present in the
-    "positions" gradient blocks of a given TensorMap:
+    To find the unique "atom" indices in the ``"samples"`` metadata present in
+    the ``"positions"`` gradient blocks of a given :py:class:`TensorMap`:
 
     .. code-block:: python
 
@@ -51,15 +52,16 @@ def unique_metadata(
         )
 
     :param tensor: the :py:class:`TensorMap` to find unique indices for.
-    :param axis: a str, either "samples" or "properties", corresponding to the
-        axis along which the named unique indices should be found.
-    :param names: a str, list of str, or tuple of str corresponding to the
-        name(s) of the indices along the specified ``axis`` for which the unique
-        values should be found.
-    :param gradient_param: a str corresponding to the gradient parameter name
-        for the gradient blocks to find the unique indices for. If none
-        (default), the unique indices of the regular TensorBlocks will be
-        calculated.
+    :param axis: a ``str``, either ``"samples"`` or ``"properties"``,
+        corresponding to the ``axis`` along which the named unique indices
+        should be found.
+    :param names: a ``str``, ``list`` of ``str``, or ``tuple`` of ``str``
+        corresponding to the name(s) of the indices along the specified ``axis``
+        for which the unique values should be found.
+    :param gradient_param: a ``str`` corresponding to the gradient parameter
+        name for the gradient blocks to find the unique indices for. If none
+        (default), the unique indices of the regular :py:class:`TensorBlock`s
+        will be calculated.
 
     :return: a sorted :py:class:`Labels` object containing the unique metadata
         for the blocks of the input ``tensor`` or its gradient blocks for the
@@ -68,7 +70,7 @@ def unique_metadata(
     """
     # Parse input args
     if not isinstance(tensor, TensorMap):
-        raise TypeError("``tensor`` must be an equistore TensorMap")
+        raise TypeError("`tensor` must be an equistore `TensorMap`")
     names = (
         [names]
         if isinstance(names, str)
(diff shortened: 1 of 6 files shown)


python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+def unique_metadata_block(
+    block: TensorBlock,
+    axis: str,
+    names: Union[List[str], Tuple[str], str],
+    gradient_param: Optional[str] = None,
+) -> Labels:
+    """
+    For a given ``axis`` (either "samples" or "properties"), and for the given
+    samples/proeprties ``names``, returns a :py:class:`Labels` object of the
+    unique indices in the input :py:class:`TensorBlock` ``block``.
```
PicoCentauri, line 96: same here

PicoCentauri (commented):
I had a quick look and it looks good. Most importantly is that also @DavideTisi and @Luthaf agree on the name. I like `unique_metadata` because with this it is clear that it returns metadata in the form of a label.

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+def unique_metadata(
+    tensor: TensorMap,
+    axis: str,
+    names: Union[List[str], Tuple[str], str],
+    gradient_param: Optional[str] = None,
+) -> Labels:
+    """
+    Returns a :py:class:`Labels` object containing the unique metadata across
+    all blocks of the input :py:class:`TensorMap`  ``tensor``, for the specified
+    ``axis`` (either "samples" or "properties") and metatdata ``names``.
```
PicoCentauri, line 22: Maybe a bit shorter sentence

```suggestion
    Returns a :py:class:`Labels` object containing the unique metadata across
    all blocks of the input :py:class:`TensorMap`  ``tensor``. Unique Labels are
    returned for the specified ``axis`` (either "samples" or "properties") and
    metatdata ``names``.
```

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+
+    Passing ``gradient_param`` as a str corresponding to a gradient parameter
+    (for instance "cell" or "positions") returns the unique indices only for the
+    gradient blocks associated with each block in the input ``tensor``,
+    according to the specified ``axis`` and ``names``. Note that gradient blocks
+    by definition have the same properties metadata as their parent
+    :py:class:`TensorBlock`.
```
PicoCentauri, line 29:
```suggestion
    Passing ``gradient_param`` as a ``str`` corresponding to a gradient parameter
    (for instance ``"cell"`` or ``"positions"``) returns the unique indices only for the
    gradient blocks associated with each block in the input ``tensor``,
    according to the specified ``axis`` and ``names``. Note that gradient blocks
    by definition have the same properties metadata as their parent
    :py:class:`TensorBlock`.
```

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    Returns a :py:class:`Labels` object containing the unique metadata in the
+    input :py:class:`TensorBlock`  ``block``, for the specified ``axis`` (either
+    "samples" or "properties") and metatdata ``names``.
+
+    Passing ``gradient_param`` as a str corresponding to a gradient parameter
+    (for instance "cell" or "positions") returns the unique indices only for the
+    gradient block associated with the input ``block`` , according to the
+    specified ``axis`` and ``names``. Note that gradient blocks by definition
+    have the same properties metadata as their parent :py:class:`TensorBlock`.
```
PicoCentauri, line 102: Same as above.

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+) -> Labels:
+    """
+    Finds the unique metadata of a list of blocks along the given ``axis`` and
+    for the specified ``names``. If ``gradient_param`` is specified, only finds
+    the unique indices for gradient blocks under the specified parameter name.
+    """
+    # Extract indices from each block
+    all_idxs = []
+    for block in blocks:
+        idxs = block.samples[names] if axis == "samples" else block.properties[names]
```
PicoCentauri, line 174: for readibility I would split this:
```suggestion
        if axis == "samples":
            idxs = block.samples[names]
         else:
             idxs = block.properties[names]
```

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+        for idx in idxs:
+            all_idxs.append(idx)
```
PicoCentauri, line 176: Maybe something like?

```suggestion
        all_idxs += idxs.tolist()
```

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+
+    # If no matching indices across all blocks return a empty Labels w/ the
+    # correct names
+    if len(all_idxs) == 0:
```
PicoCentauri, line 180:
```suggestion
    if all_idxs:
```

also works but maybe your version is more clear.

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+        # Create Labels with single entry
+        labels = Labels(names=names, values=np.array([[i for i in range(len(names))]]))
```
PicoCentauri, line 182:
```suggestion
        labels = Labels(names=names, values=np.arange(len(names)).reshape(1,-1))
```

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+):
+    """Checks input args for :py:func:`unique_metadata` and
+    :py:func:`unique_metadata_block`."""
+    # Check tensors
+    if isinstance(tensor, TensorMap):
+        blocks = tensor.blocks()
+        # Check gradients
+        if gradient_param is not None:
+            if not isinstance(gradient_param, str):
+                raise TypeError("``gradient_param`` must be a str")
```
PicoCentauri, line 207: since this will not be rendered with rst we can skip the double backticks  \`\`
```suggestion
                raise TypeError("`gradient_param` must be a str")
```

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+            # Check all blocks have a gradient under the passed param
+            if not np.all([block.has_gradient(gradient_param) for block in blocks]):
+                raise ValueError(
+                    "not all input blocks have a gradient under the"
+                    + f" passed ``gradient_param`` {gradient_param}"
```
PicoCentauri, line 212:
```suggestion
                    + f" passed `gradient_param` {gradient_param}"
```

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+                )
+            blocks = [
+                block.gradient(gradient_param) for block in blocks
+            ]  # redefine blocks
+    elif isinstance(tensor, TensorBlock):
+        blocks = [tensor]
+        # Check gradients
+        if gradient_param is not None:
+            if not isinstance(gradient_param, str):
+                raise TypeError("``gradient_param`` must be a str")
```
PicoCentauri, line 222:
```suggestion
                raise TypeError("`gradient_param` must be a str")
```

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+            # Check block has a gradient under the passed param
+            if not tensor.has_gradient(gradient_param):
+                raise ValueError(
+                    "input block does not have a gradient under the"
+                    + f" passed ``gradient_param`` {gradient_param}"
```
PicoCentauri, line 227:
```suggestion
                    + f" passed `gradient_param` {gradient_param}"
```

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+                )
+            blocks = [tensor.gradient(gradient_param)]  # redefine blocks
+    # Check axis
+    if not isinstance(axis, str):
+        raise TypeError("``axis`` must be a str, either 'samples' or 'properties'")
```
PicoCentauri, line 232:
```suggestion
        raise TypeError("`axis` must be a str, either 'samples' or 'properties'")
```

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+    if axis not in ["samples", "properties"]:
+        raise ValueError("``axis`` must be passsed as either 'samples' or 'properties'")
```
PicoCentauri, line 234:
```suggestion
        raise ValueError("`axis` must be passsed as either 'samples' or 'properties'")
```

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,245 @@
...
+    # Check names
+    if not isinstance(names, list):
+        raise TypeError("``names`` must be a list of str")
```
PicoCentauri, line 237:
```suggestion
        raise TypeError("`names` must be a list of str")
```

python/tests/operations/unique_metadata.py
```diff
@@ -0,0 +1,588 @@
...
+            self.tensor1.block(3), "properties", "properties"
+        )
+        self.assertTrue(
+            fn._utils._labels_equal(target_samples, actual_samples, exact_order=True)
+        )
+        self.assertTrue(
+            fn._utils._labels_equal(
+                target_properties, actual_properties, exact_order=True
+            )
+        )
```
PicoCentauri, line 50: Can you maybe split this into individual test functions. Preferably each unittests only has one assert which can fail. I know that this sometimes hard but this helps debugging a lot.

python/tests/operations/unique_metadata.py
```diff
@@ -0,0 +1,588 @@
...
+        actual_properties = fn.unique_metadata_block(
+            self.tensor1.block(0), "properties", "properties", parameter
+        )
+        self.assertTrue(
+            fn._utils._labels_equal(target_samples, actual_samples, exact_order=True)
+        )
+        self.assertTrue(
+            fn._utils._labels_equal(
+                target_properties, actual_properties, exact_order=True
```
PicoCentauri, line 148: same here

PicoCentauri (changes requested):
Very nice and useful function! I put some ideas for the code but don't take them as granted.  

One comment for the docstring: Try to shorten your sentences. They are sometimes super long that it is not clear what your talking and referring to. 

One comment for the tests. Try to test things individually in seperate functions and explain what are you testing. You can also use docstrings if you like. i.e. 

```python
def test_unique_indices_block_properties(self):
 
def test_unique_indices_block_samples(self):
```

For it helps that I think what combination of input paramaters I am expecting.

Do you have tests for returning empty Labels?

jwa7:
Thanks for the thorough review @PicoCentauri, I have now made those changes requested above.

jwa7:
Docstring errors fixed!

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,253 @@
...
+    if not isinstance(tensor, TensorMap):
+        raise TypeError("`tensor` must be an equistore `TensorMap`")
+    names = (
+        [names]
+        if isinstance(names, str)
+        else (list(names) if isinstance(names, tuple) else names)
+    )
+    _check_args(tensor, axis, names, gradient_param)
+    # Make a list of the blocks to find unique indices for
+    if gradient_param is None:
```
DavideTisi, line 81: I do not really understand why you want to pass to _unique_from_blocks, a list of blocks and not directly the tensor. 
- Passing the list force you to make copies (I think but I am not sure that what you are doing here in this if..else statement does a copy but maybe not ) of the blocks, which might be a bit big.
- The list makes you treat differently gradients and blocks while you could just pass the whole TensorMap with all the TensorBlocks and have a flag to make you work on the gradient. This would implies maybe more checks which are avoided here 
- Passing the TensorMap would avoid to do again some sanity checks, e.g. line 246-251 you are checking if the blocks have the same properties which would be guaranteed if the input is a TensorMap.
- Passing the list of block gives a bit of flexibility, for example you can pass blocks from different TensorMaps, but this feature is not really implemented since the input of the public functions is a single TensorMap.

This comment, actually, is embedded in a more general thought about the function that maybe need some discussion. To me it is not really clear what are the use cases, and why you want to use that, thus my comments might be a bit generic and not on point.
From the docstring I understand how the function works but not what are the usecases, I understand that the docstring is not the place where to put usage examples, but they would help me to do a better review.

For the rest the code is pretty clear.

DavideTisi (commented):
I repost the general comment here:

To me it is not really clear what are the use cases, and why you want to use that, thus my comments might be a bit generic and not on point.
From the docstring I understand how the function works but not what are the usecases, I understand that the docstring is not the place where to put usage examples, but they would help me to do a better review.

For the rest, the code is pretty clear very good job @jwa7 .

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,253 @@
...
+                    "input block does not have a gradient under the"
+                    + f" passed `gradient_param` {gradient_param}"
+                )
+            blocks = [tensor.gradient(gradient_param)]  # redefine blocks
+    # Check axis
+    if not isinstance(axis, str):
+        raise TypeError("`axis` must be a `str`, either `'samples'` or `'properties'`")
+    if axis not in ["samples", "properties"]:
+        raise ValueError(
+            "`axis` must be passsed as either `'samples'` or `'properties'`"
```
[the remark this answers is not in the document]
jwa7: Good spot, fixed now @DavideTisi

python/src/equistore/operations/unique_metadata.py
jwa7: Thanks for your comments. 

The reason that the base private function `_unique_from_blocks()` takes a list of blocks is so that it works regardless of whether you pass a list of normal TensorBlocks or Gradient TensorBlocks. It also means that it can be the base function for both `unique_metadata()` and `unique_metadata_block()`, as you can pass the function `blocks=tensor.blocks()` or `[block.gradient(gradient_param) for block in tensor.blocks()]` for the former and `blocks=[block]` or `[block.gradient(gradient_param)]` for the latter. In short, the fundamental operation in the module is to extract the metadata form blocks and find the unique values; the private function `_unique_from_blocks` does this, and the public functions just wrap this for the user. 

I suppose, as you say, the function could handle the input type internally, but to me this would be the same code just moved from the public functions to the private function. You say "This would implies maybe more checks which are avoided here" - what kind of checks do you think are missing by doing it this way round?

As for use cases, the most obvious one is for performing splitting of TensorMaps. If I want to use the `split()` function to divide my TensorMap into smaller TensorMaps, I need to know what the unique values of a given metadata index along a given axis is. For instance:

```py
from equistore.operations import unique_metadata, split

tensor = io.load("/path/to/tensor/")

# This is the example code in the docstring of `unique_metadata()`
unique_structures = unique_metadata(tensor, axis="samples", names=["structure"])
unique_structures
>>> Labels([(0,), (1,), (2,), (3,), (4,), (5,), (6,), (7,), (8,), (9,)], dtype=[('structure', '<i4')])

# Shuffle these indices
np.random.shuffle(unique_structures)
unique_structures
>>> Labels([(5,), (8,), (2,), (1,), (4,), (3,), (0,), (7,), (6,), (9,)], dtype=[('structure', '<i4')])

# Split into 2 new TensorMaps based on these unique indices
[tensor_1, tensor_2] = split(
    tensor, axis="samples", grouped_idxs=[unique_structures[:5], unique_structures[5:]]
)

# Verify structure indices in the 1st split tensor
unique_metadata(tensor_1, axis="samples", names=["structure"])
>>> Labels([(5,), (8,), (2,), (1,), (4,)], dtype=[('structure', '<i4')])

# Verify structure indices in the 2nd split tensor
unique_metadata(tensor_2, axis="samples", names=["structure"])
>>> Labels([(3,), (0,), (7,), (6,), (9,)], dtype=[('structure', '<i4')])
```

As presented here, this use of `unique_metadata()` and `split()` forms the foundation for a `split_data()` utility (used for train-test(-validation) splits) function in equisolve (see the `split_data` branch). I see `unique_metadata` as returning a characteristic intrinsic enough to a TensorMap that it belongs in equistore, as opposed to equisolve, for example.

As for your comments on the sanity checks in lines 246-251 - I take your point. The `blocks` it checks will only ever be either a list of length 1 or a list of blocks from a TensorMap. As you say, if the properties are guaranteed to be the same for all blocks, for this check I could remove the iteration over blocks here and just check the first block, i.e. `blocks[0]`. Does this work for you?

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,253 @@
...
+                    "input block does not have a gradient under the"
+                    + f" passed `gradient_param` {gradient_param}"
+                )
+            blocks = [tensor.gradient(gradient_param)]  # redefine blocks
+    # Check axis
+    if not isinstance(axis, str):
+        raise TypeError("`axis` must be a `str`, either `'samples'` or `'properties'`")
+    if axis not in ["samples", "properties"]:
+        raise ValueError(
+            "`axis` must be passed as either `'samples'` or `'properties'`"
```
DavideTisi, line 241:
```suggestion
            f"`axis` must be passed as either `'samples'` or `'properties'`, {axis} was passed."
```
or something like that here and in the other checks

python/src/equistore/operations/unique_metadata.py
DavideTisi: Anyway you answered my main questin so i think everything is pretty good

python/tests/operations/unique_metadata.py
```diff
@@ -0,0 +1,705 @@
...
+        with self.assertRaises(ValueError) as cm:
+            fn.unique_metadata_block(self.block, "ciao", ["structure"])
+        self.assertEqual(
+            str(cm.exception),
+            "`axis` must be passed as either `'samples'` or `'properties'`,"
+            + " ciao was passed",
+        )
+        # ValueError names not in block - samples
+        with self.assertRaises(ValueError) as cm:
+            fn.unique_metadata_block(self.block, "samples", ["ciao"])
```
DavideTisi, line 483: i like the italian ahahahaha

DavideTisi (approved):

python/tests/operations/unique_metadata.py
jwa7: Haha gotta sweeten the deal for my italian reviewers!

python/src/equistore/operations/unique_metadata.py
PicoCentauri: Can we put the example somewhere and the bring the answer you gave @jwa7 into the code? Discussions in the comments and appearing question might come up again if do not clarify this in source and the docs.

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,263 @@
...
+            idxs = block.samples[names]
+        else:  # "properties"
+            idxs = block.properties[names]
+        all_idxs += idxs.tolist()
+
+    # If no matching indices across all blocks return a empty Labels w/ the
+    # correct names
+    if len(all_idxs) == 0:
+        # Create Labels with single entry
+        labels = Labels(names=names, values=np.arange(len(names)).reshape(-1, 1))
```
PicoCentauri, line 187: Are we sure that this is correct with `reshape(-1, 1)`. I just tried the code and got and error. 

```python
from equistore import Labels

names = ["1", "l", "f"]
labels = Labels(names=names, values=np.arange(len(names)).reshape(-1, 1))
```

To get a propper Labels instance I had to swap to `reshape(1, -1)`:

```python
names = ["1", "l", "f"]
labels = Labels(names=names, values=np.arange(len(names)).reshape(1, -1))
```

PicoCentauri (changes requested):
Very good. I think we are almost there. I have a minor doubt about the empty Label. Everything else is fine!

python/src/equistore/operations/unique_metadata.py
jwa7: Sure, I'll add an example using `split()` into the docstring of `unique_metadata()`

python/src/equistore/operations/unique_metadata.py
jwa7: Yes, good spot! Something missed by my unit tests so I'll get a test in for that now too :)

jwa7:
I have updated the docstrings to include the code example using `split`, and also changes the behaviour of the function to not raise an error when passing ``names`` that aren't in the input tensor/block (with appropriate tests)

python/src/equistore/operations/unique_metadata.py
```diff
@@ -0,0 +1,301 @@
...
+    # Assumes samples/properties names are the same for every block in blocks
+    # block_names = (
+    #     blocks[0].samples.names if axis == "samples" else blocks[0].properties.names
+    # )
+    # for name in names:
+    #     if name not in block_names:
+    #         raise ValueError(
+    #             f"the block(s) passed must have {axis} names that match those"
+    #             + f" passed in `names`. {names} were passed, {block_names} found."
+    #         )
```
PicoCentauri, line 301: why is this commented out?

jwa7: Apologies, forgot to remove. It was an old input arg check that is no longer valid because we now do not raise an error if the names are not part of the input tensor. New commit removes this!

Issue #110, opened 2023-02-02
Allow `names` paramter in `Labels` to be a string or raise a more clear error message 


PicoCentauri:
## Expected behavior ##

When creating a `Label` instance with only one name I would write

```python
Labels(names="species", values=np.array([(1,)]))
```

and should get

```python
Labels([(1,)], dtype=[('species', '<i4')])
```

## Actual behavior ##

I get an `ValueError` raised by `labels.py` in line 75 stating that `names parameter must have an entry for each column of the array`. The reason is that the code is treating the string a list and assumes that there have to be also 7 columns in my array.

## Solution I would like ##

Either we convert single string inputs to list (as we do it in the operations) or we raise a more clear error message stating that the input type is wrong.

Additionally, I would rephrase the error message to something like `names parameter has 7 entries but the array only has 1 column(s).`

Luthaf:
fixed in #126

PR #126, opened 2023-02-15
Feat/label single names

DavideTisi:
simple fix for #110 

Now:
```
Labels(names="species", values=np.array([[1]]))
```

is equivalent to:

```
Labels(names=["species"], values=np.array([[1]]))
```

also:
```
Labels(names="", values=np.array([[]]))
```
is equivalent to:
```
Labels(names=[], values=np.array([[]]))
```

### DIFF
--- a/python/src/equistore/labels.py
+++ b/python/src/equistore/labels.py
@@ -60,23 +60,23 @@ class Labels(np.ndarray):
     def __new__(cls, names: Union[List[str], str], values: np.ndarray, **kwargs):
         """
         :param names: names of the variables in the new labels, in the case of a single
-                      name also a single string can be given: ``names = ["name"]``
-                      is equivalent to ``names = "name"``
+                      name also a single string can be given: ``names = "name"``
         :param values: values of the variables, this needs to be a 2D array of
             ``np.int32`` values
         """
-        if isinstance(names, str):
-            if len(names) == 0:
-                names = []
-            else:
-                names = [names]
         if not isinstance(values, np.ndarray):
             raise ValueError("values parameter must be a numpy ndarray")
 
         if len(values.shape) != 2:
             raise ValueError("values parameter must be a 2D array")
 
-        names = tuple(str(n) for n in names)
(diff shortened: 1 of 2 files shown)


python/src/equistore/labels.py
```diff
@@ -57,12 +57,19 @@ class Labels(np.ndarray):
...
+                      name also a single string can be given: ``names = ["name"]``
+                      is equivalent to ``names = "name"``
         :param values: values of the variables, this needs to be a 2D array of
             ``np.int32`` values
         """
+        if isinstance(names, str):
+            if len(names) == 0:
+                names = []
+            else:
+                names = [names]
```
PicoCentauri, line 72: I would put this above L79 where we reformat names. Maybe one can even combine the logic you added with

```python
names = tuple(str(n) for n in names)
```

python/src/equistore/labels.py
```diff
@@ -57,12 +57,19 @@ class Labels(np.ndarray):
             ...
     """
 
-    def __new__(cls, names: List[str], values: np.ndarray, **kwargs):
+    def __new__(cls, names: Union[List[str], str], values: np.ndarray, **kwargs):
         """
-        :param names: names of the variables in the new labels
+        :param names: names of the variables in the new labels, in the case of a single
+                      name also a single string can be given: ``names = ["name"]``
```
PicoCentauri, line 63:
```suggestion
        :param names: names of the variables in the new labels, in the case of a single
                      name also a single string can be given: ``names = "name"``
```

PicoCentauri (changes requested):
Very good. I have two things that might increase the readability of the code and the docstring.

python/src/equistore/labels.py
DavideTisi: do you like how it si now?

PicoCentauri: Yes. Nicely solved!

PicoCentauri (approved):

PR #131, opened 2023-02-20
Downgrade manylinux

Luthaf:
This is a follow up to #111, trying to run on an older glibc/produce wheels compatible with more linux distributions.


PicoCentauri:
I am fine with using manylinux2010 again, even though this is legacy. We could least should go with the [recent tag name](https://peps.python.org/pep-0600/#legacy-manylinux-tags) `manylinux_2_12_x86_64 `?

PR #136, opened 2023-02-22
Use ReadtheDocs action for posting link to PR docs

PicoCentauri:
Read the docs [offers a GH new action](https://github.com/readthedocs/actions/tree/v1/preview) to post the link to the docs. This simplifies our workflow a bit.

----
:books: Documentation preview :books:: https://equistore--136.org.readthedocs.build/en/136/

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -51,37 +51,7 @@ jobs:
           publish_dir: ./gh-pages/
           force_orphan: true
 
-      - name: post link to RTD
-        if: github.event_name == 'pull_request'
-        uses: actions/github-script@v6
+      - name: Post link to RTD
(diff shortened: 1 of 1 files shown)


PicoCentauri:
It is working. If we want we can also adjust the text, but I am fine with the default one.

Screenshot 2023-02-22 at 17 26 40

Luthaf:
Work for me! What would be really nice would be to adapt the post for wheels in the same way, but I'm not sure it can be done, since the link needs to be update for each CI run

Luthaf (approved):

PicoCentauri:
Maybe we can take a look how the action works. We could update the post between markers like the RTD is doing it:
`<!-- readthedocs-preview equistore start -->`

PR #141, opened 2023-02-24
Wrap operations as `TensorMap` and  `TensorBlock` methods

### LINKED ISSUE #40: Better error message when trying to assign to block.values directly
It could be useful to manipulate the descriptor values of a block after they have been calculated. However, expressions like 

```python
block.values *= -1
```

are not allowed an lead to `AttributeError: can't set attribute`. There is a workaround like 

```python
for r in block.values:
    r *= -1
```

which works fine, but in my view the first option should also work.

PicoCentauri:
Adding basic methods for math and manipulation operations on a `Tensormap` and `TensorBlock` class.
This will give users a lot more usability when working with `Tensormaps` and `TensorBlocks`.

The implementation here does not solve the discussion of #118. It only wraps the already implemented operations as class methods.

My approach for the imports is a bit hacky... I was not able to put the imports at the top of the file because of [circular import issues](https://stackoverflow.com/questions/7336802/how-to-avoid-circular-imports-in-python). We require the TensorBlock and TensorMap class in the operations but also the operations in the TensorMap and TensorBlock class. If somebody knows how to solve them, I am happy to learn.

We could think of adding the same basic math operations also for blocks. But, we can do this in an upcoming PR.

# TODO
- [x] Tests
- [x] Check docs rendering

----
:books: Documentation preview :books:: https://equistore--141.org.readthedocs.build/en/141/


python/src/equistore/tensor.py
```diff
@@ -105,6 +105,84 @@ def __getitem__(self, *args) -> TensorBlock:
...
+
+        return pow(self, other)
+
+    def __iadd__(self, other):
+        from equistore.operations import add
+
+        return add(self, other)
+
+    def __isub__(self, other):
+        from equistore.operations import add
```
DavideTisi, line 154: subtract

python/src/equistore/tensor.py
```diff
@@ -394,6 +460,146 @@ def property_names(self) -> List[str]:
...
+
+        :param samples_names:
+            names of samples to perform the standart deviation over
+        """
+        return self._reduce_over_samples(samples_names, "std")
+
+    def variance_over_samples(self, samples_names: List[str]) -> "TensorMap":
+        r"""Variance over the indices of the given ``samples_names``.
+
+        For addiotional details refer to the reference page of
```
DavideTisi, line 537: addiotional

PicoCentauri: Thanks @DavideTisi for always pointing out my typos 🤣

DavideTisi (approved):

python/src/equistore/tensor.py
```diff
@@ -105,6 +105,72 @@ def __getitem__(self, *args) -> TensorBlock:
...
+        from equistore.operations import divide
+
+        return divide(self, other)
+
+    def __pow__(self, other):
+        from equistore.operations import pow
+
+        return pow(self, other)
+
+    def __iadd__(self, other):
```
Luthaf, line 148: I would not do the in-place operations if they are not actually in-place

python/src/equistore/tensor.py
```diff
@@ -394,6 +460,146 @@ def property_names(self) -> List[str]:
...
+        :py:class:`TensorBlocks` corresponding to the same key must have the same
+        ``properties``. The resulting :py:class:`TensorBlocks` of the dot product of
+        two :py:class:`TensorBlocks` has ``result_block.values = block1.values @
+        block2.values.T``
+
+        :param other:
+            :py:class:`TensorMap` to multiply
+
+        :return: dot product
+        """
```
Luthaf, line 489: I don't love that we have to duplicate the documentation here

python/src/equistore/tensor.py
```diff
@@ -394,6 +460,146 @@ def property_names(self) -> List[str]:
...
+        For additional details refer to the reference page of
+        the :ref:`python-api-operations-sample-reduction` functions.
+
+        :param samples_names:
+            names of samples to sum over
+        """
+        return self._reduce_over_samples(samples_names, "sum")
+
+    def mean_over_samples(self, samples_names: List[str]) -> "TensorMap":
+        """Aaverage over the indices of the given ``samples_names`` .
```
DavideTisi, line 513: Aaverage

python/src/equistore/tensor.py
PicoCentauri: Yes, I agree.

python/src/equistore/tensor.py
PicoCentauri: We can do it numpy style and briefly state.

```
"""Computes the dot product with another :py:class:`TensorMap`.

Refer to :py:func:`equistore.dot` for full documentation.
"""
```

DavideTisi (approved):

Luthaf:
While I agree that this makes sense for the arithmetic operations (`__add__` and friends), I'm less sure about the other operations (`sum_samples`, …)

One issue is that while the core code is reasonably stable and could be made into a 0.1 release, the operations are still moving a lot. We could make a release soon by not including the operations in the set of stable functions (e.g. putting them in `equistore.unstable.XXX` until we are confident on the API), but exposing all of them as member functions of TensorMap does not allow us to select which ones we want to make stable.

PicoCentauri:
I see your point. I would be fine if for now we only add the arithmetic operations.

Once we agree on the operations I am definitely in favor of adding them as well.

PicoCentauri:
I removed the inplace operations. For me this looks still like a useful addition. Thoughts?

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -306,7 +309,9 @@ def mean_over_samples(tensor: TensorMap, samples_names: List[str]) -> TensorMap:
 
 
 def std_over_samples(tensor: TensorMap, samples_names: List[str]) -> TensorMap:
-    r"""Create a new :py:class:`TensorMap` with the same keys as
+    r"""Standard deviation over the indices of the given ``samples_names``.
```
Luthaf, line 312:
```suggestion
    r"""Compute the standard deviation of a :py:class:`TensorMap`, grouping samples according to ``samples_names``.
```

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -335,7 +340,9 @@ def std_over_samples(tensor: TensorMap, samples_names: List[str]) -> TensorMap:
 
 
 def variance_over_samples(tensor: TensorMap, samples_names: List[str]) -> TensorMap:
-    r"""Create a new :py:class:`TensorMap` with the same keys as
+    r"""Variance over the indices of the given ``samples_names``.
```
Luthaf, line 343:
```suggestion
    r"""Compute the variance of a :py:class:`TensorMap`, grouping samples according to ``samples_names``.
```

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -335,7 +340,9 @@ def std_over_samples(tensor: TensorMap, samples_names: List[str]) -> TensorMap:
...
+
+    Create a new :py:class:`TensorMap` with the same keys as
```
Luthaf, line 345:
```suggestion
    This function creates a new :py:class:`TensorMap` with the same keys as
```

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -306,7 +309,9 @@ def mean_over_samples(tensor: TensorMap, samples_names: List[str]) -> TensorMap:
 
 
 def std_over_samples(tensor: TensorMap, samples_names: List[str]) -> TensorMap:
-    r"""Create a new :py:class:`TensorMap` with the same keys as
+    r"""Standard deviation over the indices of the given ``samples_names``.
+
+    Create a new :py:class:`TensorMap` with the same keys as
```
Luthaf, line 314:
```suggestion
    This function creates a new :py:class:`TensorMap` with the same keys as
```

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -284,7 +285,9 @@ def sum_over_samples(
 
 
 def mean_over_samples(tensor: TensorMap, samples_names: List[str]) -> TensorMap:
-    """Create a new :py:class:`TensorMap` with the same keys as
+    """Mean over the indices of the given ``samples_names``.
+
+    Create a new :py:class:`TensorMap` with the same keys as
```
Luthaf, line 290:
```suggestion
    This function creates a new :py:class:`TensorMap` with the same keys as
```

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -219,7 +219,8 @@ def _reduce_over_samples(
 def sum_over_samples(
     tensor: TensorMap, samples_names: Union[List[str], str]
 ) -> TensorMap:
-    """
+    """Sum over the indices of the given ``samples_names``.
+
     Create a new :py:class:`TensorMap` with the same keys as as the input
```
Luthaf, line 224:
```suggestion
    This function creates a new :py:class:`TensorMap` with the same keys as as the input
```

python/src/equistore/operations/remove_gradients.py
```diff
@@ -7,14 +7,13 @@ def remove_gradients(
     tensor: TensorMap,
     remove: Optional[List[str]] = None,
 ) -> TensorMap:
-    """
-    Create a new :py:class:`TensorMap` without some or
-    all of the gradients from ``tensor``.
+    """Remove some or all of the gradients.
```
Luthaf, line 10:
```suggestion
    """Remove some or all of the gradients from a :py:class:`TensorMap`.
```

python/src/equistore/operations/slice.py
```diff
@@ -12,12 +12,11 @@ def slice(
     samples: Optional[Labels] = None,
     properties: Optional[Labels] = None,
 ) -> TensorMap:
-    """
-    Slices an input :py:class:`TensorMap` along the samples and/or properties
-    dimension(s). ``samples`` and ``properties`` are
-    :py:class:`Labels` objects that specify the samples/properties
-    (respectively) names and indices that should be sliced, i.e. kept in the
-    output tensor.
+    """Slice along the samples and/or properties dimension(s).
```
Luthaf, line 15:
```suggestion
    """Slice a :py:class:`TensorMap` along the samples and/or properties dimension(s).
```

python/src/equistore/operations/split.py
```diff
@@ -11,9 +11,9 @@ def split(
     axis: str,
     grouped_idxs: List[Labels],
 ) -> List[TensorMap]:
-    """
-    Splits an input :py:class:`TensorMap` into mutliple :py:class:`TensorMap`
-    objects based on some specified groups of indices, along either the
+    """Split into mutliple :py:class:`TensorMap`.
```
Luthaf, line 14:
```suggestion
    """Split a :py:class:`TensorMap` into multiple :py:class:`TensorMap`.
```

PicoCentauri:
I removed everything else then the basic operations. I have a local backup branch with everything. Once we decide to bring them in, we do not have to write everything again from scratch...

python/tests/tensor.py
```diff
@@ -334,3 +332,31 @@ def test_components_to_properties(self):
...
+        assert equistore.add(tensor, 1) == tensor + 1
+
+    def test_sub(self, tensor):
+        assert equistore.subtract(tensor, 1) == tensor - 1
+
+    def test_mul(self, tensor):
+        assert equistore.multiply(tensor, 2) == tensor * 2
+
+    def test_matmul(self, tensor):
+        tensor = tensor
```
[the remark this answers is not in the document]
PicoCentauri: Me neither, maybe it is leftover from the rebase...

PR #147, opened 2023-02-28
`equal_metadata` function

### LINKED ISSUE #146: A function like `equal(tensor, only_metadata=True)`, but with control over the metadata axes checked
A key step in the ML workflow is checking that the metadata between input and output TensorMaps is exactly equivalent in keys, and along the samples and components axes. However, the properties should not be checked, as the relationship between different input and output features is the thing being learned.

There should be a function, similar to `equal`, where the user can control which metadata Labels should be checked. Maybe this should be a modification to `equal`, where flags turn off checking of specified axes:

```py
from equistore.operations import equal
...
assert equal(tensor_1, tensor_2, only_metadata=True, do_not_check=["properties"])
```

or a new function completely:

```py
from equistore.operations import equal_metadata
...
assert equal_metadata(tensor_1, tensor_2, metadata_to_check=["keys", "samples", "components"])
```

jwa7:
Closes #146. A function to compare the metadata of 2 TensorMaps for exact equivalence. The idea is that this function will deprecate the `only_metadata` flag in the `equal()` function.

To do:
- [x] Write unit tests for `equal_metadata`
- [x] Modify `equal` to remove the `only_metdata` flag
- [x] Update the unit tests for `equal` to not test the `only_metdata` flag

----
:books: Documentation preview :books:: https://equistore--147.org.readthedocs.build/en/147/


python/src/equistore/operations/equal_metadata.py
```diff
@@ -0,0 +1,167 @@
...
+) -> bool:
+    """
+    Checks if two :py:class:`TensorMap` objects have the same metadata,
+    returning a bool.
+
+    Equivalence in the keys of the two :py:class:`TensorMap` objects is always
+    checked. In addition to this, users can control the metadata of the blocks
+    that is checked. If `check` is none (the default), all metadata is checked
+    for exact equivalence in values and order, i.e. the samples, components, and
+    properties of each block.
```
Luthaf, line 21:
```suggestion
    The equivalence of the keys of the two :py:class:`TensorMap` objects is always
    checked. If `check` is none (the default), all metadata (i.e. the samples, 
    components, and properties of each block) is checked to contain the 
    same values in the same order.
```

python/src/equistore/operations/equal_metadata.py
```diff
@@ -0,0 +1,169 @@
+"""
+Module for checking equivalence in metadata between 2 TensorMaps
+"""
+from typing import List, Optional
+
+from ..tensor import TensorMap
+from ._utils import _check_blocks, _check_maps, _check_same_gradients
```
Luthaf, line 7: I think it makes sense to move these function in this file, since they are very related. 

It might also be worth to make all functions use `equal_metadata` instead of `_check_XXX`, but this can be a future PR.

Luthaf (commented):
Looks good overall, I think we should update the code structure a bit (and maybe re-structure the metadata checks later, but this is another PR)

Luthaf (approved):

PR #152, opened 2023-03-01
Change linear system solver from numpy to scipy

agoscinski:
The scipy linear system solver is by default using the same LAPACK function under the hood (_gesv), but also allows to use more efficient and/or more stable routines in the case the matrix has certain properties (e.g. positive definitene, symettric, Hermitian). For the Ridge class in equisolve we actually use the positive definiteness of the similarity matrix (A= X^TX, b =X^Ty, solving Aw=b). So I propose to change the solver to the scipy one

----
:books: Documentation preview :books:: https://equistore--152.org.readthedocs.build/en/152/

### DIFF
--- a/python/src/equistore/operations/_dispatch.py
+++ b/python/src/equistore/operations/_dispatch.py
@@ -112,7 +112,7 @@ def dot(A, B):
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
-def solve(X, Y):
+def solve(X, Y, solver="numpy", lower=None, check_finite=True, assume_a="gen"):
     """
     Computes the solution of a square system of linear equations with a unique
     solution.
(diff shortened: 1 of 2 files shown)


agoscinski:
One test error is due to the missing dependency of scipy to equistore. Should we add this a dependency? I thought to rather add an additional argument `solver` where one can choose the scipy or numpy solver. I feel like this is a function where we want more flexibility in the future. But I am also okay to just add scipy as dependency and use its solver, since they should equal in results.
@jwa7 @DavideTisi @Luthaf

Luthaf:
I'm fine with adding scipy as a dependency, but we will need similar precision guarantee when using torch. This might mean we'll have to duplicate some of scipy logic ourself, and at this point we could also use the logic on top of numpy. 

Where are the differences between scipy and numpy documented?

agoscinski:
scipy and numpy use by default the same LAPACK routine gesv. scipy has one argument that gives better feedback as soon as something fails. I am not sure about the `overwrite_*` arguments, how numpy does here. So speedwise there are small differences because of these arguments. More importantly, scipy offers more solvers when the matrix has certain properties (e.g. positive definite in lapack called posv). The positive definiteness solver we need for Ridge in equisolve.

From a conversation with @Luthaf, it seems that the positive definite solver uses cholesky decomposition https://netlib.org/lapack/explore-html/dc/de9/group__double_p_osolve_ga9ce56acceb70eb6484a768eaa841f70d.html#ga9ce56acceb70eb6484a768eaa841f70d
So we can probably imitate this with torch using https://pytorch.org/docs/stable/generated/torch.cholesky_solve.html

Luthaf:
We should revisit this later if we still want it!

PR #160, opened 2023-03-03
Convenience functions for Labels

PicoCentauri:
If an empty array is given to the constructor of a `Labels` instance it is now automatically converted into the correct shape. This allows inputs like

```python
Labels(["sample", "structure", "atom"], np.array([]))
```

without manually reshaping as required before

```python
Labels(["sample", "structure", "atom"], np.zeros([0, 3], dtype=np.int32))
```

Also, a function Labels.empty() is added, allowing even simpler constructions like 

```python
Labels.empty(["sample", "structure", "atom"])
```

----
:books: Documentation preview :books:: https://equistore--160.org.readthedocs.build/en/160/

### DIFF
--- a/python/src/equistore/labels.py
+++ b/python/src/equistore/labels.py
@@ -69,12 +69,6 @@ class Labels(np.ndarray):
             if key != "_eqs_labels_t":
                 raise ValueError(f"unexpected kwarg to Labels: {key}")
 
-        if not isinstance(values, np.ndarray):
-            raise ValueError("values parameter must be a numpy ndarray")
-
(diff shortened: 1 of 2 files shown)


Luthaf (approved):

Luthaf:
Feel free to merge once CI is happy!

jwa7:
This is useful! The `Labels.empty()` function can be used in `unique_metadata()` when an empty Labels object needs to be returned.

PicoCentauri:
We only do useful stuff 🤣

PR #161, opened 2023-03-03
Allow to copy TensorMap

### LINKED ISSUE #98: copy function for tensor map
I think that would be just a simple helper function, because the copy function for tensor block already exists. So one needs to only create one loop. I am just not sure where this should be implemented. Python API or in the core?

Luthaf:
Fix #98

### DIFF
--- a/equistore-core/include/equistore.h
+++ b/equistore-core/include/equistore.h
@@ -531,6 +531,20 @@ struct eqs_tensormap_t *eqs_tensormap(struct eqs_labels_t keys,
 eqs_status_t eqs_tensormap_free(struct eqs_tensormap_t *tensor);
 
 /**
+ * Make a copy of an `eqs_tensormap_t`.
+ *
+ * The memory allocated by this function and the blocks should be released
+ * using `eqs_tensormap_free`.
+ *
+ * @param tensor existing tensor to copy
+ *
+ * @returns A pointer to the newly allocated tensor, or a `NULL` pointer in
+ *          case of error. In case of error, you can use `eqs_last_error()`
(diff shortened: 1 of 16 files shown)


agoscinski (commented):
I did not really understand the unwind safety idea in Rust, and I feel like I need much more time to fully understand it, but shouldn't there be some test for eqs_tensormap_copy, when it fails returning a null ptr? So far I only see a test when it works

Luthaf:
> shouldn't there be some test for eqs_tensormap_copy, when it fails returning a null ptr?

That's fair!

> I did not really understand the unwind safety idea in Rust

The core of it is when Rust code panics (or a C++ exception is thrown), the code starts unwinding the call stack until it reached a panic handler (`catch` block in C++), and while unwinding the code needs to clean up temporaries values and run their destructors.

This operation is only well defined within a single language, since e.g. Rust and C++ might have different semantics w.r.t. what happens during unwinding, so we need to catch panics/exceptions before handing back control to C in the C API implementation.

On top of that, a panic could leave a value in a half-initalized state, leading to hard to debug issues. So Rust does not allow to store data outside of a `catch_panic` call by the closure passed to `catch_panic`, unless this data is `UnwindSafe`, i.e. the data will always be valid. We can then use `std::panic::AssertUnwindSafe` to tell the compiler a given value is unwind safe, and write to it inside the `catch_panic` call.

Luthaf:
It looks like we are hitting https://github.com/readthedocs/actions/issues/23 on the CI

agoscinski (approved):
Looks good

PR #165, opened 2023-03-05
Port Python core tests to pytest

PicoCentauri:
This PR moves the Python core tests to `pytest`. The operations tests still use `unittests`. In addition, I added torch in test depencies for also running the data tests for torch. 

----
:books: Documentation preview :books:: https://equistore--165.org.readthedocs.build/en/165/

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -17,21 +17,25 @@ passenv =
 allowlist_externals =
     bash
 
-
 [testenv:tests]
 # this environement runs Python tests
 
 deps =
     discover
     numpy
+    pytest
+    torch
 
 commands =
     bash -c "rm -rf ./dist"
     pip wheel --no-deps -w dist .
     bash -c "python -m pip install --force-reinstall --no-deps ./dist/equistore-*.whl"
 
+    # Run Python unittest tests
     discover -p "*.py" -s python/tests
 
+   # Run pytest tests
+    pytest
 
 [testenv:lint]
 # this environement lints the Python code with flake8 (code linter), black (code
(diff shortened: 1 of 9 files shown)


Luthaf:
I don't live that we need to have files in the `tests` folder containing `test` in their names, feels a bit redundant. Could we use something [like this](https://docs.pytest.org/en/7.1.x/example/pythoncollection.html#customizing-test-collection) instead?

```ini
[pytest]
python_files = tests/*.py
```

PicoCentauri:
We could. I used this to avoid clashes with `unittest` but maybe this works.

PicoCentauri:
> I don't love that we need to have files in the `tests` folder containing `test` in their names, feels a bit redundant. Could we use something [like this](https://docs.pytest.org/en/7.1.x/example/pythoncollection.html#customizing-test-collection) instead?
> 
> ```ini
> [pytest]
> python_files = tests/*.py
> ```

seems to work

DavideTisi (approved):

DavideTisi:
very nice

tox.ini
```diff
@@ -17,21 +17,25 @@ passenv =
...
+    pytest
+    torch
 
 commands =
     bash -c "rm -rf ./dist"
     pip wheel --no-deps -w dist .
     bash -c "python -m pip install --force-reinstall --no-deps ./dist/equistore-*.whl"
 
+    # Run Python unittest tests
     discover -p "*.py" -s python/tests
```
Luthaf, line 35: `discover` only runs unittest tests?

tox.ini
```diff
@@ -17,21 +17,25 @@ passenv =
...
 
-
 [testenv:tests]
 # this environement runs Python tests
 
 deps =
     discover
     numpy
+    pytest
+    torch
```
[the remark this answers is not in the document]
PicoCentauri: Yes, We have something similar in rascaline. I can tackle this once we ported everything.

tox.ini
Luthaf: Yes, test speedup are nice!

PR #167, opened 2023-03-06
Try to fix docs permission errors

Luthaf:
The job pushing documentation to gh-pages is broken right now (https://github.com/lab-cosmo/equistore/actions/runs/4341978102/jobs/7582234507), I suspect this is because of the tentative permission fix from #163. This PR tries to workaroud the issue in another way, having a separate workflow for posting the link

### DIFF
--- a/.github/workflows/documentation-links.yml
+++ b/.github/workflows/documentation-links.yml
@@ -1,0 +1,17 @@
+name: readthedocs/actions
+
+on:
+  pull_request_target:
+    types:
+      - opened
+
+permissions:
+  pull-requests: write
+
+jobs:
+  documentation-links:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/documentation-links.yml
```diff
@@ -0,0 +1,16 @@
+name: readthedocs/actions
+on:
+  pull_request_target:
+    types:
+      - opened
```
Luthaf, line 5: This is also different from the `docs` workflow

Luthaf: Seems to result in a skipped job =/ https://github.com/lab-cosmo/equistore/actions/runs/4343538892/jobs/7585704675

Luthaf:
This seems to work: https://github.com/lab-cosmo/equistore/pull/169

PR #178, opened 2023-03-09
import `io` such that it is consistent with other equistore functions

### LINKED ISSUE #154: Importing `equistore.io` automatically, and perhaps a rename?
**Importing**

I notice that these functions aren't automatically loaded, i.e. you can't do:

```py
import equistore
equistore.io.load(...)
>>>
  File "<stdin>", line 1, in <module>
AttributeError: module 'equistore' has no attribute 'io'
```
you have to do:
```py
import equistore.io
equistore.io.load(...)
```

Having it imported would be consistent with the new importing convention, with qualified names for functions, explicit imports for classes, i.e. from @Luthaf 's suggestion:

```py
import equistore
from equistore import TensorMap, TensorBlock, Labels

def func(...) -> TensorMap:
    pass

a = TensorBlock(...)
b = TensorMap(...)

c = equistore.function(a, b)
```

**A rename?**

In a similar vein to numpy and torch, should we just have `equistore.save` and `equistore.load` instead of `equistore.io.save` and `equistore.io.load`?

jwa7:
Currently, one has to do imports in the following way:

```py
import equistore
import equistore.io
from equistore import Labels


tensor = equistore.io.load("tensor.npz")
samples_to_keep = Labels(...)
equistore.slice(tensor, samples=samples_to_keep)
``` 

In my opinion, the explicit import of the `io` module is clunky. This PR simply imports `io` 's functions in `__init__.py`, such that imports are simplified, and the distinction between implicit imports of functions vs explicit imports of classes is clear:

```py
import equistore
from equistore import Labels


tensor = equistore.io.load("tensor.npz")
samples_to_keep = Labels(...)
equistore.slice(tensor, samples=samples_to_keep)
``` 

This is not a breaking change, so AFAIK will not affect current workflows or tests.

Closes #154.

----
:books: Documentation preview :books:: https://equistore--178.org.readthedocs.build/en/178/


Luthaf:
Do we want to still allow import from the `io` module? It would be nice if there is only one way to do things, according to [`import this`](https://peps.python.org/pep-0020/)

jwa7:
Changes made such that one can do: `equistore.load()` and `equistore.save()`, but for advanced functions one has to use `equistore.io.create_numpy_array()`. However, this is not a breaking change as one can still use `equistore.io.save()`. Now, however, users do not need to explicitly do `import equistore.io`.

PR #179, opened 2023-03-09
drop_blocks and tests

### LINKED ISSUE #168: Introduce a `drop_blocks` operation
As mentioned in issue #145 , an equistore operation that drop blocks from a TensorMap by their key name would be useful.I already have a working prototype function, but it is currently untested: 

```py
def drop_blocks(tensor: TensorMap, keys: Labels) -> TensorMap:
    """
    Drop specified key/block pairs from a TensorMap.

    :param tensor: the TensorMap to drop the key-block pair from.
    :param keys: a :py:class:`Labels` object containing the keys of the 
        blocks to drop

    :return: the input :py:class:`TensorMap` with the specified key/block 
        pairs dropped.
    """
    if not np.all(tensor.keys.names == keys.names):
        raise ValueError(
            "The input tensor's keys must have the same names as the specified"
            + f" keys to drop. Should be {tensor.keys.names} but got {keys.names}"
        )
    new_keys = tensor.keys[tensor.keys != keys]
    return TensorMap(keys=new_keys, blocks=[tensor[key].copy() for key in new_keys])
```

I'm tagging this as a `good first issue` because it is a simple function that should be fairly simple to write unit tests for. Tests should check that all the appropriate keys in the output TensorMap are there, and the ones requested to be dropped are not. Then, the metadata and data of each block still present should be checked again the original TensorMap, including gradients.

cbenmahm:
fix #168 

----
:books: Documentation preview :books:: https://equistore--179.org.readthedocs.build/en/179/


python/tests/operations/drop_blocks.py
```diff
@@ -0,0 +1,131 @@
...
+
+import numpy as np
+
+import pytest
+import equistore
+from equistore import TensorMap, Labels
+
+
+DATA_ROOT = os.path.join(os.path.dirname(__file__), "..", "data")
+
```
jwa7, line 11:
```suggestion
@pytest.fixture
def test_tensor_map() -> TensorMap:
    return equistore.io.load(
            os.path.join(DATA_ROOT, "qm7-power-spectrum.npz"),
            use_numpy=True,
        )
```

jwa7: Then for each function, add `test_tensor_map` as an argument. For instance:

```py
def drop_all(self, test_tensor_map):
    ...
```
and of course in the body of the functions the variable name should be changed from `tensor` to `test_tensor_map`.

python/tests/operations/drop_blocks.py
```diff
@@ -0,0 +1,131 @@
...
+            use_numpy=True,
+        )
+        non_existent_key = Labels(tensor.keys.names, np.array([[0, 0, 0]]))
+        with pytest.raises(ValueError):
+            equistore.drop_blocks(tensor, non_existent_key)
+
+
+def check_consistency(tensor1: TensorMap, tensor2: TensorMap):
+    if tensor1.keys.names != tensor2.keys.names:
+        raise ValueError("the two tensors have different labels")
```
jwa7, line 126:
```suggestion
        raise ValueError("the two tensors have different key names")
```

python/tests/operations/drop_blocks.py
```diff
@@ -0,0 +1,131 @@
...
+
+    keys = tensor2.keys
+    for key in keys:
+        if tensor2[key] != tensor1[key]:
+            raise ValueError("tensor2 has values that are not present in tensor1")
```
jwa7, line 131:
```suggestion
            raise ValueError(f"the blocks indexed key {key} are not equivalent")
```

python/tests/operations/drop_blocks.py
```diff
@@ -0,0 +1,131 @@
...
+import pytest
+import equistore
+from equistore import TensorMap, Labels
+
+
+DATA_ROOT = os.path.join(os.path.dirname(__file__), "..", "data")
+
+
+class TestDropBlock:
+    # this test will fail for now because of issue #175
```
jwa7, line 14: Nice thank you!

jwa7 (changes requested):
Looking really good, thanks @cbenmahm ! Some changes requested, particular with creating a pytest fixture and a few minor things, but mostly there :)

jwa7:
Also, we currently have a build failure on the docs. From the top directory of equistore, where you would run `tox`, try running `tox -e docs` and observing the output to see what the problem might be. If you need help let me know :)

python/src/equistore/operations/drop_blocks.py
```diff
@@ -0,0 +1,33 @@
...
+        )
+    diff = np.setdiff1d(keys, tensor.keys)
+    if len(diff) > 0:
+        raise ValueError(
+            "some keys in `keys` are not present in `tensor`."
+            f" Non-existent keys: {diff}"
+        )
+    # for key in keys:
+    #     if not (key in tensor.keys):
+    #         raise ValueError(f"passed key {key} does not exist in `tensor`")
```
jwa7, line 30:
```suggestion
```
these commented out lines need to be removed

python/tests/operations/drop_blocks.py
```diff
@@ -0,0 +1,113 @@
...
+def test_tensor_map() -> TensorMap:
+    return equistore.io.load(
+        os.path.join(DATA_ROOT, "qm7-power-spectrum.npz"),
+        # the npz is using DEFLATE compression, equistore only supports STORED
+        use_numpy=True,
+    )
+
+
+class TestDropBlock:
+    def test_drop_all(self, test_tensor_map):
```
jwa7, line 25: Some short docstrings on each test function describing what the test does would be useful

jwa7 (changes requested):

Luthaf:
Unrelated to this PR, but you can also add yourself to the [AUTHOR](https://github.com/lab-cosmo/equistore/blob/master/AUTHORS) file @cbenmahm!

jwa7 (approved):
Looks great, thanks @cbenmahm !

python/tests/operations/drop_blocks.py
```diff
@@ -0,0 +1,121 @@
...
+    def test_not_existent(self, test_tensor_map):
+        # test when keys that don't appear in the TensorMap are dropped
+        non_existent_key = Labels(
+            test_tensor_map.keys.names, np.array([[-1, -1, -1], [0, 0, 0]])
+        )
+        error_msg = (
+            "some keys in `keys` are not present in `tensor`."
+            f" Non-existent keys: {non_existent_key}"
+        )
+        with pytest.raises(ValueError, match=re.escape(error_msg)):
```
Luthaf, line 109: I don't think we need regex for this, since we are matching the full error message and not just part of it

PR #180, opened 2023-03-09
Make sure to uninstall existing python package in tox

Luthaf:
Otherwise we would just add new files in the package without removing previous ones. This was causing @jwa7 to see tests passing locally while failing in CI in #178 

----
:books: Documentation preview :books:: https://equistore--180.org.readthedocs.build/en/180/


PicoCentauri:
Is this due to the fact the we do not have `usedevelop` or is this a general problem.

Issue #176, opened 2023-03-08
Porting `equistore.operations` to `pytest`


PicoCentauri:
We already have the core tests in pytest but need to port the other functions as well. Whoever wants to help with this go for it. When writing the tests we should try to improve them a bit. Two things I mainly try to do is:

1. Reduce the code duplication especially in creating the `TensorMap`\'s a bit. For this, the `@pytest.fixture` decorator is a good way.
2. Use the `@pytest.mark.parametrize` to simplify tests that are doing the same logic on different values.

For reference, you can take a look at #165 and #170.

## Operation tests to port

- [x] `add` #165 
- [x] `allclose`
- [x] `divide`
- [x] `dot`
- [x] `empty_like`
- [x] `equal`
- [x] `join`
- [x] `lstsq`
- [x] `multiply`
- [x] `ones_like`
- [x] `pow`
- [x] `reduce_over_samples`
- [x] `remove_gradients`
- [x] `slice`
- [x] `solve`
- [ ] `split`
- [x] `subtract`
- [x] `unique_metadata`
- [x] `zeros_like`

PR #182, opened 2023-03-09
Run doctests and unittest-based tests with pytest

Luthaf:
The current hack to run doctests is starting to fail for some reason (see for example https://github.com/lab-cosmo/equistore/actions/runs/4374795802/jobs/7654672493), so let's move to pytest for this. 

This PR also uses pytest to run all unittest-based tests, so we only have to change the test syntax to finish #176

----
:books: Documentation preview :books:: https://equistore--182.org.readthedocs.build/en/182/


Luthaf:
The failure of the current hack is is really weird: 

```
 ======================================================================
ERROR: test_doctests (doctests.TestDoctests)
----------------------------------------------------------------------
  File "/home/runner/work/equistore/equistore/python/tests/doctests.py", line 16, in test_doctests
    submodule = importlib.import_module("equistore." + module.name)
  File "/opt/hostedtoolcache/Python/3.10.10/x64/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'equistore.operations.equal_metadata'
```

The code was trying to all sub-module by looking at all files, and then importing them. Here it is trying to import `equistore/operations/equal_metadata.py`, which does not exists on master or other impacted branches!!! This file is added in #147, which is not merged yet. This makes no sense to me ...

Luthaf:
OK, culprit it 54000b8ae0d5592b1efc25adba141cada05cb278. @jwa7 you pushed that to master =)

Luthaf:
I've reverted this commit and added protection to the master branch, this should prevent such accidental pushes

PicoCentauri:
Thanks for adding this!

Issue #149, opened 2023-02-28
`detach()` torch-based TensorMaps before saving them


jwa7:
When trying to save a TensorMap whose block values are `torch.Tensor` objects, an equistore error is raised if the values of the block have not been detached from their computational graph:

```py
from equistore import io
...
# tmp is a torch-based TensorMap
equistore.io.save("tmp.npz", tmp)
```
gives error:
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
File /opt/miniforge3/envs/rho/lib/python3.10/site-packages/equistore/utils.py:31, in catch_exceptions.<locals>.inner(*args, **kwargs)
     30 try:
---> 31     function(*args, **kwargs)
     32 except Exception as e:

File /opt/miniforge3/envs/rho/lib/python3.10/site-packages/equistore/data/array.py:138, in _eqs_array_data(this, data)
    134     # `.numpy()` will fail if the data is on GPU or requires gradient
    135     # tracking, and the resulting array is sharing data storage with the
    136     # tensor, meaning we can take a pointer to it without the array being
    137     # freed immediately.
--> 138     array = array.numpy()
    140 if not array.data.c_contiguous:

RuntimeError: Can't call numpy() on Tensor that requires grad. Use tensor.detach().numpy() instead.

The above exception was the direct cause of the following exception:

EquistoreError                            Traceback (most recent call last)
Cell In[23], line 6
      2 from rholearn import utils
      4 tmp = utils.tensor_to_torch(input, True, torch.float64, torch.device("cpu"))
----> 6 equistore.io.save("tmp.npz", tmp)

File /opt/miniforge3/envs/rho/lib/python3.10/site-packages/equistore/io.py:61, in save(path, tensor, use_numpy)
     59 else:
     60     lib = _get_library()
---> 61     lib.eqs_tensormap_save(path.encode("utf8"), tensor._ptr)

File /opt/miniforge3/envs/rho/lib/python3.10/site-packages/equistore/status.py:37, in _check_status(status)
     35 e = LAST_EXCEPTION
     36 LAST_EXCEPTION = None
---> 37 raise EquistoreError(last_error(), status) from e

EquistoreError: external error: calling eqs_array_t.data failed (status -1)
```

Ideally, as saving destroys pointers to the graph anyway, the `equistore.io.save` function should detach block values tensors before serialization, using `Tensor.detach()`.

jwa7:
[...]

PR #183, opened 2023-03-09
`to` function for converting the backend TensorMap values torch <-> numpy

### LINKED ISSUE #70: Create a utility function to convert ``TensorMap`` block values to type ``torch.tensor``
In order to be compatible as input to PyTorch models, the ``TensorBlock.values`` tensors in each block of a ``TensorMap`` need to be of type ``torch.tensor``, instead of a ndarray, i.e. ``numpy.ndarray`` or ``_RustNDArray``.

**Desired Behaviour**

There should be a utility function accessible through the Python API, perhaps in a new file such as _equistore/python/src/equistore/operations/tensormap_to_torch.py_, that is used in the following way:
```py
new_tensormap = tensormap_to_torch(tensormap)
```

**Implications**

This necessarily has to be implemented as part of the Python API as there is no Rust-side solution to creating ``torch.tensor`` objects. This means that the above function ``tensormap_to_torch()`` will have to return a new ``TensorMap`` object, thus creating a copy of the object in memory. However, doing it this way round means that all the infrastructure that builds descriptors need not be written to be compatible with PyTorch; only at the step before building a model do we need to convert our blocks to ``torch.tensor`` objects.

jwa7:
Closes #70 and #149.

Still to do:

- [x] Finish off functionality
- [x] Write unit tests

To discuss:

- Let the arg `dtype` be passed as a `torch.dtype` object, or as a`str`? The former is more flexible.
- Should the `detach()` function just be a wrapper for `to`? Do we need a `detach` function as well as `to`?
- Add device support (and maybe dtype support) for the private functions `_to_numpy` and `_to_numpy_block`?


python/src/equistore/operations/to.py
```diff
@@ -0,0 +1,188 @@
+import numpy as np
+import torch
+
+from ..block import TensorBlock
+from ..labels import Labels
+from ..tensor import TensorMap
+
+
+def to(
```
Luthaf, line 9: I would rather have two separate functions for TensorMap and TensorBlock, like the rest of the operations API

python/src/equistore/operations/to.py
```diff
@@ -0,0 +1,179 @@
...
+            f"``block`` should be an equistore `TensorBlock`, got {type(block)}"
+        )
+    if not isinstance(backend, str):
+        raise TypeError("`backend` should be passed as a `str`")
+    if backend == "torch":
+        # Check args
+        if not isinstance(requires_grad, bool):
+            raise TypeError("`requires_grad` should be passed as a `bool`")
+        if not isinstance(dtype, torch.dtype):
+            raise TypeError("`dtype` should be passed as a `torch.dtype`")
```
Luthaf, line 93: Not really, torch also accepts strings here. Overall, I think it is fine not to type check these, and pass them through to the backend which will handle them for us.

python/src/equistore/operations/to.py
```diff
@@ -0,0 +1,179 @@
...
+        if not isinstance(device, torch.device):
+            raise TypeError("`device` should be passed as a `torch.device`")
+        return _to_torch_block(block, requires_grad, dtype, device)
+    elif backend == "numpy":  # numpy
+        # TensorBlock
+        return _to_numpy_block(block)
```
Luthaf, line 99: this should also take a dtype

python/src/equistore/operations/to.py
```diff
@@ -0,0 +1,179 @@
...
+    block: TensorBlock,
+    requires_grad: bool,
+    dtype: torch.dtype,
+    device: torch.device,
+) -> TensorBlock:
+    """
+    Creates a new :py:class:`TensorBlock` where block values are PyTorch
+    :py:class:`torch.tensor` objects. Assumes the block values are already as a
+    type that is convertible to a :py:class:`torch.tensor`, such as a numpy
+    array or RustNDArray. The resulting torch tensor dtypes are enforced as
```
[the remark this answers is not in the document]
Luthaf: RustNDArray does not exists anymore

python/src/equistore/operations/to.py
```diff
@@ -0,0 +1,179 @@
...
+    Takes a TensorBlock object whose values are torch.tensor objects and
+    converts them to numpy arrays of dtype np.float64. Returns a new TensorBlock
+    object.
+    """
+    if isinstance(block.values, np.ndarray):
+        return block.copy()
+
+    # Create new block, with the values tensor converted to a torch tensor.
+    new_block = TensorBlock(
+        values=np.array(block.values.detach().numpy(), dtype=np.float64),
```
Luthaf, line 163: I don't love that we assume torch here, but I think we have to since we need to call detach =/

Do you know if we could use dlpack functionalities here instead?

python/src/equistore/operations/to.py
```diff
@@ -0,0 +1,179 @@
...
+        components=block.components,
+        properties=block.properties,
+    )
+
+    # Add gradients to each block, again where the values tensor are torch
+    # tensors.
+    for parameter, gradient in block.gradients():
+        new_block.add_gradient(
+            parameter,
+            np.array(gradient.data.detach().numpy(), dtype=np.float64),
```
jwa7, line 174: This will need updating to reflect the new Gradients as TensorBlocks change

python/src/equistore/operations/to.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def _to_numpy_block(block: TensorBlock) -> TensorBlock:
+    """
+    Takes a TensorBlock object whose values are torch.tensor objects and
+    converts them to numpy arrays of dtype np.float64. Returns a new TensorBlock
+    object.
+    """
+    if isinstance(block.values, np.ndarray):
+        return block.copy()
```
jwa7, line 159: What memory considerations should we make here @Luthaf @frostedoyster ? Do we want to provide the option to release the previous tensor from memory? If so, should the default behaviour be safe (return a copy) or unsafe?

See discussion in https://github.com/lab-cosmo/equistore/pull/248

python/src/equistore/operations/to.py
frostedoyster: I removed the detach, not sure if that still works. If necessary, we can only detach if values is an instance of torch.Tensor

Luthaf: > If necessary, we can only detach if values is an instance of torch.Tensor

Yes, we'll have to detach torch tensor, otherwise we'll get an error in the call to `.numpy()`

python/equistore-operations/tests/to.py
```diff
@@ -0,0 +1,248 @@
...
+    block = equistore.block_from_array(values)
+    grad_block = equistore.block_from_array(grad_values)
+    grad_grad_block = equistore.block_from_array(grad_grad_values)
+    grad_block.add_gradient("grad_grad", grad_grad_block)
+    block.add_gradient("grad", grad_block)
+    return block
+
+
+@pytest.fixture
+def tmap():
```
Luthaf, line 25:
```suggestion
def tensor():
```

Please optimize code for readability, not ease of writing. Much more time is spent reading code than writing it: https://www.goodreads.com/quotes/835238-indeed-the-ratio-of-time-spent-reading-versus-writing-is

frostedoyster: Yes but that doesn't return a tensor

Luthaf:
```
    tmap = TensorMap(keys=Labels.arange("dummy", 2), blocks=[block1, block2])
    return tmap
```

Luthaf: We use `tensor` for all fixtures returning a TensorMap: https://github.com/lab-cosmo/equistore/blob/7e8b41926ca61039f9113510daa8191270aa23c0/python/equistore-operations/tests/add.py#L15

jwa7:
I've made some changes - using dispatch, merged master, and some updates to docs, typing and style. I've broken the CI so will fix next.

Just a couple of questions about the default value of `requires_grad`. If I have a torch-based TensorMap and I use `to` with the backend set to "torch". At the moment, in this case, the `requires_grad` arg is essentially ignored.

1) Should we allow the user, in the above example, to change `requires_grad`?
2) Should the default value of `requires_grad` be `None` instead?

frostedoyster (commented):
Is that return necessary? Regarding the requires_grad for the torch-to-torch conversion, I don't know... actually the only case where it is useful is probably when going from numpy to torch. We could also remove it altogether and then people would have to modify requires_grad after calling `equistore.to`

python/equistore-operations/equistore/operations/_dispatch.py
```diff
@@ -400,3 +400,46 @@ def rand_like(array, shape=None, requires_grad=False):
...
+        if backend == "numpy":
+            return array.detach().cpu().numpy()
+
+        elif backend == "torch":
+            # We need this to keep gradients of the tensor
+            return array.to(dtype=dtype, device=device)
+
+        else:
+            raise ValueError(f"Unknown backend: {backend}")
+        return np.array(array, dtype=dtype)
```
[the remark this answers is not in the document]
jwa7: Re: requires_grad - yes that makes sense. Thoughts @Luthaf ?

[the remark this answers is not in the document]
Luthaf: I would agree that we could leave it off, and add a separate operation to change requires_grad (which should be doable without creating new TensorMap/TensorBlock)

python/equistore-operations/equistore/operations/_dispatch.py
```diff
@@ -402,7 +402,7 @@ def rand_like(array, shape=None, requires_grad=False):
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
-def to(array, backend: str = None, dtype=None, device=None, requires_grad=False):
+def to(array, backend: str = None, dtype=None, device=None, requires_grad=None):
```
Luthaf, line 405: I think it would be nicer long term to have separate `to_torch` and `to_numpy` functions, taking whichever parameter they need to. This makes it easier to add other backends later.

Luthaf (commented):
Only one small comment and this should be good to go

PR #196, opened 2023-03-14
Added random-uniform-like operation on TensorMaps

abmazitov:
----
:books: Documentation preview :books:: https://equistore--196.org.readthedocs.build/en/196/

### DIFF
--- a/python/src/equistore/operations/random_like.py
+++ b/python/src/equistore/operations/random_like.py
@@ -31,7 +31,7 @@ def random_uniform_like_block(
 
     :param block: Input block from which the metadata is taken.
     :param parameters: Which gradient parameter to copy. If ``None`` (default)
-                       all gradient of ``tensor`` are present in the new tensor.
+                       all gradients of ``block`` are present in the new block.
                        If empty list ``[]`` no gradients information are copied.
     """
     values = _dispatch.random_uniform(block.values)


Luthaf:
The linter is failing! You can run it locally with `tox -e lint`, and reformat the code automatically with `tox -e format`

python/tests/operations/random_like.py
```diff
@@ -0,0 +1,145 @@
...
+
+import numpy as np
+
+import equistore
+
+
+DATA_ROOT = os.path.join(os.path.dirname(__file__), "..", "data")
+
+
+class TestRandomLike(unittest.TestCase):
```
DavideTisi, line 12: i know that it will be a pain but we are moving to pytest, and any new addition to the code have to have the test in pytest not unittest

python/src/equistore/operations/random_like.py
```diff
@@ -0,0 +1,68 @@
...
+def random_uniform_like_block(
+    block: TensorBlock, parameters: Union[List[str], str] = None
+) -> TensorBlock:
+    """Return a new :py:class:`TensorBlock` with the same metadata as block, and all
+    values randomly sampled from the uniform distribution between 0 and 1.
+    :param block: Input block from which the metadata is taken.
+    :param parameters: Which gradient parameter to copy. If ``None`` (default)
+                       all gradient of ``tensor`` are present in the new tensor.
+                       If empty list ``[]`` no gradients information are copied.
+    :param requires_grad: If autograd should record operations for the returned tensor.
```
DavideTisi, line 36: `required grad` is not a parameter of the function

python/tests/operations/random_like.py
```diff
@@ -0,0 +1,145 @@
...
+            self.assertTrue(np.all(rand_block.properties == block.properties))
+            self.assertEqual(len(rand_block.components), len(block.components))
+            self.assertTrue(
+                np.all(
+                    [
+                        np.all(rand_block.components[i] == block.components[i])
+                        for i in range(len(block.components))
+                    ]
+                )
+            )
```
DavideTisi, line 34: I would remove the check of the metadata and use before the loop
`python
assert equistore.equal(rand_tensor,tensor,only_metadata=True)
`
Like this it checks if the metadata are the same, then you can leave the loop to checks the values

DavideTisi: same thing for the gradinet, just leave the check of the values

python/tests/operations/random_like.py
```diff
@@ -0,0 +1,145 @@
...
+                            np.all(
+                                rand_gradient.components[i] == gradient.components[i]
+                            )
+                            for i in range(len(gradient.components))
+                        ]
+                    )
+                )
+
+                self.assertTrue(np.all(rand_gradient.data >= 0))
+                self.assertTrue(np.all(rand_gradient.data <= 1))
```
DavideTisi, line 101: same as before for the metadata

DavideTisi (changes requested):
very good, i requested few changed in the test should be easy since you can reduce the number of check with `equistore.equal(only_metadata=True)`

python/tests/operations/random_like.py
Luthaf: Is a gradinet a small gradient 😁 ? 🇮🇹 🤌

abmazitov:
Ok, thank you for the response, I will do the changes

python/tests/operations/random_like.py
DavideTisi: it is a gradient "gratinato"

DavideTisi:
Some tests are failing because since `1a495992d6d6e82cc508fed4926a91babea311ac` 
`equal` does not have `only_metadata` anymore, but you have to use
`equistore.equal_metadata`

DavideTisi (changes requested):
very good, two things one change the name of the file equal to the name of the function, just to be consistent to the rest of ops. And you should add the operation in the docs

abmazitov:
> 

I was actually thinking to add another way of generating random array, namely with a gaussian distribution. Should make it in a separate file or put inside the random_like.py ?

PR #197, opened 2023-03-14
New tensor `transpose` operation

### LINKED ISSUE #191: Write a `transpose` operation
As mentioned in #101, we need a transpose function. Ideally this should work for TensorMaps whose blocks have components (i.e. equivariants) as well as gradients. The tentative idea after discussion with @Luthaf and @SanggyuChong  is that properties of `X` become the new samples in `X.T`, whilst the values samples and gradient samples become concatenated into a combined "q" Labels object. 

The construction of "q" should be well documented and follow certain ordering rules. The base reordering of the block/gradient values should be dispatched to torch/numpy.

@SanggyuChong and I will start work on this tomorrow on COSMO day.

jwa7:
COSMO day collaboration: @DivyaSuman14,@SanggyuChong, and I. Closes #191.

New `transpose` function for TensorMaps and TensorBlocks. Returns a new TensorMap where axes have been transposed. Each returned block has samples that correspond to the original block's properties. Components axes are reversed. Properties correspond to the original block's samples. 

In the case where the blocks of a tensor have gradients, the transposed block is joined along the new properties (previously samples) axis with all of its transposed gradients. Due to clashing dimensions of values and gradients samples, a new hybrid label "q" is created and used as the metadata for the returned blocks, joined along the properties.

Collapsing of the properties axis labels to a single numeric index "q" results in information loss. We should also think about providing a user flag for the function that allows the samples metadata of the original TensorMap to be returned, such that the original TensorMap can be constructed. In other words, one need to retain the original samples metadata if one wants be able to call transpose twice on a TensorMap and reclaim the original TensorMap. With the loss of information associated with creation of the "q" index, this is not yet possible.

To Do:
- [x] Test the function for usability
- [ ] Write unit tests
- [ ] Write function docstrings
- [ ] Properly document the action of the operation and the meaning of the new axes, in particular the "q" label used for the new properties
- [ ] EXTRA: provide a user flag for returning the original TensorMap's sample metadata., as well as functionality (and tests_ to allow users to transpose twice and reclaim their original TensorMap.
- [ ] All functions that involve taking transposes of block values (i.e. `dot` and `solve`) should be updated to use `equistore.transpose()` instead.

----
:books: Documentation preview :books:: https://equistore--197.org.readthedocs.build/en/197/

### DIFF
--- a/docs/src/operations/reference/linear_algebra/index.rst
+++ b/docs/src/operations/reference/linear_algebra/index.rst
@@ -7,3 +7,4 @@ Linear algebra
     dot() <dot>
     lstsq() <lstsq>
     solve() <solve>
+    transpose() <transpose>
--- a/docs/src/operations/reference/linear_algebra/transpose.rst
+++ b/docs/src/operations/reference/linear_algebra/transpose.rst
@@ -1,0 +1,6 @@
+transpose
+=========
+
+.. autofunction:: equistore.transpose
+
+.. autofunction:: equistore.transpose_block
--- a/python/metatensor-operations/metatensor/operations/transpose.py
+++ b/python/metatensor-operations/metatensor/operations/transpose.py
@@ -1,0 +1,84 @@
+"""
+Module to transpose a :py:class:`TensorMap` or :py:class:`TensorBlock`.
+"""
+import numpy as np
+
+from ..block import TensorBlock
+from ..labels import Labels
+from ..tensor import TensorMap
+from . import _dispatch
+
+
+def transpose(tensor: TensorMap) -> TensorMap:
+    """
+    Transposes a :py:class:`TensorMap`.
+    """
+    # Check input
+    if not isinstance(tensor, TensorMap):
+        raise TypeError(f"Expected a TensorMap, got {type(tensor)}.")
+    # Transpose TensorMap
+    keys = tensor.keys
+    return TensorMap(keys, [transpose_block(tensor[key]) for key in keys])
+
+
+def transpose_block(block: TensorBlock) -> TensorBlock:
+    """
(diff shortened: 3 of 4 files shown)


agoscinski:
Started a rebase, I would like to let the transpose still keep its reflexive characteristics when its applied two times to itself, in contrary to the current implementation. As far as I understand from a conversation with @jwa7 is that the way it was implemented here is for the purpose of covariance/correlation matrices. I would try to move this stacking of samples with this components to such operations.

Luthaf:
> I would like to let the transpose still keep its reflexive characteristics when its applied two times to itself, in contrary to the current implementation.

I think I agree with you, but this means `transpose` can not support gradients, right?

> I would try to move this stacking of samples with this components to such operations.

The idea would be that a covariance matrix including gradients would be something like this?

```py
X = ...
X_m = ...

K = metatensor.dot(X, metatensor.transpose(metatensor.gradients_to_samples(X_m)))
```

> Started a rebase

Might be better to start a new branch!

agoscinski:
> I think I agree with you, but this means transpose can not support gradients, right?
> The idea would be that a covariance matrix including gradients would be something like this?X = ...
> X_m = ...
> K = metatensor.dot(X, metatensor.transpose(metatensor.gradients_to_samples(X_m)))

Yes I cannot a think of a way to keep gradients in a transpose without changes in core. Since we are moving in a direction to move the functionality of a the transpose for inner and outer products to a an operation, I think this is the way to go to not allow gradients in this operation.

jwa7:
Closing for now as this doesn't seem particularly useful to anyone at the moment. We can always pick it back up later if that changes

PR #198, opened 2023-03-14
 Add an example for zero_like function

melikahnd:
This is an example for zero_like function

----
:books: Documentation preview :books:: https://equistore--198.org.readthedocs.build/en/198/


Luthaf:
The linter is failing! You can run it locally with tox -e lint, it will tell you what to change

Luthaf:
Replaced by #211

PR #199, opened 2023-03-14
Add allclose example to docs

victorprincipe:
Here I have added examples to the `allclose.rst` file in the docs.

The examples are included from a python script (`allclose_example.py`) that has been added to the same directory as the `.rst` file.

Various examples have been included for the following methods:
- `allclose`
-  `allclose_raise`
- `allclose_block`
- `allclose_block_raise`

----
:books: Documentation preview :books:: https://equistore--199.org.readthedocs.build/en/199/


docs/src/reference/python/operations/logic/allclose_example.py
```diff
@@ -0,0 +1,127 @@
...
+    components=[],
+    properties=Labels(
+        ["PROPERTIES"], np.array([[0], [1], [2]])
+     ),
+)
+
+allclose_block(block1, block2)
+#Output:
+# False
+### This fails because allclose_block checks that the properties of two blocks are equal too
```
agoscinski, line 52: here I dont have a strong opinion, but maybe calling them `property_1` and `property_2` is easier to understand rather than one is upper case of the other and that is why they are different.

docs/src/reference/python/operations/logic/allclose_example.py
```diff
@@ -0,0 +1,127 @@
...
+# as to why it fails
+allclose_block_raise(block1, block2)
+#Output:
+# ValueError: Inputs to 'allclose' should have the same properties:
+# properties names are not the same or not in the same order.
+
+# Recreate block1, but change first value in the block from 1 to 1.00001
+block3 = TensorBlock(
+    values=np.array([
+         [1.00001, 2, 4],
```
agoscinski, line 64: Maybe it visually easier to grasp your examples below when you define 1.00001 as 1+1e-5

docs/src/reference/python/operations/logic/allclose_example.py
```diff
@@ -0,0 +1,127 @@
...
+
+allclose(tensor1, tensor4, rtol=1e-5)
+#Output:
+# False
+### This fails, because allclose also checks that the keys of each tensor are equal
+
+# Using allclose_raise will raise a ValueError if the check fails, along with an explanation as to why 
+allclose_raise(tensor1, tensor4, rtol=1e-5)
+#Output:
+# ValueError: Inputs to allclose should have the same key indices.
```
agoscinski, line 127: Maybe for the `allclose_raises` one could write in the documentation here
https://github.com/lab-cosmo/equistore/blob/901fe071cc3c8ec87530d5669e8cf93fd8f2e611/python/src/equistore/operations/allclose.py#L51-L71
that it raises a ValueError if the keys are different, similar as in
https://github.com/numpy/numpy/blob/8cec82012694571156e8d7696307c848a7603b4e/numpy/linalg/linalg.py#L2201-L2204
for the convergence with
```
    Raises
    ------
    LinAlgError
        If computation does not converge.
```

agoscinski (changes requested):
I think it is already looking good, but one important change that needs to be done is to move the examples to the corresponding functions `allclose`, `allclose_block` and `allclose_block_raise`, `allclose_raise` as it is done for the `reduce_over_samples` function here
https://github.com/lab-cosmo/equistore/blob/16ff8ea4b1b466892473e53bf042e1683417e410/python/src/equistore/operations/reduce_over_samples.py#L266-L327
In the doc it looks then like this
https://lab-cosmo.github.io/equistore/latest/reference/python/operations/manipulation/samples-reduction.html#equistore.sum_over_samples

This has the advantage that it is in the help page of the function using `help` and we can automatically test if the doc test is printing the same output as specified in the doc with `pytest --doctest-modules --pyargs equistore`

DavideTisi:
Very cool, one thing in the doc the actual docstrings are not shown anymore

docs/src/reference/python/operations/logic/allclose_example.py
victorprincipe: In the description of the function, it already says to look at the `allclose` method for how two blocks are considered (not) equal. Please see below.

https://github.com/lab-cosmo/equistore/blob/901fe071cc3c8ec87530d5669e8cf93fd8f2e611/python/src/equistore/operations/allclose.py#L62-L64

docs/src/reference/python/operations/logic/allclose_example.py
victorprincipe: Bon idée

docs/src/reference/python/operations/logic/allclose_example.py
victorprincipe: Très bon idée

victorprincipe:
I've added the examples to the functions themselves, and removed the examples section from the `.rst` file for the doc. Please let me know what you think.

Also, in response to this @DavideTisi:

> Very cool, one thing in the doc the actual docstrings are not shown anymore

Seems like they are quoted here: https://lab-cosmo.github.io/equistore/latest/reference/python/operations/manipulation/samples-reduction.html

victorprincipe:
Also, maybe we should run some other tests apart from just building the docs, as I altered module files.

victorprincipe:
> Very cool, one thing in the doc the actual docstrings are not shown anymore

Ah sorry, understand what you meant now. This was an issue with tax I believe which was fixed in the master branch. I have rebased and now it should work.

docs/src/reference/python/operations/logic/allclose_example.py
agoscinski: you are right, we don't need to repeat the information.

python/src/equistore/operations/allclose.py
```diff
@@ -63,6 +149,71 @@ def allclose_raise(
...
+    ... )
+    ...
+    >>> #Create tensors from blocks, using same keys
+    >>> keys = Labels(names=["key"], values=np.array([[0]]))
+    ...
+    >>> tensor1 = TensorMap(keys, [block1])
+    >>> tensor2 = TensorMap(keys, [block2])
+    ...
+    >>> # Call allclose_raise, which should return a ValueError because:
+    >>> # 1. The two NaNs are not considered equal of the two tensors is 1e-5,
```
agoscinski, line 201: Should it be like this? I don't understand the current sentence
```suggestion
    >>> # 1. The two NaNs are not considered equal,
```

python/src/equistore/operations/allclose.py
```diff
@@ -63,6 +149,71 @@ def allclose_raise(
...
+    >>> # 2. The difference between the first value in the blocks
+    >>> #    which is greater than the default rtol of 1e-13
```
agoscinski, line 203: Should it be like this? "The difference [...] is greater than [...]"
```suggestion
    >>> # is greater than the default rtol of 1e-13
```

agoscinski (changes requested):
Looks almost ready to squash and rebase to me, just two sentences I did not quite get.

python/src/equistore/operations/allclose.py
victorprincipe: Yes, typo, will commit the suggestion

python/src/equistore/operations/allclose.py
victorprincipe: Yes, will also commit this.

agoscinski (approved):
I think it looks good now

PR #201, opened 2023-03-14
Added documentation for gradient() and add_gradient()

HowWeiBin:
Added documentation code snippets for gradient and add_gradient

----
:books: Documentation preview :books:: https://equistore--201.org.readthedocs.build/en/201/

### DIFF
--- a/python/src/equistore/block.py
+++ b/python/src/equistore/block.py
@@ -280,6 +280,7 @@ class TensorBlock:
         Add a set of gradients with respect to ``parameters`` in this block.
 
         Here is an example of how to use this method:
+        
         >>> import numpy as np
         >>> block_1 = TensorBlock(
         ...             values=np.full((3, 1, 1), 1.0),
--- a/python/src/equistore/block.py
+++ b/python/src/equistore/block.py
@@ -207,11 +207,12 @@ class TensorBlock:
 
     def gradient(self, parameter: str) -> "Gradient":
         """
-        Get the gradient of the ``values`` in this block with respect to
-        the given ``parameter``.
+        Get the gradient of the block ``values``  with respect to the
+        given ``parameter``.
 
         Here is an example of how to use this method:
         >>> import numpy as np
+        >>> from equistore import TensorBlock, Labels
         >>> block_1 = TensorBlock(
         ...             values=np.full((3, 1, 1), 1.0),
         ...             samples=Labels(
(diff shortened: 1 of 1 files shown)


agoscinski (changes requested):
The linter is not happy, you can see the errors here
https://github.com/lab-cosmo/equistore/actions/runs/4416932020/jobs/7741914595#step:7:512
or you can get it locally by running `tox -e lint`
I think some of the linter errors you can automatically resolve with `tox -e format`.

python/src/equistore/block.py
```diff
@@ -7,7 +7,7 @@
 from .data import Array, ArrayWrapper, eqs_array_to_python_array
 from .labels import Labels
 from .status import _check_pointer
-
+import numpy as np
```
agoscinski, line 10: The linter seems not to be soo smart, so it throws an error here
https://github.com/lab-cosmo/equistore/actions/runs/4416932020/jobs/7741914595#step:7:512
In the code you don't need the Labels, so it complains. It apparently cannot see that it is used in the doc.
This should fix the problem
```suggestion
import numpy as np # noqa: F401
```

agoscinski: We decided to move it to the doc, so here we remove it
```suggestion

```
(see second suggestion where to add it)

python/src/equistore/block.py
```diff
@@ -205,6 +205,47 @@ def gradient(self, parameter: str) -> "Gradient":
         Get the gradient of the ``values`` in this block with respect to
         the given ``parameter``.
 
+        Here is an example of how to use this method:
+
```
agoscinski, line 209: and here we add it
```suggestion
        >>> import numpy as np
```

HowWeiBin:
Recommended changes have been made, along with some additional edits for the formatting issues raised initially.

python/src/equistore/block.py
```diff
@@ -205,6 +205,52 @@ def gradient(self, parameter: str) -> "Gradient":
...
+        ...     components=[Labels(["components"], np.array([[0]], dtype=np.int32))],
+        ... )
+        >>> block_1.add_gradient(
+        ...     "parameter2",
+        ...     samples=Labels(
+        ...        ["sample", "parameter2"], np.array([[0, -2], [2, 3]], dtype=np.int32)
+        ...     ),
+        ...     data=np.full((2, 1, 1), 15.0),
+        ...     components=[Labels(["components"], np.array([[0]], dtype=np.int32))],
+        ... )
```
agoscinski, line 236: It would be helpful if we can show in the second block how to add components in gradients,
```suggestion
        >>> block_1.add_gradient(
        ...     "cell",
        ...     data=np.full((2, 6, 1, 1), 15.0),
        ...     samples=Labels(
        ...        ["sample"], np.array([[0], [2]], dtype=np.int32)
        ...     ),
        ...     components=[Labels(["cell"], np.array([[0], [1], [2], [3], [4], [5]], dtype=np.int32)), Labels(["components"], np.array([[0]], dtype=np.int32))],
        ... )
```

python/src/equistore/block.py
```diff
@@ -224,6 +270,34 @@ def add_gradient(
         """
         Add a set of gradients with respect to ``parameters`` in this block.
 
+        Here is an example of how to use this method:
```
agoscinski, line 273: Same comments as for the last example apply here. Can you swap the examples between `gradient` and  `add_gradient`? At the moment the examples in `gradient` are more comprehensive, but I believe that a user will rather run into problems when using `add_gradient` and try to read the doc from this function. So it would be better if this one is more comprehensive.

python/src/equistore/block.py
```diff
@@ -205,6 +205,52 @@ def gradient(self, parameter: str) -> "Gradient":
         Get the gradient of the ``values`` in this block with respect to
         the given ``parameter``.
```
agoscinski, line 206: Does this sound better to you?
```suggestion
        Get the gradient of the block ``values`` with respect to the
        given ``parameter``.      
```

python/src/equistore/block.py
```diff
@@ -224,6 +270,34 @@ def add_gradient(
         """
         Add a set of gradients with respect to ``parameters`` in this block.
```
agoscinski, line 271: Added information how to add components to a gradient. Also tried to make the first sentence a bit clearer. Do you think it is clear?
```suggestion
        Add a gradient containing gradient values with respect to ``parameter`` to this block.
        The gradient must contain the values components, but can contain additional components.
        These additional components must be prepended to the list of values components, 
        with the values components following.
```
```suggestion
        Get the gradient of the block ``values`` with respect to the
        given ``parameter``. 
      
```

agoscinski (changes requested):
Looks already good! Needs just some tweaks.

agoscinski:
Sorry, I just realized there is an add_gradient PR https://github.com/lab-cosmo/equistore/pull/204 so we just push everything into the gradient function to avoid merge conflicts.

python/src/equistore/block.py
agoscinski: we do this in a different PR

python/src/equistore/block.py
```diff
@@ -229,6 +279,34 @@ def add_gradient(
...
+        ...     ),
+        ...     data=np.full((2, 1, 1), 11.0),
+        ...     components=[Labels(["components"], np.array([[0]], dtype=np.int32))],
+        ... )
+        >>> print(block_1)
+        TensorBlock
+            samples (3): ['samples']
+            components (1): ['components']
+            properties (1): ['properties']
+            gradients: ['position']
```
agoscinski, line 308: Regarding the documentation tests error, I don't know why, but that worked for me locally. I think the doctest may not end with an indentation.
```suggestion
        TensorBlock
        samples (3): ['samples']
        components (1): ['components']
        properties (1): ['properties']
        gradients: ['position']
```

python/src/equistore/block.py
```diff
@@ -229,6 +279,34 @@ def add_gradient(
         """
         Add a set of gradients with respect to ``parameters`` in this block.
 
+        Here is an example of how to use this method:
+        >>> import numpy as np
```
agoscinski, line 283:
```suggestion
        >>> import numpy as np
        >>> from equistore import TensorBlock, Labels
```

agoscinski (changes requested):
okay these changes should be last ones

python/src/equistore/block.py
agoscinski: this breaks however the doctests (which i did not check properly). i ll check for a different solution

python/src/equistore/block.py
```diff
@@ -229,6 +279,34 @@ def add_gradient(
         """
         Add a set of gradients with respect to ``parameters`` in this block.
 
+        Here is an example of how to use this method:
```
agoscinski, line 282: Seems to fix the tox documentation tests
```suggestion
        Here is an example of how to use this method:
        
```

agoscinski (approved):
Thanks for the work :pray: ! Looks good!

PR #202, opened 2023-03-14
Revamping the documentation for reduction operations

ceriottm:
* Also added reductions for blocks
* Also renamed a couple of parameters that were in bad english
* Also added an introduction text for reduction operations

----
:books: Documentation preview :books:: https://equistore--202.org.readthedocs.build/en/202/


docs/src/reference/python/operations/manipulation/samples-reduction.rst
```diff
@@ -1,10 +1,34 @@
...
 
-.. autofunction:: equistore.variance_over_samples
+.. autofunction:: equistore.var_over_samples_block
+
+.. autofunction:: equistore.var_over_samples
+
+.. autofunction:: equistore.std_over_samples_block
 
-.. autofunction:: equistore.std_over_samples
+.. autofunction:: equistore.std_over_samples
```
Luthaf, line 34: I would put the functions operating on tensormap first in this page, and leave the ones operating on blocks for the bottom of the page

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -222,33 +237,33 @@ def _reduce_over_samples_block(
...
 ) -> TensorMap:
     """
     Create a new :py:class:`TensorMap` with the same keys as as the input
     ``tensor``, and each :py:class:`TensorBlock` is obtained summing the
-    corresponding input :py:class:`TensorBlock` over the ``samples_names``
+    corresponding input :py:class:`TensorBlock` over the ``sample_names``
     indices.
 
-    "sum", "mean", "std" or "variance" reductions can be performed.
+    "sum", "mean", "std" or "var" reductions can be performed.
```
Luthaf, line 248: I would leave this as `variance`, saving 3 characters here makes the documentation harder to read

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -263,21 +278,82 @@ def _reduce_over_samples(
...
 
 
+def sum_over_samples_block(
+    block: TensorBlock, sample_names: Union[List[str], str]
+) -> TensorBlock:
+    """Sum a :py:class:`TensorBlock`, combining the samples
+    according to ``sample_names``.
+
+    This function creates a new :py:class:`TensorBlock` in which each sample is
+    obtained averaging over the ``sample_names`` indices, so that the resulting
```
Luthaf, line 288:
```suggestion
    obtained summing over the ``sample_names`` indices, so that the resulting
```

python/src/equistore/operations/reduce_over_samples.py
```diff
@@ -321,89 +397,162 @@ def sum_over_samples(
...
+
+    return _reduce_over_samples_block(
+        block=block, sample_names=sample_names, reduction="var"
     )
 
 
-def variance_over_samples(tensor: TensorMap, samples_names: List[str]) -> TensorMap:
-    r"""Compute the variance of a :py:class:`TensorMap`, grouping samples according to
-    ``samples_names``.
+def var_over_samples(tensor: TensorMap, sample_names: List[str]) -> TensorMap:
```
Luthaf, line 539: Same here, `var` => `variance`

python/src/equistore/operations/reduce_over_samples.py
ceriottm: So, this was deliberate. These mirror array operations in `numpy` and `torch` , and it is kind of inconsistent to use the same names for sum, mean, std and then use `variance` where numpy and torch use `var`.

python/src/equistore/operations/reduce_over_samples.py
ceriottm: Same as above. I think that the closer we stay to numpy syntax, the more natural it will feel to users.

python/src/equistore/operations/reduce_over_samples.py
Luthaf: Ok, fair enough

Luthaf (approved):
Looks good! I'll wait for everyone opinion on doc organization (cf message on slack) before merging.

docs/src/reference/python/operations/manipulation/samples-reduction.rst
```diff
@@ -1,10 +1,40 @@
...
+in the same way. The reduction operation loops over the samples in each block/map, 
+and combines all those that only differ by the values of the indices associated 
+with the names listed in the ``sample_names`` argument. 
+One way to see these operations is that the sample indices describe the non-zero
+entries in a *sparse* array, and the reduction acts much like :func:`numpy.sum`, 
+where ``sample_names`` plays the same role as the ``axis`` argument. 
+Whenever gradients are present, the reduction is performed also on the gradients.
+
+See also :py:func:`equistore.sum_over_samples_block` and :py:func:`equistore.sum_over_samples`  for 
+a detailed discussion with examples.
```
PicoCentauri, line 18: Very nice. I think an even better place for this is the module docstring in python/src/equistore/operations/reduce_over_samples.

We can then just include the text here.

PicoCentauri: I moved everything into the source code.

ceriottm:
OK to merge?

PR #205, opened 2023-03-14
added example to ones_like

soniasalomoni:
----
:books: Documentation preview :books:: https://equistore--205.org.readthedocs.build/en/205/

### DIFF
--- a/python/src/equistore/operations/ones_like.py
+++ b/python/src/equistore/operations/ones_like.py
@@ -1,6 +1,7 @@
 from typing import List, Union
+import numpy as np
 
-from equistore import TensorBlock, TensorMap
+from equistore import TensorBlock, TensorMap, Labels
 
 from . import _dispatch
 from ._utils import _check_parameters_in_gradient_block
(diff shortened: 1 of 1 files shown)


python/src/equistore/operations/ones_like.py
```diff
@@ -1,6 +1,7 @@
 from typing import List, Union
+import numpy as np
```
agoscinski, line 2: The linter seems not to be soo smart, so it throws an error here
https://github.com/lab-cosmo/equistore/actions/runs/4417733640/jobs/7745861938#step:7:532
In the code you don't need numpy, so it complains. It apparently cannot see that it is used in the doc.
This should fix the problem
```suggestion
import numpy as np # noqa: F401
```

python/src/equistore/operations/ones_like.py
```diff
@@ -1,6 +1,7 @@
...
 
-from equistore import TensorBlock, TensorMap
+from equistore import TensorBlock, TensorMap, Labels
```
agoscinski, line 4: Same issue and fix as above
https://github.com/lab-cosmo/equistore/actions/runs/4417733640/jobs/7745861938#step:7:533

python/src/equistore/operations/ones_like.py
```diff
@@ -46,6 +160,9 @@ def ones_like_block(
                        If empty list ``[]`` no gradients information are copied.
     :param requires_grad: If autograd should record operations for the returned tensor.
                           This option is only relevant for torch.
+
+
+
```
agoscinski, line 165: Nothing super important, but could you remove these empty lines?

agoscinski (commented):
Thanks!

DavideTisi:
to make the linter happy talking to @Luthaf we thought of adding the import in the example in the docsstring

python/src/equistore/operations/ones_like.py
agoscinski: Okay talked with @DavideTisi and @Luthaf we move this import to the docs, so you can remove the import here and insert it into the docstring
```suggestion
```

python/src/equistore/operations/ones_like.py
agoscinski: same here we remove the Labels import

python/src/equistore/operations/ones_like.py
```diff
@@ -20,6 +21,119 @@ def ones_like(
                        If empty list ``[]`` no gradients information are copied.
     :param requires_grad: If autograd should record operations for the returned tensor.
                           This option is only relevant for torch.
+
+    Here is an example using this function:
+
+    First we create a ``TensorMap`` with just one block with two gradients, named ``alpha`` and `beta`.
+
```
agoscinski, line 28: and add the import here
```suggestion
    >>> import numpy as np
    >>> from equistore import Labels
```

python/src/equistore/operations/ones_like.py
agoscinski:
```suggestion
from equistore import TensorBlock, TensorMap
```

Luthaf:
Closed in favor of #211

Issue #207, opened 2023-03-15
Check that operations do not modify the input tensors in-place


jwa7:
Following on from the issue raised in #206, I am opening this issue to ensure all operations are checked for in-place modifications, and unit tests are updated to catch such errors.

For the unit tests for every operation, add a simple `assert equistore.equal(tensor, tensor)` after the `tensor` has been used in an operation to ensure that it has not been modified by that operation. This is a simple but very important test, as in-place modifications of TensorMaps can break an entire workflow.

Luthaf:
> For the unit tests for every operation, add a simple assert equistore.equal(tensor, tensor) after the tensor has been used in an operation to ensure that it has not been modified by that operation. 

I don't see how that helps: even if `tensor` has been modified in-place, it is still equal to itself. We could make a copy before the operation and check the tensor after the operation against the copy before though.


---

(I removed the bug label, this is not an issue about a specific bug, but about how to prevent some of them)

jwa7:
Closed by issue #212 and PR #208

PR #208, opened 2023-03-15
Fix inplace operation bug, add tests

### LINKED ISSUE #206: Bug in `subtract` and `add`
There is a bug in `subtract`, and by extension `add`. The block values are modified in place, meaning that the tensors passed to the function are modified.

To recreate:

```py
from equistore import Labels, TensorBlock, TensorMap 

tensor = TensorMap(
    keys=Labels(names=("_",), values=np.array([[0]])),
    blocks=[
        TensorBlock(
            values=np.full((3, 1, 1), 1.0),
            samples=Labels(["samples"], np.array([[0], [2], [4]])),
            components=[Labels(["components"], np.array([[0]]))],
            properties=Labels(["properties"], np.array([[0]])),
        )
    ],
)

tensor.block(0).values
>>> 
array([[[1.]],

       [[1.]],

       [[1.]]])
```
Subtract tensor from itself. The values should be zero
```py
new_tensor = equistore.subtract(tensor, tensor)
new_tensor.block(0).values
>>>
array([[[-2.]],

       [[-2.]],

       [[-2.]]])
```
Making a copy gives the correct answer:
```py
new_tensor = equistore.subtract(tensor, tensor.copy())
new_tensor.block(0).values
>>>
array([[[0.]],

       [[0.]],

       [[0.]]])
```
But the values of `tensor` have been modified. They should all still be ``+1``, but instead are now ``-1``. This is because in `add()` the negative of the block values is taken, instead of creating a copy.
```py
tensor.block(0).values
>>>
array([[[-1.]],

       [[-1.]],

       [[-1.]]])
```

We need to be very careful with in-place modifications as they can break a whole workflow. We should check all operations, not just `add` and `subtract` to make sure that if returning a new TensorMap, the input TensorMaps to the function are not modified in place. Our unit tests should catch these too - a simple `assert equal(tensor, tensor)` after the `tensor` has been used in an operation, to make sure it has not been modified in the process, should suffice.

jwa7:
Fixes #206 #207 

----
:books: Documentation preview :books:: https://equistore--208.org.readthedocs.build/en/208/

### DIFF
--- a/python/src/equistore/operations/multiply.py
+++ b/python/src/equistore/operations/multiply.py
@@ -34,10 +34,11 @@ def multiply(A: TensorMap, B: Union[float, TensorMap]) -> TensorMap:
 
     :return: New :py:class:`TensorMap` with the same metadata as ``A``.
     """
-
+    A = A.copy()
     blocks = []
     if isinstance(B, TensorMap):
         _check_maps(A, B, "multiply")
+        B = B.copy()
         for key, blockA in A:
             blockB = B.block(key)
             _check_blocks(
--- a/python/src/equistore/operations/multiply.py
+++ b/python/src/equistore/operations/multiply.py
@@ -34,13 +34,12 @@ def multiply(A: TensorMap, B: Union[float, TensorMap]) -> TensorMap:
 
     :return: New :py:class:`TensorMap` with the same metadata as ``A``.
     """
-    A = A.copy()
     blocks = []
     if isinstance(B, TensorMap):
         _check_maps(A, B, "multiply")
-        B = B.copy()
-        for key, blockA in A:
-            blockB = B.block(key)
+        for key in A.keys:
+            blockA = A[key]
+            blockB = B[key]
             _check_blocks(
                 blockA,
                 blockB,
--- a/python/src/equistore/operations/subtract.py
+++ b/python/src/equistore/operations/subtract.py
@@ -33,7 +33,7 @@ def subtract(A: TensorMap, B: Union[float, TensorMap]) -> TensorMap:
     """
     if isinstance(B, TensorMap):
         _check_maps(A, B, "subtract")
-        B = multiply(A=B, B=-1)
+        B = multiply(B, -1)
     else:
         # check if can be converted in float (so if it is a constant value)
         try:
--- a/python/src/equistore/operations/subtract.py
+++ b/python/src/equistore/operations/subtract.py
@@ -3,6 +3,7 @@ from typing import Union
 from ..tensor import TensorMap
 from ._utils import _check_maps
 from .add import add
+from .multiply import multiply
 
 
 def subtract(A: TensorMap, B: Union[float, TensorMap]) -> TensorMap:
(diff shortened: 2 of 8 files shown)


python/tests/operations/add.py
```diff
@@ -292,6 +300,37 @@ def test_self_add_error(self):
             str(cm.exception),
             "B should be a TensorMap or a scalar value. ",
         )
+        self.assertTrue(equistore.equal(A, A_copy))  # check not modified in place
```
Luthaf, line 303: I'm not sure we need to check every where that nothing have been modified, this is already checked in the two test functions above

python/tests/operations/multiply.py
```diff
@@ -298,6 +310,42 @@ def test_self_multiply_error(self):
             str(cm.exception),
             "B should be a TensorMap or a scalar value. ",
         )
+        self.assertTrue(equistore.equal(A, A_copy))
```
Luthaf, line 313: same comment as for `add`, we don't need to check this in all tests

python/src/equistore/operations/multiply.py
```diff
@@ -34,10 +34,11 @@ def multiply(A: TensorMap, B: Union[float, TensorMap]) -> TensorMap:
 
     :return: New :py:class:`TensorMap` with the same metadata as ``A``.
     """
-
+    A = A.copy()
```
Luthaf, line 37: I don't think this is necessary. The code below (_multiply_block_constant/_multiply_block_block) should not make any in-place modification, so this would just be wasting memory

Luthaf (changes requested):
I don't think unconditionally copying the input is the best way to fix the problem, since it introduces a additional large cost to these operations. 

Since only `substract` has the bug, we should fix substract by either writing it in a similar way as `add`, or as `add(A, mul(B, -1.0))`. My preference would be to write it in a similar way to add: there is a bit more code duplication but we control the performance characteristics.

jwa7:
Thanks for the comments @Luthaf . I have removed the unnecessary copying of TensorMaps, removed the redundant test case in each of the files, and slimmed down the number of places I perform the `self.assertTrue(equistore.equal(A, A_copy))` checks.

python/src/equistore/operations/subtract.py
```diff
@@ -30,15 +31,9 @@ def subtract(A: TensorMap, B: Union[float, TensorMap]) -> TensorMap:
...
-
     if isinstance(B, TensorMap):
         _check_maps(A, B, "subtract")
-        for key, blockB in B:
-            B.block(key).values[:] = -1 * B.block(key).values[:]
-            for parameter in blockB.gradients_list():
-                B.block(key).gradient(parameter).data[:] = (
-                    -B.block(key).gradient(parameter).data[:]
-                )
+        B = multiply(A=B, B=-1)
```
Luthaf, line 36:
```suggestion
        B = multiply(B, -1)
```

Luthaf: So this is fine for now, but also not the best solution: we are still creating a new tensormap to hold `-B`, but we don't have to. I'll open an issue to improve later

jwa7:
Once #210 has been merged, this branch should be rebased and updated to ensure the updated `pow` tests are there

DavideTisi (approved):

PR #209, opened 2023-03-15
New absolute operation for TensorMaps

DivyaSuman14:
Absolute function for TensorMaps and TensorBlocks. Returns a new TensorMap with the absolute values of the given TensorMap. The returned block have the same samples, properties and components as that of the given TensorMap.

----
:books: Documentation preview :books:: https://equistore--209.org.readthedocs.build/en/209/

### DIFF
--- a/python/src/equistore/operations/abs.py
+++ b/python/src/equistore/operations/abs.py
@@ -11,12 +11,12 @@ def abs(A: TensorMap) -> TensorMap:
     r"""
     Return a new :py:class:`TensorMap` with the same metadata as A and absolute
     values of ``A``.
-    
+
     .. math::
         A \rightarrow = \vert A \vert
 
     If gradients are present in ``A``:
-    
+
     .. math::
         \nabla(A) \rightarrow \nabla(\vert A \vert) = (A/\vert A \vert)*\nabla A
 
--- a/python/src/equistore/operations/abs.py
+++ b/python/src/equistore/operations/abs.py
@@ -8,7 +8,7 @@ from . import _dispatch
 
 
 def abs(A: TensorMap) -> TensorMap:
-    """
+    r"""
     Return a new :py:class:`TensorMap` with the same metadata as A and absolute
     values of ``A``.
 
--- a/python/src/equistore/operations/abs.py
+++ b/python/src/equistore/operations/abs.py
@@ -26,7 +26,7 @@ def abs(A: TensorMap) -> TensorMap:
 
 def _abs_block(block: TensorBlock) -> TensorBlock:
     """
-    Returns a TensorBlock with the absolute values of the block values and
+    Returns a :py:class:`TensorBlock` with the absolute values of the block values and
     associated gradient data.
     """
     values = _dispatch.abs(block.values)
@@ -48,6 +48,9 @@ def _abs_block(block: TensorBlock) -> TensorBlock:
 
     for parameter, gradient in block.gradients():
         diff_components = len(gradient.components) - len(block.components)
+        # The sign_values have the same dimensions as that of the block.values.
+        # To multiply with gradient.data it is reshaped to have same dimensions
+        # as that of the gradient.data.
         new_grad = gradient.data[:] * sign_values[gradient.samples["sample"]].reshape(
             (-1,) + (1,) * diff_components + _shape
         )
--- a/python/src/equistore/operations/abs.py
+++ b/python/src/equistore/operations/abs.py
@@ -10,11 +10,10 @@ from . import _dispatch
 def abs(A: TensorMap) -> TensorMap:
     """
     Return a new :py:class:`TensorMap` with the same metadata as A and absolute
-    values of `A`.
+    values of ``A``.
 
-    ..math::
-
-        \nabla(|A|) = (A/|A|)*\nabla A
+    .. math::
+        \nabla(\vert A \vert) = (A/\vert A \vert)*\nabla A
 
     :param A: input :py:class:`TensorMap` whose absolute values are needed.
 
--- a/python/src/equistore/operations/abs.py
+++ b/python/src/equistore/operations/abs.py
@@ -1,0 +1,63 @@
+"""
+Module to find the absolute values of a :py:class:`TensorMap`, returning a new
+:py:class:`TensorMap`.
+"""
+from ..block import TensorBlock
+from ..tensor import TensorMap
+from . import _dispatch
+
+
+def abs(A: TensorMap) -> TensorMap:
+    """
+    Return a new :py:class:`TensorMap` with the same metadata as A and absolute
+    values of `A`.
+
+
+    :param A: input :py:class:`TensorMap` whose absolute values are needed.
+
+    :return: New :py:class:`TensorMap` with the same metadata as ``A`` and
+        absolute values of ``A``.
+    """
+    blocks = []
+    for key in A.keys:
+        blocks.append(_abs_block(block=A[key]))
+    return TensorMap(A.keys, blocks)
+
+
+def _abs_block(block: TensorBlock) -> TensorBlock:
+    """
+    Returns a TensorBlock with the absolute values of the block values and
+    associated gradient data.
+    """
+    values = _dispatch.abs(block.values)
+
+    result_block = TensorBlock(
+        values=values,
+        samples=block.samples,
+        components=block.components,
+        properties=block.properties,
+    )
+    if not block.has_gradient():
+        return result_block
+
+    sign_values = _dispatch.sign(block.values)
+    _shape = ()
+    for c in block.components:
+        _shape += (len(c),)
+    _shape += (len(block.properties),)
+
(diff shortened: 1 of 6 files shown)


python/tests/operations/abs.py
```diff
@@ -0,0 +1,218 @@
...
+        )
+        keys = Labels(
+            names=["key_1", "key_2"], values=np.array([[0, 0], [1, 0]], dtype=np.int32)
+        )
+        A = TensorMap(keys, [block_1, block_2])
+        tensor_abs = equistore.abs(A)
+        tensor_result = TensorMap(keys, [block_res1, block_res2])
+        print("oo", tensor_result[0].values)
+        print("pp", tensor_abs[0].values)
+        assert equistore.equal(tensor_result, tensor_abs)
```
Luthaf, line 218: This should be an allclose assert, since complex abs might have some [ULPS](https://en.wikipedia.org/wiki/Unit_in_the_last_place) differences with double sqrt.

DavideTisi: It is still weird that in both mine and @DivyaSuman14 's machine the test do not fail

Luthaf: CI is using Intel mac, maybe both of you are using M1?

python/src/equistore/operations/abs.py
```diff
@@ -0,0 +1,65 @@
...
+
+    :param A: input :py:class:`TensorMap` whose absolute values are needed.
+
+    :return: New :py:class:`TensorMap` with the same metadata as ``A`` and
+        absolute values of ``A``.
+    """
+    blocks = []
+    for key in A.keys:
+        blocks.append(_abs_block(block=A[key]))
+    return TensorMap(A.keys, blocks)
```
jwa7, line 24:
```suggestion
    blocks = []
    keys = A.keys
    for key in keys:
        blocks.append(_abs_block(block=A[key]))
    return TensorMap(keys, blocks)
```

python/src/equistore/operations/abs.py
```diff
@@ -0,0 +1,65 @@
...
+from ..tensor import TensorMap
+from . import _dispatch
+
+
+def abs(A: TensorMap) -> TensorMap:
+    """
+    Return a new :py:class:`TensorMap` with the same metadata as A and absolute
+    values of `A`.
+
+
```
jwa7, line 15: It would be good to have the maths here - it's not completely trivial what happens to the gradients upon taking the absolute so describing (at least with using a math block) what happens in the operation would be useful.

python/src/equistore/operations/abs.py
```diff
@@ -0,0 +1,65 @@
...
+    _shape = ()
+    for c in block.components:
+        _shape += (len(c),)
+    _shape += (len(block.properties),)
+
+    for parameter, gradient in block.gradients():
+        diff_components = len(gradient.components) - len(block.components)
+        # The sign_values have the same dimensions as that of the block.values.
+        # To multiply with gradient.data it is reshaped to have same dimensions
+        # as that of the gradient.data.
```
jwa7, line 53:
```suggestion
        # Reshape the sign_values to allow multiplication with gradient.data
```

python/tests/operations/abs.py
DavideTisi: I have intel

DavideTisi: But anyway I do not know I changed to allclose

python/tests/operations/abs.py
```diff
@@ -0,0 +1,216 @@
...
+            properties=Labels(["properties"], np.array([[0], [1]], dtype=np.int32)),
+        )
+        keys = Labels(
+            names=["key_1", "key_2"], values=np.array([[0, 0], [1, 0]], dtype=np.int32)
+        )
+        A = TensorMap(keys, [block_1, block_2])
+        tensor_abs = equistore.abs(A)
+        tensor_result = TensorMap(keys, [block_res1, block_res2])
+
+        assert equistore.equal(tensor_result, tensor_abs)
```
jwa7, line 48: Let's add a check here to make sure the operation is not modifying `A` in place (re: #208)
```suggestion
        A = TensorMap(keys, [block_1, block_2])
        A_copy = A.copy()
        tensor_abs = equistore.abs(A)
        tensor_result = TensorMap(keys, [block_res1, block_res2])

        assert equistore.equal(tensor_result, tensor_abs)
        assert equistore.equal(A, A_copy)
```

jwa7 (changes requested):
Requested a couple of changes but looking great, thanks @DivyaSuman14 !

python/tests/operations/abs.py
```diff
@@ -0,0 +1,218 @@
...
+                    gradient.samples,
+                    gradient.components,
+                )
+
+            abs_blocks.append(new_block)
+
+        tensor_result = TensorMap(A.keys, abs_blocks)
+        assert equistore.equal(tensor_result, tensor_abs)
+
+    def test_self_abs_tensor_complex(self):
```
DavideTisi, line 179: very nice, I didn't think about that, this is the first test with complex number i think, well done

python/tests/operations/abs.py
```diff
@@ -0,0 +1,218 @@
...
+import equistore
+import equistore.io
+from equistore import Labels, TensorBlock, TensorMap
+
+
+DATA_ROOT = os.path.join(os.path.dirname(__file__), "..", "data")
+TEST_FILE = "qm7-spherical-expansion.npz"
+
+
+class TestAbs:
```
jwa7, line 14:
```suggestion
class TestAbs:
    """
    Unit tests for the :py:func:`equistore.abs` function.
    
    Note that unit tests for values/gradients containing complex 
    numbers is currently untested.
    """
```

python/tests/operations/abs.py
```diff
@@ -0,0 +1,218 @@
...
+        )
+        A = TensorMap(keys, [block_1, block_2])
+        tensor_abs = equistore.abs(A)
+        tensor_result = TensorMap(keys, [block_res1, block_res2])
+
+        assert equistore.equal(tensor_result, tensor_abs)
+
+    def test_self_abs_tensor_components(self):
+        A = equistore.load(
+            os.path.join(DATA_ROOT, "qm7-spherical-expansion.npz"), use_numpy=True
```
DavideTisi, line 144:
```suggestion
            os.path.join(DATA_ROOT, TEST_FILE), use_numpy=True
```

python/tests/operations/abs.py
```diff
@@ -0,0 +1,218 @@
...
+        keys = Labels(
+            names=["key_1", "key_2"], values=np.array([[0, 0], [1, 0]], dtype=np.int32)
+        )
+        A = TensorMap(keys, [block_1, block_2])
+        tensor_abs = equistore.abs(A)
+        tensor_result = TensorMap(keys, [block_res1, block_res2])
+
+        assert equistore.equal(tensor_result, tensor_abs)
+
+    def test_self_abs_tensor_components(self):
```
DavideTisi, line 142: this test is a bit tautological but for me it is fine, I like it and it will me removed when we implement the finite difference. Also probably a more readable strategy than what we do with other operations, where we reconstruct everything

DavideTisi (changes requested):
A small changes in the test and it is fine

python/tests/operations/abs.py
jwa7: (or perhaps just write a simple unit test for a TensorMap with complex numbers)

DavideTisi: ?

DavideTisi (approved):

python/src/equistore/operations/abs.py
```diff
@@ -0,0 +1,67 @@
...
+    values of ``A``.
+
+    .. math::
+        \nabla(\vert A \vert) = (A/\vert A \vert)*\nabla A
+
+    :param A: input :py:class:`TensorMap` whose absolute values are needed.
+
+    :return: New :py:class:`TensorMap` with the same metadata as ``A`` and
+        absolute values of ``A``.
+    """
```
jwa7, line 22:
```suggestion
    r"""
    Return a new :py:class:`TensorMap` with the same metadata as A and absolute
    values of ``A``.
    
    .. math::
        A \rightarrow = \vert A \vert

    If gradients are present in ``A``:
    
    .. math::
        \nabla(A) \rightarrow \nabla(\vert A \vert) = (A/\vert A \vert)*\nabla A

    :param A: the input :py:class:`TensorMap`.

    :return: a new :py:class:`TensorMap` with the same metadata as ``A`` and
        absolute values of ``A``.
    """
```

jwa7 (approved):
I'd suggest a tiny modification to the docstring, but happy for you to merge once done!

python/src/equistore/operations/abs.py
DavideTisi: nice idea but we never use the -> in the others docstring

PR #210, opened 2023-03-15
Bugfix/pow withcomponents

DavideTisi:
The code was not working with a TB with more than one component, now also, the code is cleaner.
I added a test for the specific case and ported the test to pytest


python/src/equistore/operations/pow.py
```diff
@@ -45,16 +43,25 @@ def _pow_block_constant(block: TensorBlock, constant: float) -> TensorBlock:
         components=block.components,
         properties=block.properties,
     )
+    if len(block.gradients_list()) == 0:
+        return result_block
```
Luthaf, line 47: I'm not sure what we gain with this early return? If there are no gradients, the loop below will be skipped anyway

python/src/equistore/operations/pow.py
```diff
@@ -45,16 +43,25 @@ def _pow_block_constant(block: TensorBlock, constant: float) -> TensorBlock:
...
-        for isample in range(len(block.samples)):
-            isample_grad1 = np.where(gradient.samples["sample"] == isample)[0]
-            values_grad.append(
-                constant
-                * gradient.data[isample_grad1]
-                * block.values[isample] ** (constant - 1)
+        diff_components = len(gradient.components) - len(block.components)
+        values_grad.append(
+            constant
+            * gradient.data[:]
```
Luthaf, line 59: this should work as well

```suggestion
            * gradient.data
```

python/src/equistore/operations/pow.py
```diff
@@ -45,16 +43,25 @@ def _pow_block_constant(block: TensorBlock, constant: float) -> TensorBlock:
...
 
     for parameter, gradient in block.gradients():
         values_grad = []
-        for isample in range(len(block.samples)):
-            isample_grad1 = np.where(gradient.samples["sample"] == isample)[0]
-            values_grad.append(
-                constant
-                * gradient.data[isample_grad1]
-                * block.values[isample] ** (constant - 1)
+        diff_components = len(gradient.components) - len(block.components)
```
Luthaf, line 56: A tiny optimization (not sure it will have any effects though)

```suggestion
        gradient_data = gradient.data
        diff_components = len(gradient_data.shape) - len(block.values.shape)
```

Luthaf (approved):
This looks fine! 

A related question is should we warn the user when using non-linear operations on TM with components? This is often (but not always) wrong, so it would be a nice quality of life feature. Might be worth discussing in a separate issue.

python/src/equistore/operations/pow.py
DavideTisi: ah right we do it in the abs because before we compute `_shape` we do the `dispatch`

jwa7 (approved):
Looks goog!

python/src/equistore/operations/pow.py
DavideTisi: gradient_data.shape is not the len of the components

DavideTisi: same for len(block.values.shape) != len(block.components)

maybe we need to put
`len(gradient_data.shape) - len(block.values.shape)-2`

DavideTisi: sorry you are right 
`diff_components = len(gradient_data.shape) - len(block.values.shape)` works

jwa7:
Just merged #208, so this branch will require a rebase and addition of a small inplace test @DavideTisi . Once done I'm happy for you to merge this branch provided CI is happy :)

Issue #159, opened 2023-03-03
`join` does not work with gradients of 0th samples


PicoCentauri:
With the new `AtomicComposition` [calculator](https://github.com/Luthaf/rascaline/pull/154) we can produce a descriptor that has gradients where all of them are zero. These zeros are not stored explicitly but only an empty array is saved. This causes problems for the `join` function as the below example shows

```
import equistore
from equistore import Labels, TensorBlock, TensorMap

block = TensorBlock(
    values = np.zeros([2,1]),
    properties=Labels(["prop"], np.zeros([1,1], dtype=int)),
    samples=Labels(["structure"], np.array([[0], [1]])),
    components=[]
)

block.add_gradient(
    parameter="positions",
    data=np.zeros((0, 1)),
    samples=Labels(["sample", "structure"], np.empty((0, 2), dtype=np.int32)),
    components=[],
)

tensor = TensorMap(Labels.single(), [block])

equistore.join([tensor, tensor], "properties")
```

producing an error

```python
ValueError                                Traceback (most recent call last)
Cell In[43], line 20
     11 block.add_gradient(
     12     parameter="positions",
     13     data=np.zeros((0, 1)),
     14     samples=Labels(["sample", "structure"], np.empty((0, 2), dtype=np.int32)),
     15     components=[],
     16 )
     18 tensor = TensorMap(Labels.single(), [block])
---> 20 equistore.join([tensor, tensor], "properties")

File /opt/homebrew/lib/python3.11/site-packages/equistore/operations/join.py:80, in join(tensor_maps, axis)
     77 blocks_to_join = [ts.block(key) for ts in tensor_maps]
     79 if axis == "properties":
---> 80     blocks.append(_join_blocks_along_properties(blocks_to_join))
     81 elif axis == "samples":
     82     blocks.append(_join_blocks_along_samples(blocks_to_join))

File /opt/homebrew/lib/python3.11/site-packages/equistore/operations/join.py:153, in _join_blocks_along_properties(blocks)
    148 for parameter, first_gradient in first_block.gradients():
    149     # Different blocks might contain different gradient samples
    150     gradient_sample_values = np.unique(
    151         np.hstack([block.gradient(parameter).samples for block in blocks])
    152     ).tolist()
...
---> 76     raise ValueError("values parameter must be a 2D array")
     78 if isinstance(names, str):
     79     if len(names) == 0:

ValueError: values parameter must be a 2D array
```

[...]

PR #222, opened 2023-03-20
Use `keys_to_x` for join operation

### LINKED ISSUE #195: operations.join fails with non-empty components tensor blocks
Joining two TensorMaps fails when the Tensorblocks contain non-empty components

Join two tensor maps with empty components
```python
import equistore
import equistore.operations as eqst
import numpy as np

s_labels = equistore.Labels(names=["idx"],values=np.arange(1).reshape(-1,1))
components = [] #equistore.Labels(["sym"], np.asarray([[0]], dtype=np.int32))]
properties = equistore.Labels(["eV"], np.asarray([[0]], dtype=np.int32))

tensor_map_a = equistore.TensorMap(keys=equistore.Labels(["frame_a"], np.asarray([(0,)], np.int32)),\
                                   blocks=[equistore.TensorBlock(values=np.zeros(1).reshape(1,1),\
                                   samples=s_labels,components=components,properties=properties)])

tensor_map_b = equistore.TensorMap(keys=equistore.Labels(["frame_a"], np.asarray([(0,)], np.int32)),\
                                   blocks=[equistore.TensorBlock(values=np.zeros(1).reshape(1,1),\
                                                                 
                                   samples=s_labels,components=components,properties=properties)])

eqst.join([tensor_map_a,tensor_map_a],axis="properties")

# ->
# TensorMap with 1 blocks
#  keys: ['frame_a']
#          0
#

```

Join two tensor maps with non-empty components

```python
s_labels = equistore.Labels(names=["idx"],values=np.arange(1).reshape(-1,1))
components = [equistore.Labels(["sym"], np.asarray([[0]], dtype=np.int32))]
properties = equistore.Labels(["eV"], np.asarray([[0]], dtype=np.int32))

tensor_map_a = equistore.TensorMap(keys=equistore.Labels(["frame_a"], np.asarray([(0,)], np.int32)),\
                                   blocks=[equistore.TensorBlock(values=np.zeros(1).reshape(1,1,1),\
                                   samples=s_labels,components=components,properties=properties)])

tensor_map_b = equistore.TensorMap(keys=equistore.Labels(["frame_a"], np.asarray([(0,)], np.int32)),\
                                   blocks=[equistore.TensorBlock(values=np.zeros(1).reshape(1,1,1),\
                                   samples=s_labels,components=components,properties=properties)])

#
eqst.join([tensor_map_a,tensor_map_a],axis="properties")

#-> EquistoreError: invalid parameter: data and labels don't match: 
# the array shape along axis 1 is 2 but we have 1 entries for the corresponding component
```

I use python 3.10.4 and equistore commit 16ff8ea

PicoCentauri:
Fixes #195 and #159

Instead of reimplementing everything from scratch `join` now uses the `keys_to_properties` or `keys_to_samples` methods for joining a list of TensorMaps. This does solve the two issues, simplifies the code and leads to a 7-8 times speedup compared to the current master branch:

```python
tensor = equistore.load(
    os.path.join("../python/tests/data/qm7-power-spectrum.npz"), use_numpy=True
)

%timeit join([tensor,tensor], axis="properties")
```

`5.03 ms ± 82.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)`

```python
%timeit equistore.join([tensor,tensor], axis="properties")
```

`37.1 ms ± 336 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)`

The only drawback is that we now have a `tensor` label after the joining. I think this is acceptable for now. I added a TODO and we can try to simplify the labels once we have labels operations. 

----
:books: Documentation preview :books:: https://equistore--222.org.readthedocs.build/en/222/

### DIFF
--- a/python/src/equistore/operations/join.py
+++ b/python/src/equistore/operations/join.py
@@ -86,19 +86,28 @@ def join(tensor_maps: List[TensorMap], axis: str):
         keys_values += [(i,) + value for value in tensor.keys.tolist()]
 
         for _, block in tensor:
-            samples = block.samples
-            properties = block.properties
-
-            if axis == "properties" and not names_are_same:
+            # We would already raised an error if `axis == "samples"`. Therefore, we can
+            # neglect the check for `axis == "properties"`.
+            if names_are_same:
+                properties = block.properties
+            else:
                 properties = Labels(
-                    ["property"], np.arange(len(properties)).reshape(-1, 1)
+                    ["property"], np.arange(len(block.properties)).reshape(-1, 1)
                 )
 
-            new_block = TensorBlock(block.values, samples, block.components, properties)
+            new_block = TensorBlock(
+                values=block.values,
+                samples=block.samples,
+                components=block.components,
+                properties=properties,
+            )
 
             for parameter, gradient in block.gradients():
                 new_block.add_gradient(
-                    parameter, gradient.data, gradient.samples, gradient.components
+                    parameter=parameter,
+                    data=gradient.data,
+                    samples=gradient.samples,
+                    components=gradient.components,
                 )
 
             blocks.append(new_block)
--- a/python/src/equistore/operations/join.py
+++ b/python/src/equistore/operations/join.py
@@ -1,11 +1,11 @@
-from typing import List
 import functools
 import operator
+from typing import List
 
 import numpy as np
 
 from ..labels import Labels
-from ..tensor import TensorMap, TensorBlock
+from ..tensor import TensorBlock, TensorMap
 from .equal_metadata import _check_maps
 
 
@@ -17,8 +17,9 @@ def join(tensor_maps: List[TensorMap], axis: str):
     `properties` dimension and for ``axis='samples'`` they will be the along the
     samples dimension.
 
-    `join` will create an additional label `tensor` specifiying the original
-    index in the list of `tensor_maps`.
+    ``join`` will create an additional label `tensor` specifiying the original index in
+    the list of `tensor_maps`.  If `sample`/`property` names are not the same in all
+    `tensor_maps` they will be unified with a general name ``"property"``.
 
     :param tensormaps:
         sequence of :py:class:`TensorMap` for join
@@ -39,10 +40,6 @@ def join(tensor_maps: List[TensorMap], axis: str):
     if len(tensor_maps) < 1:
         raise ValueError("provide at least one `TensorMap` for joining")
 
-    if axis == "samples":
-        common_labels_names = _get_common_labels_names()
-    elif axis == "properties":
-        pass
     if axis not in ("samples", "properties"):
         raise ValueError(
             "Only `'properties'` or `'samples'` are "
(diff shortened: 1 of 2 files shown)


python/src/equistore/operations/join.py
```diff
@@ -80,155 +43,28 @@ def join(tensor_maps: List[TensorMap], axis: str):
...
-            for i_grad, sample_grad in enumerate(gradient.samples):
-                i_sample = gradient_samples.position(sample_grad)
-                gradient_data[
-                    i_sample,
-                    ...,
-                    properties_start : properties_start + len(gradient.properties),
-                ] = gradient.data[i_grad]
+    for i, tensor in enumerate(tensor_maps):
+        keys_values += [value + (i,) for value in tensor.keys.tolist()]
+        blocks += [block.copy() for _, block in tensor]
```
Luthaf, line 52: After thinking a bit more about it, I think this is where we should remove the copy. Instead of using block.copy(), recreating the block with 

```
TensorBlock(
    block.values,
    block.samples,
    block.components,
    block.properties,
)
```

should only increment the reference count of everything, not making any copy, and just create a new block sharing data. Then we have one unique copy below, which is required anyway.

PicoCentauri: If this will just increase the count, this should work here!

PicoCentauri:
I just figured out that we still something for unifying the Labels. i.e. for joining tensors with different property names we can not create a new tensor.

PicoCentauri:
Not copying the block gave us another factor of two. We are now 16 times faster than before.

python/src/equistore/operations/join.py
PicoCentauri: Yes, I did a test for a spherical expansion and an atomic composition. This worked. Also the length 0 gradients of the atomic composition are densified.

python/src/equistore/operations/join.py
```diff
@@ -117,118 +71,45 @@ def _join_labels(labels: List[Labels]) -> Labels:
...
+            properties = block.properties
 
-    result_block = TensorBlock(
-        values=_dispatch.concatenate([block.values for block in blocks], axis=0),
-        samples=_join_labels([block.samples for block in blocks]),
-        components=first_block.components,
-        properties=first_block.properties,
-    )
+            if not names_are_same:
+                if axis == "samples":
```
Luthaf, line 86: I think `axis == "samples and not names_are_same` should be an error. It's fine to loose metadata on the property axis, less so on the sample axis

PicoCentauri: Yes, I will raise an error and add a test.

python/src/equistore/operations/join.py
```diff
@@ -117,118 +71,47 @@ def _join_labels(labels: List[Labels]) -> Labels:
...
-
-        gradient_data = np.zeros(
-            (len(gradient_samples),)
-            + first_gradient.data.shape[1:-1]
-            + (len(properties),)
+    # It's fine to loose metadata on the property axis, less so on the sample axis!
+    if axis == "samples" and not names_are_same:
+        raise ValueError(
+            "Sample names are not the same! Joining along samples with unequal sample "
+            "names can lead to severre error."
```
Luthaf, line 78:
```suggestion
            "Sample names are not the same! Joining along samples with different sample "
            "names will loose information and is not supported"
```

Luthaf (approved):

PR #226, opened 2023-03-21
Disable shallow copies of TensorMap/TensorBlock

Luthaf:
They will result in double free, since two Python objects now refer to the same C pointer.

This fixes the issue in https://github.com/lab-cosmo/equistore/issues/134#issuecomment-1477760811 by making this test raise an exception instead of trying to double free the pointer.

This should only impact people trying to use `copy.copy` to make copies, instead of `TensorMap.copy` or `copy.deepcopy`.

----
:books: Documentation preview :books:: https://equistore--226.org.readthedocs.build/en/226/

### DIFF
--- a/python/tests/blocks.py
+++ b/python/tests/blocks.py
@@ -1,3 +1,5 @@
+import copy
+
 import numpy as np
 import pytest
 from numpy.testing import assert_equal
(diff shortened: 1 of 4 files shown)


python/tests/tensor.py
```diff
@@ -16,14 +18,28 @@ def large_tensor(self):
...
+        del clone
+
+        assert id(other_clone.block(0).values) != block_1_values_id
+        assert_equal(other_clone.block(0).values, np.full((3, 1, 1), 1.0))
 
-        assert_equal(copy.block(0).values, np.full((3, 1, 1), 1.0))
+        # Using copy.copy
+        msg = "shallow copies of TensorMap are not possible, use a deepcopy instead"
+        with pytest.raises(ValueError, match=msg):
+            copy.copy(other_clone)
```
PicoCentauri, line 42: same as above

```suggestion
def test_copy_error(self, tensor):
        msg = "shallow copies of TensorMap are not possible, use a deepcopy instead"
        with pytest.raises(ValueError, match=msg):
            tensor.copy()
```

python/tests/blocks.py
```diff
@@ -194,19 +196,35 @@ def test_copy(self):
...
-        assert_equal(copy.values, np.full((3, 3, 2), 2.0))
-        assert copy.samples.names == ("samples",)
-        assert len(copy.samples) == 3
-        assert tuple(copy.samples[0]) == (0,)
-        assert tuple(copy.samples[1]) == (2,)
-        assert tuple(copy.samples[2]) == (4,)
+        # Using copy.copy
+        msg = "shallow copies of TensorBlock are not possible, use a deepcopy instead"
+        with pytest.raises(ValueError, match=msg):
+            copy.copy(other_clone)
```
PicoCentauri, line 227: Can't we split this in an extra test. Feels like independant. Sorry I am a fan of small tests with single asserts.

PicoCentauri (changes requested):
Yes nice. I would only split the tests a bit to make same smaller and easier to follow.

python/tests/blocks.py
Luthaf: fair point

Luthaf: done!

PicoCentauri (approved):
Nice!

PR #227, opened 2023-03-21
Gradients as blocks

frostedoyster:
Sold my soul in exchange for compilation

----
:books: Documentation preview :books:: https://equistore--227.org.readthedocs.build/en/227/

### DIFF
--- a/equistore-core/src/blocks.rs
+++ b/equistore-core/src/blocks.rs
@@ -29,23 +29,14 @@ impl<'a, T> IntoIterator for &'a ImmutableVec<T> {
     }
 }
 
-/// Single data array with the corresponding metadata inside a `TensorBlock`
-#[derive(Debug)]
-pub struct BasicBlock {
-    pub data: eqs_array_t,
-    pub samples: Arc<Labels>,
-    pub components: ImmutableVec<Arc<Labels>>,
-    pub properties: Arc<Labels>,
-}
-
 fn check_data_and_labels(
     context: &str,
-    data: &eqs_array_t,
+    values: &eqs_array_t,
     samples: &Labels,
     components: &[Arc<Labels>],
     properties: &Labels,
 ) -> Result<(), Error> {
-    let shape = data.shape()?;
+    let shape = values.shape()?;
 
     if shape.len() != components.len() + 2 {
         return Err(Error::InvalidParameter(format!(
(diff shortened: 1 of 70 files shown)


frostedoyster:
@Luthaf I think the next step would be to rename all the gradient "data" to "values" everywhere in the C, C++, and Python APIs and to change the documentation accordingly. What do you think?

equistore-core/src/c_api/blocks.rs
```diff
@@ -186,25 +186,25 @@ pub unsafe extern fn eqs_block_labels(
         }
 
         let values_gradients = CStr::from_ptr(values_gradients).to_str().unwrap();
-        let basic_block = match values_gradients {
-            "values" => (*block).values(),
+        let block = match values_gradients {
+            "values" => (*block).get_self(),
```
Luthaf, line 190: Using a reference does not work here?

```suggestion
            "values" => &(*block),
```

equistore-core/src/blocks.rs
```diff
@@ -104,152 +95,91 @@ fn check_component_labels(components: &[Arc<Labels>]) -> Result<(), Error> {
...
     /// initialized without any gradients.
     pub fn new(
-        data: impl Into<eqs_array_t>,
+        values: eqs_array_t,
         samples: Arc<Labels>,
         components: Vec<Arc<Labels>>,
         properties: Arc<Labels>,
     ) -> Result<TensorBlock, Error> {
+        check_values_and_labels(
+            "values and labels don't match", &values, &samples, &components, &properties
```
Luthaf, line 122:
```suggestion
            "data and labels don't match", &values, &samples, &components, &properties
```

I would prefer to keep "data" in the error messages, since we are still talking about "data & metadata". The fact that you access the data through the `values` attribute feels unrelated

equistore-core/src/blocks.rs
```diff
@@ -104,152 +95,91 @@ fn check_component_labels(components: &[Arc<Labels>]) -> Result<(), Error> {
...
+            components, 
+            properties,
             gradients: HashMap::new(),
             gradient_parameters: Vec::new(),
         })
     }
 
     /// Try to copy this `TensorBlock`. This can fail if we are unable to copy
-    /// one of the underlying `eqs_array_t` data arrays
+    /// one of the underlying `eqs_array_t` values arrays
```
Luthaf, line 137: Same here, the doc could keep using data, and the data is passed in the `values` argument

Luthaf:
> @Luthaf I think the next step would be to rename all the gradient "data" to "values" everywhere in the C, C++, and Python APIs and to change the documentation accordingly. What do you think?

Yes, that should work. 

We also need a way to access nested gradients, I think a nice way to do it is to allow the `values_gradients` parameter in the C API to be `"positions/positions"` or `"cell/positions"`, where `/` indicates that we want the gradient of a gradient. We can then expose that functionality to C++/Rust/Python, and add some tests around it.

equistore-core/src/c_api/blocks.rs
frostedoyster: Well, it didn't work yesterday but it does now... Surely I'd messed something up. I'll change it

Luthaf: Maybe it was for the mutable reference? mutable and immutable references have slightly different rules for re-borrowing them. Can you also try this for the mutable one & we'll figure out a solution if rustc is not happy 🦀

Luthaf: Oh, you did that already =)

frostedoyster: I don't know what happened because yesterday the mutable one was not working so I can definitely say that I have no idea what I'm doing

equistore-core/src/blocks.rs
frostedoyster: I've tried going through all the comments but, in some cases, it is difficult to tell whether I should be using values or data... simply because they're fundamentally the same thing. Should we rename everything to data (or to values)? In case we go for data, it will be a massive breaking change, but I think this is the correct thing to do in the long run

Luthaf: So in my mind "values" in in opposition to "gradients", while "data" is in opposition to "metadata". I can help going through the code to find the best one, but I don't think we can use only one of them everywhere here

frostedoyster: Ok, in that case I'll try my best and then you'll check

frostedoyster (commented):
@Luthaf I changed the gradient access as you suggested, but I don't think this fits well with our interfaces. To be discussed in person. I've also written a Python test (that doesn't work for now) which should exemplify how adding a gradient of a gradient should work in my head

python/tests/nested_gradients.py
```diff
@@ -0,0 +1,34 @@
...
+            data = grad, 
+            samples = Labels(
+                ["sample"],
+                np.array([[0]])
+            ),
+            components = []
+        )
+        block.gradient("gradient").add_gradient(
+            parameter = "gradient_of_gradient", 
+            data = grad,
```
Luthaf, line 28:
```suggestion
            data = grad_grad, 
```

frostedoyster:
@Luthaf I'm working on the Python API. We have issues trying to build gradients of gradients because calling `block.gradient("something")` returns a view of the gradient TensorBlock, and you can't add a new gradient to that. At the moment, I don't see a way to create gradients of gradients with what we have now... We would probably need something along the lines of `add_gradient(t_block: TensorBlock)`.

Luthaf:
Right. We could modify Python's TensorBlock to track separately "is this block inside a TensorMap" (i.e. we can no longer add gradients) and "does this block have a parent" (i.e. we need to keep it alive). 

It might also make sense to have the check about "is this block inside a TensorMap" done in equistore-core (adding a boolean flag to TensorBlock, which is recursively updated when constructing a TensorMap.

equistore/src/c_api.rs
```diff
@@ -400,21 +400,23 @@ extern "C" {
...
     pub fn eqs_block_labels(
         block: *const eqs_block_t,
-        values_gradients: *const ::std::os::raw::c_char,
         axis: usize,
         labels: *mut eqs_labels_t,
     ) -> eqs_status_t;
     #[must_use]
-    #[doc = " Get the array handle for either values or one of the gradient in this `block`.\n\n The `values_gradients` parameter controls whether this function looks up\n labels for `\"values\"` or one of the gradients in this block.\n\n @param block pointer to an existing block\n @param values_gradients either `\"values\"` or the name of gradients to lookup\n @param data pointer to an empty `eqs_array_t` that will be set to the\n             requested array\n\n @returns The status code of this operation. If the status is not\n          `EQS_SUCCESS`, you can use `eqs_last_error()` to get the full\n          error message."]
-    pub fn eqs_block_data(
+    #[doc = " Comments..."]
```
frostedoyster, line 410: I've run the script about 50 times but this doesn't change...

frostedoyster: It did finally change around the 70th time

Luthaf:
> Especially in block_mut.rs where I had to change a few <'_> to <'a> and I don't know what that does.

this sounds bad ^_^

`'_` is indicating an [elided lifetime](https://doc.rust-lang.org/reference/lifetime-elision.html), meaning the lifetime of the returned object is tied to a borrow of `&self` (i.e. `&'_ self`). By returning a object with lifetime `'a`, you are no longer constraining the lifetime or borrowing the parent object, which can result in undefined behavior. 

I'll have a look, but my guess would be that we need a separate type for "gradient blocks in a TensorMap" and  "gradient blocks in a TensorBlock".

Luthaf:
> @Luthaf, are you happy with the core changes? If yes, I will do the rebasing and take a closer look at the PR.

I'm finishing the rebase of this right now! It required some more changes,notably to the serialization format. I'll send a new PR replacing this one in a bit

Luthaf:
Moved to #253

Issue #225, opened 2023-03-21
Enforce that all blocks's data have the same origin


Luthaf:
With the new implementation of `join` in #222, we could end up building a TensorMap where half the blocks are on torch tensors on GPU, and half the blocks are numpy arrays. We should protect against that, and enforce when constructing a TensorMap that all blocks have at least the same data origin.

Luthaf:
fixed by #228

PR #228, opened 2023-03-21
Ensure all blocks in a TensorMap have the same data origin

frostedoyster:
Addresses #225

----
:books: Documentation preview :books:: https://equistore--228.org.readthedocs.build/en/228/


Luthaf (commented):
This looks nice, could you add a test checking that the error message is actually thrown? 

Either [around here](https://github.com/lab-cosmo/equistore/blob/734cc45f9403fdc37cfc6e8ecd44d7cb9837c637/equistore-core/src/tensor/mod.rs#L256), adding two new arrays types with different origins (I can help with that); or in Python, using both torch & numpy arrays.

Luthaf:
Thanks!

Issue #134, opened 2023-02-22
Benchmarking operations


PicoCentauri:
Performance is a crucial part when performing operations ond equistore objects. This is important for repeating tasks during the training of a model but also when an model is applied during a simulation. To track the performance of our operation functions @ceriottm suggested to add benchmarks.

I support this idea and anybody who would like to start on this can give it a try. I suggest we use something like [ASV](https://asv.readthedocs.io/en/stable/index.html) to also track the speed over over lifetime. But, I am also open for other ideas how we can realize benchmarking.

agoscinski:
Might be useful to create a utility to generate random TensorMaps.

agoscinski:
First benchmark in PR https://github.com/lab-cosmo/equistore/pull/203

DavideTisi:
very nice is also possible to add some references result? I mean it would be nice to compare the result we have with a plain numpy implementation

agoscinski:
For that we need to also add a benchmark for numpy which should not be much work. let me show the file I used for the benchmark
```python
class TimeSuite:
    def setup(self):
        self.tm = tensor_map() # function from tests

    def time_equistore_add(self):
        equistore.add(self.tm, self.tm)
```

So your would need to add a benchmark for the numpy operation, which would be something like 

```python
    def time_numpy_add(self):
        for block in self.tm.blocks():
            block.values + block.values
            for _, gradient in block.gradients():
                gradient.data + gradient.data
```

Then you get as output

```
benchmarks.TimeSuite.time_equistore_add [intel/virtualenv-py3.8]
  12.1±1ms
  started: 2023-03-21 11:55:37, duration: 406ms

benchmarks.TimeSuite.time_numpy_add [intel/virtualenv-py3.8]
  298±20μs
  started: 2023-03-21 11:55:37, duration: 405ms
```

You can also compare directly different results from different commits which is super nice

```
$ asv continuous b8a65fa2 ad767286
· Creating environments
· Discovering benchmarks
· Running 10 total benchmarks (2 commits * 1 environments * 5 benchmarks)
[  0.00%] · For project commit b8a65fa2 <stash~1> (round 1/2):
[  0.00%] ·· Building for virtualenv-py3.8....................
[  0.00%] ·· Benchmarking virtualenv-py3.8
[ 10.00%] ··· Running (benchmarks.TimeSuite.time_equistore_add--)....
[ 25.00%] · For project commit ad767286 <master> (round 1/2):
[...]

PR #229, opened 2023-03-21
Embedding asv benchmark framework

agoscinski:
Result from discussion in issue #134 and PR #203 

I implemented two ways one an set up asv (both usable from tox using asv-toxenv and asv-asvenv, one uses the current python venv which allows us to use the tox venv, the other uses the venv defined in the asv-asvenv.conf.json file. The second way allows us to run benchmarks for certain branches or commits which I find to be a super useful feature, since you can also run on commits that did not have  asv set up. This we cannot do with the conf file used in asv-toxenv.

One can run the benchmarks for all commits in a branch 
```
tox -e asv-asvenv -- <branch>
```
or more useful to run it for one commit (even before the asv was installed)
```
tox -e asv-asvenv -- <commit>^!
```

Somehow it is not installing the equistore correctly for different commits. The file path is always in most recent commit id for me
 ```
/home/alexgo/lib/equistore/python/asv_benchmarks/.asv/env/257290e6b526ec9a4c873ad6daa76c75466a3d08/lib/python3.8/site-packages/equistore/__init__.py
````
if I use
```
tox -e asv-asvenv -- --verbose ad76728666daaf4cdf7dd94e9b22fabaae011a9b^!
```

----
:books: Documentation preview :books:: https://equistore--229.org.readthedocs.build/en/229/

### DIFF
--- a/python/asv_benchmarks/asv-asvenv.conf.json
+++ b/python/asv_benchmarks/asv-asvenv.conf.json
@@ -1,0 +1,206 @@
+{
+    // The version of the config file format.  Do not change, unless
+    // you know what you are doing.
+    "version": 1,
+
+    // The name of the project being benchmarked
+    "project": "equistore",
+
+    // The project's homepage
+    "project_url": "https://lab-cosmo.github.io/equistore/latest/",
+
+    // The URL or local path of the source code repository for the
+    // project being benchmarked
+    "repo": "../../",
+
+    // The Python project's subdirectory in your repo.  If missing or
+    // the empty string, the project is assumed to be located at the root
+    // of the repository.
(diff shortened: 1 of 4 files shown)


agoscinski:
I close the PR, because the development on this is stale. My goal is to try out pytest-benchmarks to see if the integration is easier than with asv. pytest-benchmarks has the additional functionality to plot distributions of benchmark results which is handy in some cases. But my main hope is that it will be easier for people to write benchmarks when they already are familiar with the pytest framework. Because it is always quite opaque when something is re-executed or cached in these benchmark suites. In pytest it is also opaque, but we already use it for testing, so people do not need to get acquainted two times with a framework.

PR #230, opened 2023-03-22
Write "getting started" documentation

Luthaf:
This PR splits of #91, and do not include the how-to/tutorial which is worked on over there. Instead, this PR focusses on the "getting started" section of the documentation, including an explanation of the goals of this code and the core classes of equistore. 

Future work on the documentation should include more practical how-to and tutorials, but this should already be a good starting point!


docs/src/get-started/concepts.rst
```diff
@@ -0,0 +1,239 @@
+.. _userdoc-core-concepts:
+
+Core concepts
+=============
+
+This page explains the main types and core concepts you need to understand to
+use equistore, without tying it to how a specific library is using equistore to
+store it's own data.
```
rosecers, line 8:
```suggestion
use equistore. We discuss equistore without the context
of any specific library, but provide as example, but provide 
an illustrative example later with SOAP to give further context.
```

While staying general at first is great, people may really benefit from one concrete demonstration.

docs/src/get-started/concepts.rst
```diff
@@ -0,0 +1,239 @@
...
+.. py:currentmodule:: equistore
+
+.. _concept-TensorMap:
+
+TensorMap
+^^^^^^^^^
+
+The core type of equistore is the :py:class:`TensorMap`: a high dimensional
+block-sparse tensor containing both data and metadata. A TensorMap contains a
+list of block (represented as :ref:`concept-TensorBlock`), each associated with
```
rosecers, line 19:
```suggestion
list of blocks (represented as :ref:`concept-TensorBlock`), each associated with
```

docs/src/get-started/concepts.rst
```diff
@@ -0,0 +1,239 @@
...
+
+
+Different key dimensions can have different purposes, but some typical keys
+dimensions you'll encounter when working with atomistic data are the following:
+
+- **atomic species dimensions**: when using a one-hot encoding of different
+  atomic species (or atomic elements) in a structure, the resulting data is
+  sparse, containing implicit zeros if a given species is not present in a
+  structure. This is the case of the ``species_center`` and various
+  ``species_neighbor`` key dimensions produced by `rascaline`_.
```
rosecers, line 46:
```suggestion
- **atomic species**: for different elements, we often encode their
  using one-hot encodings. However, when a species is not present
  in a dataset, traditional numpy-like formats will be sparse, with 
  implicit zeros wherever the species is not present.
  This is the case of the ``species_center`` and various
  ``species_neighbor`` key dimensions produced by `rascaline`_.
```

docs/src/get-started/concepts.rst
```diff
@@ -0,0 +1,239 @@
...
+
+Labels
+^^^^^^
+
+A fundamental part of equistore is to carry simultaneously the data used in
+machine learning and the associated metadata. The first kind of metadata we
+encountered was the keys of a :py:class:`TensorMap`, stored as an instance of
+the :py:class:`Labels` class. This class is also used to store all other
+metadata in equistore, i.e. all the metadata associated with a given
+:py:class:`TensorBlock`.
```
rosecers, line 71:
```suggestion
machine learning and the associated metadata. :py:class:`Labels` 
serves as a superclass for many forms of metadata, including
our :py:class:`TensorMap` keys.
```

docs/src/get-started/concepts.rst
```diff
@@ -0,0 +1,239 @@
...
+
+The final core object of equistore is the :py:class:`TensorBlock`, containing a
+dense array of data and metadata describing the different axes of this array.
+The simplest possible TensorBlock is represented below, and contains three things:
+
+- a 2-dimensional data array;
+- metadata describing the rows of this array, called **samples** and stored as a
+  set of :py:class:`Labels`;
+- metadata describing the columns of this array, called **properties**, also
+  stored as a set of :py:class:`Labels`.
```
rosecers, line 108:
```suggestion
- `Data`: a 2-dimensional data array;
- `Samples`: metadata describing the rows of this array, called **samples** and stored as a
  set of :py:class:`Labels`;
- `Properties`: metadata describing the columns of this array, called **properties**, also
  stored as a set of :py:class:`Labels`.
```

docs/src/get-started/concepts.rst
```diff
@@ -0,0 +1,239 @@
...
+while properties store information about **how** these objects are represented.
+Taking a couple of examples for clarity:
+
+- if a TensorBlock is storing an atom-centered machine learning representation,
+  the samples would contain the index of the atom ``"center"`` and the index of
+  the corresponding ``"structure"``; while the properties would contain
+  information about the e.g. the basis functions used to define the
+  representation. The :ref:`Labels figure <fig-labels>` above contains an
+  example of samples and properties that one would find in machine learning
+  representation.
```
rosecers, line 120: Can you start with a simpler example that this? It's a bit "being thrown into the deep end" on a complex thought

docs/src/get-started/concepts.rst
```diff
@@ -0,0 +1,239 @@
...
+- Alternatively, for a TensorBlock storing the energy of a list of structures,
+  the samples would contain only a single ``"structure"`` index, and the
+  properties would contain a single ``"energy"`` dimension with a single entry.
+
+In general, for a 2-dimensional data array, the value at index ``(i, j)`` is
+described by the ``i``'th entry of the samples and the ``j``'th entry of the
```
rosecers, line 126: This renders weird

docs/src/get-started/concepts.rst
```diff
@@ -0,0 +1,239 @@
...
+.. _Rust's ndarray: https://docs.rs/ndarray/
+.. _equistore::Array: ../reference/rust/equistore/trait.Array.html
+
+Gradients
+---------
+
+In addition to storing data and metadata together, a :py:class:`TensorBlock` can
+also store values and gradients together. The gradients are stored in another
+:py:class:`TensorBlock`, associated with a **parameter** name, describing with
+respect to **what** the gradients are taken. Regarding metadta, the gradient
```
rosecers, line 173:
```suggestion
respect to **what** the gradients are taken. Regarding metadata, the gradient
```

docs/src/get-started/concepts.rst
```diff
@@ -0,0 +1,239 @@
...
+with respect to positions of an equivariant representation based on spherical
+harmonics will need both the gradient direction :math:`x/y/z` and the spherical
+harmonics :math:`m` as components. This has two implications:
+
+- :py:class:`TensorBlock` can have an arbitrary number of components, which will
+  always occur *"in between"* samples and properties metadata;
+- when values in a :py:class:`TensorBlock` have components, and gradient with
+  respect to some parameter would add more components, the resulting gradient
+  components will contain first the new, gradient-specific components, and later
+  all of the components already present in the values.
```
rosecers, line 239: I would spin this into a new page for `What's in a Tensor Block?`, as these are as core of concepts like `TensorMap`, `Labels`, and `TensorBlock`, and keeping the page shorter will help with comprehension.

docs/src/get-started/goals.rst
```diff
@@ -0,0 +1,98 @@
+.. _userdoc-goals:
+
+Goals of this library
+=====================
+
+With the creation of equistore, we have three main use cases in mind:
+
+1. provide an interchange format for the atomistic machine learning ecosystem,
```
rosecers, line 8: Is `interchange format` a standardized term in the field? If not, it's a weird wording.

docs/src/get-started/goals.rst
```diff
@@ -0,0 +1,98 @@
...
+arrays. We also store both data and gradients of this data with respect to
+arbitrary parameters together, enabling for example training of models using
+energy, forces and virial. Finally, we also make sure that the data storage is
+as efficient as possible and can exploit the inherent sparsity of atomistic
+data, in particular in gradients.
+
+As a developer a library in the atomistic machine learning ecosystem, you can
+provide conversion functions to and from equistore
+:py:class:`equistore.TensorMap` to enable using your library in conjunction with
+the rest of the equistore-compatible libraries!
```
rosecers, line 46: Where would these conversion functions live? That is unclear.

docs/src/get-started/goals.rst
```diff
@@ -0,0 +1,98 @@
...
+prototyping tool for new models. While it is possible to use equistore to only
+exchange data between libraries (and immediately convert everything to library
+specific format); we also provide tools to operate on equistore data, staying in
+the equistore format.
+
+We call these tools :ref:`operations <python-api-operations>`, and they
+available in the Python interface to equistore. By using combining multiple
+operations, you can build custom machine learning models, using data and
+representations coming from arbitrary equistore-compatible libraires in the
+ecosystem.
```
rosecers, line 78: It's unclear the benefit of keeping things in equistore format, rather than exporting as numpy/torch and running your normal operations

docs/src/index.rst
```diff
@@ -1,9 +1,63 @@
-Equistore documentation
-=======================
+Equistore
```
rosecers, line 1: Same comments as in the equivalent text in the README and other docs.

python/src/equistore/data/extract.py
```diff
@@ -51,8 +49,11 @@ def register_external_data_wrapper(origin, klass):
     Register a non-Python data origin and the corresponding class wrapper.
 
     The wrapper class constructor must take two arguments (raw ``eqs_array`` and
-    python ``parent`` object) and return a subclass of either ``numpy.ndarray``
-    or ``torch.Tensor``, which keeps ``parent`` alive.
+    python ``parent`` object) and return a subclass of either
+    :py:class:`numpy.ndarray` or :py:class:`torch.Tensor`, which keeps
+    ``parent`` alive. The :py:class:`equistore.data.ExternalCpuArray` class
+    should provide the right behavior for data living in CPU memory, and can
+    serve as an example for more advanced custom arrays.
```
rosecers, line 56: Which German wrote this.

But in all seriousness, is the spelling as `klass` intentional? Why?

rosecers (commented):
Hi all,

This is a massive step up from where it was before. I made a few cosmetic changes, but my overarching suggestion would be to provide one example (simpler than SOAP) of something stored in `equistore` formats. Temperature-dependent data could be cool.

I've sent this to my peeps to see if they have comments. If I get any, I'll pass them along.

python/src/equistore/data/extract.py
Luthaf: Ich bin der Deutsche!

I used `klass` because `class` is a keyword and can not be used as a parameter name. Typical alternative are `cls`, `_class`, but I kinda like the german version

README.rst
```diff
@@ -4,14 +4,14 @@ Equistore
 |test| |docs|
 
 Equistore is a specialized data storage format suited to all your atomistic
-machine learning needs and more. Think of an NumPy "ndarray" or a pytorch "Tensor"
-carrying extra metadata for atomistic systems.
+machine learning needs and more. Think of an NumPy "ndarray" or a pytorch
+"Tensor" carrying extra metadata for atomistic systems.
```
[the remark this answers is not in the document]
Luthaf: It's not actually specific to materials, since we also do molecules & friends. How about "particle systems"? Or be more explicit and use `atomistic and other particles systems`?

docs/src/get-started/concepts.rst
Luthaf: Agree with that, but I would put the example in a separate page (with a link to it from this one)

docs/src/get-started/concepts.rst
Luthaf: > However, when a species is not present in a dataset, traditional numpy-like formats will be sparse, with implicit zeros wherever the species is not present.

that's not the case: numpy-like storage would have to store **explicit** zeros

docs/src/get-started/concepts.rst
Luthaf: Labels is not a superclass (i.e. parent class w.r.t. inheritance), it is the class/type which all metadata is an instance of

docs/src/get-started/goals.rst
Luthaf: It is kind of taking inspiration on https://en.wikipedia.org/wiki/Interchange_File_Format and friends. But I can find a better naming

docs/src/get-started/goals.rst
Luthaf: I tried to clarify this

README.rst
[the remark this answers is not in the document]
rosecers: That'd be great! Just making sure the verbiage is flexible enough to point to what we do 😂

docs/src/get-started/concepts.rst
rosecers: Agreed!

docs/src/get-started/goals.rst
rosecers: I got the reference, it just doesn't sound quite right grammatically, so unless it's a standard term, I would probably use "interchangeable" or something of the like

python/src/equistore/data/extract.py
rosecers: While I appreciate good deviant behavior, it might cause some confusion not to use the typical nomenclature. Up to you!

rosecers:
> > This is a massive step up from where it was before. I made a few cosmetic changes, but my overarching suggestion would be to provide one example (simpler than SOAP) of something stored in equistore formats. Temperature-dependent data could be cool.
> 
> I agree with Rose this already makes it much easier to start. We should definitely add a Python example. Do you have something specific in mind @rosecers. Maybe we can also recycle something from the docs branch.

I think something totally non-descriptor focused, like the RMSD of different atoms as we raise the temperature of a simulation, could be very cool. Any other thoughts? 

I say non-descriptor focused because (1) I think the format is useful beyond descriptor calculations and (2) someone should be able to understand the storage format without understand descriptor indexing.

PR #231, opened 2023-03-23
Do not create sub-classes of ndarray in zeros_like & friends

Luthaf:
Otherwise it keeps ExternalCpuArray around for too long, leaking memory. We found this out with @abmazitov.

Related to #53.


python/tests/operations/reduce_over_samples.py
```diff
@@ -265,7 +265,9 @@ def test_mean_samples_block(self):
         )
         bl1 = tensor_ps[0]
 
-        reduce_tensor_se = equistore.mean_over_samples(tensor_se, sample_names="center")
+        reduce_tensor_se = equistore.mean_over_samples(
+            tensor_se, sample_names=["center"]
```
DavideTisi, line 269: why you changed `"center"` -> `["center"]`. Is this not allowed anymore?

DavideTisi (approved):

DavideTisi (approved):

PR #232, opened 2023-03-23
Components should always contain at least one entry

Luthaf:
----
:books: Documentation preview :books:: https://equistore--232.org.readthedocs.build/en/232/

### DIFF
--- a/python/src/equistore/operations/equal_metadata.py
+++ b/python/src/equistore/operations/equal_metadata.py
@@ -240,6 +240,9 @@ def _check_blocks(a: TensorBlock, b: TensorBlock, props: List[str], fname: str):
                 if not (c1.names == c2.names):
                     raise ValueError(err_msg + err_msg_names)
 
+                if not (len(c1) == len(c2)):
+                    raise ValueError(err_msg + err_msg_1)
+
                 if not np.all(c1 == c2):
                     raise ValueError(err_msg + err_msg_1)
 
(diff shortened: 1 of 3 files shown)


python/src/equistore/operations/equal_metadata.py
```diff
@@ -240,6 +240,9 @@ def _check_blocks(a: TensorBlock, b: TensorBlock, props: List[str], fname: str):
                 if not (c1.names == c2.names):
                     raise ValueError(err_msg + err_msg_names)
 
+                if not (len(c1) == len(c2)):
+                    raise ValueError(err_msg + err_msg_1)
+
```
jwa7, line 245: I'm not sure I see the point of this check. Doesn't the check on line 246 do the same thing?
```py
if not np.all(c1 == c2):
    raise ValueError(err_msg + err_msg_1)
```

jwa7 (commented):
Please could you briefly explain what the aim of these changes are - not sure I quite see it? (Perhaps better in person)

python/src/equistore/operations/equal_metadata.py
Luthaf: The line below does check the same thing, but creates a warning about elementwise comparison in numpy.

Luthaf:
This is mainly about consistency and restricting the number of possible states a TensorBlock can be in. I see no use-case for having 0 components (if you think of one, I'm happy to revisit!). This is different from samples and properties where having 0 of them can be useful/happen through slicing/...

jwa7:
Discussion in person: understood and approved

PR #233, opened 2023-03-24
Add torch bindings

Luthaf:
This only contains a very basic scaffolding of the package for now. The goal it to make multiple small PR targetting the 'torch' branch, and only merge this branch/PR once everything is ready.

This draft PR is here to start running CI on the code.

----
:books: Documentation preview :books:: https://equistore--233.org.readthedocs.build/en/233/

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,6 +11,7 @@ envlist =
 
 
 [testenv]
+passenv = *
 package = external
 package_env = build-equistore-core
 lint_folders = "{toxinidir}/python" "{toxinidir}/setup.py"
@@ -67,7 +68,6 @@ commands =
 
 [testenv:torch-tests]
 # this environement runs the tests of the equistore-torch Python package
-passenv = *
 deps =
     pytest
     numpy
@@ -163,7 +163,6 @@ commands =
 [testenv:build-tests]
 # this environement makes sure one can build sdist and wheels for Python
 package = skip
-passenv = *
 deps =
     build
     twine  # a tool to check sdist and wheels metadata
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,15 +5,21 @@
 # Required
 version: 2
 
-# Set the version of Python and other tools we need
 build:
   os: ubuntu-22.04
   apt_packages:
     - cmake
   tools:
+    # Set the version of Python and other tools we need
     python: "3.10"
     rust: "1.61"
   jobs:
+    post_install:
+      # install equistore-torch with the CPU version of PyTorch we can not use
+      # the `python` section below since it does not allow to specify
+      # `--extra-index-url`
+      - pip install --extra-index-url https://download.pytorch.org/whl/cpu python/equistore-torch
+
     post_build:
       - cargo doc --no-deps --package equistore
       - rm -rf $READTHEDOCS_OUTPUT/html/reference/rust/
@@ -23,11 +29,11 @@ build:
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
-   configuration: docs/src/conf.py
+  configuration: docs/src/conf.py
 
 # Declare the Python requirements required to build the docs
 python:
-   install:
-   - requirements: docs/requirements.txt
-   - method: pip
-     path: .
+  install:
+  - requirements: docs/requirements.txt
+  - method: pip
+    path: .
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -7,6 +7,10 @@ import toml
 
 ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", ".."))
 
+# when importing equistore-torch, this will change the definition of the classes
+# to include the documentation
+os.environ["EQUISTORE_IMPORT_FOR_SPHINX"] = "1"
+
 # -- Project information -----------------------------------------------------
 
 project = "equistore"
--- a/docs/src/get-started/installation.rst
+++ b/docs/src/get-started/installation.rst
@@ -91,6 +91,7 @@ Add the following to your project ``Cargo.toml``
     [dependencies]
     equistore = {git = "https://github.com/lab-cosmo/equistore.git"}
 
+.. _install-torch-script-lib:
 
 Installing the TorchScript bindings
 -----------------------------------
--- a/docs/src/reference/cxx/index.rst
+++ b/docs/src/reference/cxx/index.rst
@@ -21,8 +21,9 @@ installation):
     # Link to equistore, this makes the header accessible
     target_link_libraries(MyExecutable equistore)
 
-    # alternatively, use the static build of equistore
+    # alternatively, you can explicitly use the static or shared build of equistore
     # target_link_libraries(MyExecutable equistore::static)
+    # target_link_libraries(MyExecutable equistore::shared)
 
 The functions and types provided in ``equistore.hpp`` can be grouped as follow:
 
--- a/docs/src/reference/python/index.rst
+++ b/docs/src/reference/python/index.rst
@@ -16,10 +16,10 @@ provided in ``equistore`` can be grouped as follows:
 - advanced functionalities like the :ref:`array format <python-api-array>` for
   data storage
 
-The Python interface also provides multiple :ref:`operations
-<python-api-operations>` including mathematical, logical, and other utility
-functions that can applied on these core objects. These are documented in the
-next section.
+The Python interface also provides multiple operations, including mathematical,
+logical, and other utility functions that can applied on these core objects.
+These are documented in the next :ref:`documentation section
+<python-api-operations>`.
 
 .. toctree::
     :maxdepth: 1
--- a/docs/src/reference/torch/cxx/labels.rst
+++ b/docs/src/reference/torch/cxx/labels.rst
@@ -5,3 +5,8 @@ Labels
 
 .. doxygenclass:: equistore_torch::LabelsHolder
     :members:
+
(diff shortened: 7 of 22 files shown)


agoscinski (commented):
You create for all classes a holder class with an inheritance to torch::CustomClassHolder so one can register them as torch classes. Do you know what torch::CustomClassHolder does in the background? From here it seems just like boilerplate code without any use. 

 Is the change of namespace when loading the equistore-torch package already in this PR?

Luthaf:
My understanding is that `torch::CustomClassHolder` has fields to allow using `torch::intrusive_ptr` (a `std::shared_ptr` which stores the reference count inside the object).

> Is the change of namespace when loading the equistore-torch package already in this PR?

You mean the thing we discussed about reloading equistore-operations in a different context? No, not yet. But 
`equistore-torch` is already a namespace package, which can be imported with `from equistore.torch import TensorMap`

equistore-torch/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "equistore-torch"
+version = "0.0.0"
+edition = "2021"
```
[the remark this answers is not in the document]
Luthaf: this is the latest edition of Rust: https://doc.rust-lang.org/edition-guide/rust-2021/index.html

equistore-torch/src/array.cpp
```diff
@@ -0,0 +1,51 @@
...
+std::unique_ptr<equistore::DataArrayBase> TorchDataArray::create(std::vector<uintptr_t> shape) const {
+    throw std::runtime_error("not implemented");
+}
+
+double* TorchDataArray::data() {
+    throw std::runtime_error("not implemented");
+}
+
+const std::vector<uintptr_t>& TorchDataArray::shape() const {
+    throw std::runtime_error("not implemented");
```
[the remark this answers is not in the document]
Luthaf: yes! I have another branch filling this up, but this one is just about setting up the build system & tests (and it is already quite painful, as you can see by the number of force-push I did ^_^)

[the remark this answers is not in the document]
jwa7: Haha I see, thanks!

equistore-torch/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
...
+    endif()
+endif()
+
+find_package(equistore 0.1 CONFIG REQUIRED)
+
+# FindCUDNN.cmake distributed with PyTorch is a bit broken, so we have a
+# fixed version in `cmake/FindCUDNN.cmake`
+set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake;${CMAKE_MODULE_PATH}")
+
+find_package(Torch 1.13 REQUIRED)
```
agoscinski, line 36: Is there a reason for the version number one can mention here?

agoscinski (commented):
went through one quarter

equistore-torch/CMakeLists.txt
Luthaf: This is the minimal version of torch we support

equistore-core/tests/utils/mod.rs
```diff
@@ -4,13 +4,24 @@
 use std::path::Path;
 use std::process::Command;
 
-pub fn cmake_config(source_dir: &Path, build_dir: &Path, build_type: &str) -> Command {
+fn build_type() -> &'static str {
+    // assume that debug assertion means that we are building the code in
+    // debug mode, even if that could be not true in some cases
```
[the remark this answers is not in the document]
Luthaf: no, the comment means the reverse of what you wrote =). We would like debug mode to mean "not optimization", but I don't think we can detect that. So instead we check for "does the code also run "debug" checks/assertions" (i.e. `debug_assert!` in the code), which could be also turned on for some release/with optimization builds

equistore-core/tests/cpp/external/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ function(unpack_library _name_)
 endfunction()
 
 
-
 add_library(catch STATIC catch/catch.cpp)
 target_include_directories(catch PUBLIC catch)
+
+target_compile_features(catch PUBLIC cxx_std_14)
```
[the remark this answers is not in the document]
Luthaf: This might actually no longer be required. Let me check

.github/workflows/tests.yml
```diff
@@ -56,8 +57,9 @@ jobs:
           apt install -y software-properties-common
           add-apt-repository -y ppa:git-core/ppa
           add-apt-repository ppa:deadsnakes/ppa
-          apt install -y cmake make gcc g++ git curl rustc cargo
-          apt install -y python3.7 python3-pip
+          apt install -y cmake make gcc g++ git curl wget rustc cargo sudo
```
PicoCentauri, line 60: Where do we need `sudo` and isn't `curl` instead of `wget` available?

python/equistore-torch/pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+python_files = *.py
+testpaths = tests
```
PicoCentauri, line 3: I think we have this in the pyproject.toml in the other subpackges.

tox.ini
```diff
@@ -97,7 +116,6 @@ commands =
     blackdoc --check --diff {[testenv]lint_folders}
     isort --check-only --diff {[testenv]lint_folders}
 
-
```
PicoCentauri, line 100: I have the feeling every time we touch the `tox.ini` file we either add or delete lines between the sections 🤣

PicoCentauri (changes requested):
Looks similar to the other sub-packages and good!

equistore-core/tests/utils/mod.rs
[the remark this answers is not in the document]
agoscinski: I thought you mean with that comment that sometimes you can run debug mode with some kind of optimization, so the `debug_assert` is not correctly detecting the debug mode. But since this is not the case, then I don't understand the part `even if that could be not true in some cases` in the comment. The `debug_assert` does exactly mean not optimized.

equistore-torch/CMakeLists.txt
agoscinski: yes, but why this torch version. If there is a reason it would be nice to write it here.

equistore-core/tests/utils/mod.rs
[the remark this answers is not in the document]
Luthaf: no, `debug_assert` just mean "add some more internal consistency checks". Optimizations (inlining, loop unrolling, constant propagations, and other things LLVM does) are controlled by a separate `opt-level` setting, which I don't think we can query from inside the code.

.github/workflows/tests.yml
Luthaf: these are leftover when I was figuring out how to get libtorch installed on CI for the C++ tests (now we use the pip version).

I'll remove this!

equistore-torch/CMakeLists.txt
Luthaf: I did not try other versions, I'll give this a go. I'll also add a tests checking the build with whatever version we end up saying we support

Luthaf: This is now updated to 1.11 as the minimal version, and we run tests on CI to check against it.

.github/workflows/torch-tests.yml
```diff
@@ -0,0 +1,59 @@
+name: TorchScript tests
```
agoscinski, line 1: We will also need in the future a workflow compiled with CUDA, right?

Luthaf: Only if we can get a CI maching able to run CUDA code. I wanted to experiment with self-hosted github action runner, and use the imx server for this, it should be doable (we only need a demon on the head node that stays conected to github and answers requests).

equistore-core/tests/utils/mod.rs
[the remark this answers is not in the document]
agoscinski: I don't fully get it

> debug_assert! statements are only enabled in non optimized builds by default. An optimized build will not execute debug_assert! statements unless -C debug-assertions is passed to the compiler

https://doc.rust-lang.org/std/macro.debug_assert.html

What kind of optimization it can then detect?

tox.ini
```diff
@@ -63,6 +64,24 @@ commands =
     pytest --import-mode=append {posargs}
 
 
+[testenv:torch-tests]
+# this environement runs the tests of the equistore-torch Python package
+passenv = *
+deps =
+    pytest
+    numpy
+    torch
```
agoscinski, line 73: How does this tox env get the torch cpu wheels? I don't see anywhere here an index url for the cpu specified.

tox.ini
```diff
@@ -29,7 +30,7 @@ deps =
     cmake
 
 commands =
-    pip wheel python/equistore-core --no-build-isolation --no-deps --check-build-dependencies --wheel-dir {envtmpdir}/dist
+    pip wheel python/equistore-core --no-deps --no-build-isolation --check-build-dependencies --wheel-dir {envtmpdir}/dist
```
agoscinski, line 33: is the order important?

Luthaf: not really, I just changed this back and forth a lot of time ^^. I'll check to use the same order for all commands in this file

Luthaf: I've added a tox variable to hold all of these flags

PR #235, opened 2023-03-24
More use of pytest functionalities

PicoCentauri:
- User `tensor_map` and `large_tensor_map` as pytest.fixture: -> simplifies code
- port abs to pytest
- Remove classes for tests

----
:books: Documentation preview :books:: https://equistore--235.org.readthedocs.build/en/235/

### DIFF
--- a/python/tests/__init__.py
+++ b/python/tests/__init__.py
@@ -1,0 +1,2 @@
+# This file is required to allow relative imports of the functions in this directory
+# from child directories i.e. `from .. import utils`.
--- a/python/tests/blocks.py
+++ b/python/tests/blocks.py
@@ -9,225 +9,234 @@ import equistore.status
 from equistore import Labels, TensorBlock
 
 
-class TestBlocks:
-    @pytest.fixture
-    def block(self):
-        b = TensorBlock(
-            values=np.full((3, 2), -1.0),
-            samples=Labels(["samples"], np.array([[0], [2], [4]])),
+@pytest.fixture
+def block():
+    b = TensorBlock(
+        values=np.full((3, 2), -1.0),
+        samples=Labels(["samples"], np.array([[0], [2], [4]])),
+        components=[],
+        properties=Labels(["properties"], np.array([[5], [3]])),
+    )
+
+    return b
+
+
+def test_gradient_no_sample_error(block):
+    msg = (
+        "invalid parameter: gradients samples must have at least "
+        "one dimension named 'sample', we got none"
+    )
+    with pytest.raises(equistore.status.EquistoreError, match=msg):
+        block.add_gradient(
+            "parameter",
+            data=np.zeros((0, 2)),
+            samples=Labels([], np.empty((0, 2))),
             components=[],
-            properties=Labels(["properties"], np.array([[5], [3]])),
         )
 
-        return b
 
-    def test_gradient_no_sample_error(self, block):
-        msg = (
-            "invalid parameter: gradients samples must have at least "
-            "one dimension named 'sample', we got none"
-        )
-        with pytest.raises(equistore.status.EquistoreError, match=msg):
-            block.add_gradient(
-                "parameter",
-                data=np.zeros((0, 2)),
-                samples=Labels([], np.empty((0, 2))),
-                components=[],
-            )
-
-    def test_repr(self, block):
-        expected = """TensorBlock
+def test_repr(block):
+    expected = """TensorBlock
     samples (3): ['samples']
     components (): []
     properties (2): ['properties']
     gradients: no"""
-        assert block.__repr__() == expected
+    assert block.__repr__() == expected
 
-    def test_repr_zero_samples(self):
-        block = TensorBlock(
-            values=np.zeros((0, 2)),
-            samples=Labels([], np.empty((0, 2))),
-            components=[],
-            properties=Labels(["properties"], np.array([[5], [3]])),
-        )
-        expected = """TensorBlock
+
+def test_repr_zero_samples():
+    block = TensorBlock(
+        values=np.zeros((0, 2)),
+        samples=Labels([], np.empty((0, 2))),
+        components=[],
+        properties=Labels(["properties"], np.array([[5], [3]])),
+    )
+    expected = """TensorBlock
     samples (0): []
     components (): []
     properties (2): ['properties']
     gradients: no"""
-        assert block.__repr__() == expected
+    assert block.__repr__() == expected
 
-    def test_repr_zero_samples_gradient(self, block):
-        block.add_gradient(
-            "parameter",
-            data=np.zeros((0, 2)),
-            samples=Labels(["sample"], np.empty((0, 2))),
-            components=[],
-        )
 
-        expected_block = """TensorBlock
+def test_repr_zero_samples_gradient(block):
+    block.add_gradient(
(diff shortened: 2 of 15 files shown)


python/tests/tensor.py
```diff
@@ -1,45 +1,37 @@
...
-    @pytest.fixture
-    def large_tensor(self):
-        return large_tensor_map()
-
-    def test_copy(self, tensor):
-        copy = tensor.copy()
-        block_1_values_id = id(tensor.block(0).values)
-
-        del tensor
+        del tensor_map
```
Luthaf, line 14: this might not be doing what we want with `tensor_map` as a fixture. The goal is to remove all references to tensor_map, but pytest will still be holding one here.

python/tests/tensor.py
```diff
@@ -285,10 +277,10 @@ def test_keys_to_samples(self, tensor):
         )
         assert_equal(gradient.data, expected)
 
-    def test_keys_to_samples_unsorted(self, tensor):
-        tensor = tensor.keys_to_samples("key_2", sort_samples=False)
+    def test_keys_to_samples_unsorted(self, tensor_map):
```
Luthaf, line 280: should we name the fixture `tensor` instead? Makes for simpler code everywhere

PicoCentauri: yes, we can.

python/tests/tensor.py
PicoCentauri: ahh you might be right. Can we test this somehow if all references are gone. Maybe pytest is clever :-)

PicoCentauri: According to the pytest docs a fixture has a scope which by default is function. So every function should get its own independent copy.

https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-fixture

So it should be fine.

python/tests/operations/abs.py
```diff
@@ -1,220 +1,69 @@
...
-            ],
-        )
-        keys = Labels(
-            names=["key_1", "key_2"], values=np.array([[0, 0], [1, 0]], dtype=np.int32)
-        )
-        A = TensorMap(keys, [block_1, block_2])
-        tensor_abs = equistore.abs(A)
-        tensor_result = TensorMap(keys, [block_res1, block_res2])
+    @pytest.fixture
+    def tensor_map_abs(self, tensor):
```
Luthaf, line 12: This is a bit puzzling to me. Where do `tensor` comes from? We never explicitly imported it in scope.

python/tests/tensor.py
PicoCentauri: The `tensors` is not doing copies right? It is creating a new `TensorMap` from scratch. `pytest` will give an independent instance to each test.

PicoCentauri:
I see your point though that this appears a bit magical and if you think it is not enough to explain this somewhere we can also use your approach.

PicoCentauri:
I had to do a lot of rebasing. Maybe @frostedoyster, @jwa7 and @Luthaf can you check if I did this correctly. I think it should be mainly be the changes of #226, #231 and #239.

Sorry for the extra work!

frostedoyster:
Things look good to me

python/tests/blocks.py
```diff
@@ -9,225 +9,234 @@
...
+@pytest.fixture
+def block():
+    b = TensorBlock(
+        values=np.full((3, 2), -1.0),
+        samples=Labels(["samples"], np.array([[0], [2], [4]])),
+        components=[],
+        properties=Labels(["properties"], np.array([[5], [3]])),
+    )
+
+    return b
```
Luthaf, line 21:
```suggestion
    return TensorBlock(
        values=np.full((3, 2), -1.0),
        samples=Labels(["samples"], np.array([[0], [2], [4]])),
        components=[],
        properties=Labels(["properties"], np.array([[5], [3]])),
    )
```

python/tests/operations/abs.py
```diff
@@ -1,210 +1,78 @@
...
+from equistore import TensorBlock, TensorMap
 
+from .. import utils
 
-DATA_ROOT = os.path.join(os.path.dirname(__file__), "..", "data")
-TEST_FILE = "qm7-spherical-expansion.npz"
 
+@pytest.fixture
+def tensor():
+    """A pytest fixture for the test function :func:`utils.tensor`."""
```
Luthaf, line 14:
```suggestion
```

I find this more confusing than helpful

python/tests/operations/abs.py
```diff
@@ -1,210 +1,78 @@
...
-            components=[],
-            properties=Labels.arange("properties", 2),
-        )
-        block_2 = TensorBlock(
-            values=np.array([[-1, -2], [-3, 4], [5, -6]]),
-            samples=Labels(["samples"], np.array([[0], [2], [7]])),
-            components=[],
-            properties=Labels.arange("properties", 2),
+@pytest.fixture
+def tensor_map_abs(tensor):
```
Luthaf, line 19:
```suggestion
def tensor_map_complex(tensor):
```

python/tests/tensor.py
```diff
@@ -3,70 +3,78 @@
...
-    @pytest.fixture
-    def tensor(self):
-        return tensor_map()
 
-    @pytest.fixture
-    def large_tensor(self):
-        return large_tensor_map()
+@pytest.fixture
+def tensor():
+    """A pytest fixture for the test function :func:`utils.tensor`."""
```
Luthaf, line 14:
```suggestion
```

python/tests/tensor.py
Luthaf: I'll add a commit checking for this

Luthaf: yep, pytest was keeping more references (6 total), I changed this test not to use fixture

python/tests/__init__.py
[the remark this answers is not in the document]
PicoCentauri: To allow import from child directories. I will put a comment in the file.

python/tests/operations/abs.py
PicoCentauri: Fine with removing.

PicoCentauri:
Thanks for the commit and the comments @Luthaf. I put a comment line in the test for the tensor copy to prevent us from changing this back to fixture once we forgot about this discussion.

Luthaf (approved):

python/tests/operations/abs.py
```diff
@@ -1,210 +1,77 @@
...
-        keys = Labels(names=["key_1", "key_2"], values=np.array([[0, 0], [1, 0]]))
-        A = TensorMap(keys, [block_1, block_2])
-        tensor_abs = equistore.abs(A)
-        tensor_result = TensorMap(keys, [block_res1, block_res2])
 
-        assert equistore.equal(tensor_result, tensor_abs)
+        for parameter, gradient in block.gradients():
+            new_block.add_gradient(
+                parameter=parameter,
+                data=-gradient.data,
```
DivyaSuman14, line 56: Right now since all the values of block.values are negative having -gradient.data works fine, but maybe we can do another test for the general case.

DivyaSuman14: Okay, I'll do that.

PicoCentauri: Thanks @DivyaSuman14! We can also take a look together and write the test.

PR #238, opened 2023-03-29
Pickling support by converting to numpy

agoscinski:
First draft of pickling by converting everything to numpy arrays. Creates additional copies for conversion to numpy array, but when numpy arrays and pickle protocol 5 is used, no spurious copies should be made for the values. Need to check this. Also current test only works, because it only handles numpy arrays. I am sure with torch tensor it will fail (gradients are not converted)

----
:books: Documentation preview :books:: https://equistore--238.org.readthedocs.build/en/238/

### DIFF
--- a/python/src/equistore/tensor.py
+++ b/python/src/equistore/tensor.py
@@ -1,6 +1,14 @@
 import copy
 import ctypes
+import os
+import pickle
+import sys
+import tempfile
 from typing import List, Union
+
+
+if (sys.version_info.major >= 3) and (sys.version_info.minor >= 8):
+    from pickle import PickleBuffer
 
 import numpy as np
 
@@ -68,6 +76,22 @@ class TensorMap:
         obj._blocks = []
         return obj
 
+    @classmethod
+    def _from_pickle(cls, data: bytes):
+        """
+        Passed to pickler to reconstruct TensorMap from bytes object
+        """
+        import equistore
+
+        path = os.path.join(tempfile.gettempdir(), str(id(data)))
+        try:
+            with open(path, "wb") as f:
+                pickle.dump(data, f)
+            tensor_map = equistore.load(path)
+        finally:
+            os.remove(path)
+        return tensor_map
+
     def __del__(self):
         if hasattr(self, "_lib") and self._lib is not None and hasattr(self, "_ptr"):
             self._lib.eqs_tensormap_free(self._ptr)
@@ -87,6 +111,26 @@ class TensorMap:
         non Python-owned) data and metadata
         """
         return copy.deepcopy(self)
+
+    def __reduce_ex__(self, protocol: int):
+        """
+        Used by the Pickler to dump TensorMap object to bytes object.
+        When protocol >= 5 it supports PickleBuffer which reduces number of
+        copyies needed
+        """
+        import equistore
+
+        path = os.path.join(tempfile.gettempdir(), str(id(self)) + ".npz")
+        try:
+            equistore.save(path, self)
+            with open(path, "rb") as f:
+                data = f.read()
+        finally:
+            os.remove(path)
+        if protocol >= 5:
+            return self._from_pickle, (PickleBuffer(data),), None
+        else:
+            return self._from_pickle, (data,)
 
     def __iter__(self):
         keys = self.keys
--- a/python/src/equistore/block.py
+++ b/python/src/equistore/block.py
@@ -110,6 +110,12 @@ class TensorBlock:
         new_ptr = self._lib.eqs_block_copy(self._ptr)
         return TensorBlock._from_ptr(new_ptr, parent=None)
 
+    def __reduce__(self):
+        raise NotImplementedError(
+            "Pickling for TensorBlocks does not work. Please wrap it into a TensorMap"
+            " to save it."
+        )
+
     def copy(self) -> "TensorBlock":
         """
         Get a deep copy of this block, including all the (potentially non
--- a/python/src/equistore/io.py
+++ b/python/src/equistore/io.py
@@ -120,7 +120,7 @@ def save(path: str, tensor: TensorMap, use_numpy=False):
     if not path.endswith(".npz"):
         path += ".npz"
         warnings.warn(
-            msg=f"adding '.npz' extension, the file will be saved at '{path}'",
+            message=f"adding '.npz' extension, the file will be saved at '{path}'",
             stacklevel=1,
         )
 
--- a/python/tests/serialization.py
+++ b/python/tests/serialization.py
@@ -12,6 +12,10 @@ from equistore import TensorMap
 from . import utils
 
 
+# using tmpdir as pytest-built-in fixture
+# https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures
+
+
 @pytest.fixture
 def tensor():
     return utils.tensor()
@@ -68,6 +72,20 @@ def test_save(use_numpy, tmpdir, tensor):
             assert_equal(data[f"{prefix}/data"], gradient.data)
             assert_equal(data[f"{prefix}/samples"], gradient.samples)
             assert_equal(data[f"{prefix}/components/0"], gradient.components[0])
+
+
+def test_save_warning(tmpdir, tensor):
+    # does not have .npz ending and causes warning
+    tmpfile = "serialize-test"
+
+    with pytest.warns() as record:
+        with tmpdir.as_cwd():
+            equistore.save(tmpfile, tensor)
+
+    assert (
+        str(record[0].message)
+        == f"adding '.npz' extension, the file will be saved at '{tmpfile}.npz'"
+    )
 
 
 if (sys.version_info.major >= 3) and (sys.version_info.minor >= 8):
--- a/python/tests/serialization.py
+++ b/python/tests/serialization.py
@@ -1,4 +1,6 @@
 import os
+import pickle
+import sys
 
 import numpy as np
 import pytest
@@ -66,3 +68,43 @@ def test_save(use_numpy, tmpdir, tensor):
             assert_equal(data[f"{prefix}/data"], gradient.data)
             assert_equal(data[f"{prefix}/samples"], gradient.samples)
             assert_equal(data[f"{prefix}/components/0"], gradient.components[0])
+
+
+if (sys.version_info.major >= 3) and (sys.version_info.minor >= 8):
+    protocols = (4, 5)
+else:
+    protocols = (4,)
+
+
+@pytest.mark.parametrize("protocol", protocols)
+def test_pickle(protocol, tmpdir, tensor):
+    """
+    Checks that pickling and unpickling a tensor map
+    results in the same tensor map
+    """
+
+    tmpfile = "serialize-test.pickle"
+
+    with tmpdir.as_cwd():
+        with open(tmpfile, "wb") as f:
+            pickle.dump(tensor, f, protocol=protocol)
+
+        with open(tmpfile, "rb") as f:
+            tensor_loaded = pickle.load(f)
+
+    assert_equal(tensor.keys, tensor_loaded.keys)
+    assert_equal(len(tensor.blocks()), len(tensor_loaded.blocks()))
+    for i, (_, block) in enumerate(tensor):
+        ref_block = tensor.blocks()[i]
+        assert_equal(type(block.values), type(ref_block.values))
+        assert_equal(block.values, ref_block.values)
+        assert_equal(block.samples, ref_block.samples)
+        assert_equal(block.components, ref_block.components)
+        assert_equal(block.properties, ref_block.properties)
+
+        for parameter in block.gradients_list():
+            gradient = block.gradient(parameter)
+            ref_gradient = ref_block.gradient(parameter)
+            assert_equal(gradient.data, ref_gradient.data)
+            assert_equal(gradient.samples, ref_gradient.samples)
+            assert_equal(gradient.components, ref_gradient.components)


Luthaf:
I'm not fond of having a different format than the one we already use for `equistore.save`/`equistore.load`, especially since that means maintaining the two of them when modifying the data structure (such as in #227). 

Why did you go this route instead of using `equistore.save`/`equistore.load` and making version of them adapted for pickle?

Luthaf:
> As far as I understand for save and load we actually don't use npz format (use_numpy=False branch)

No, we do! Both branches create a NPZ file, which is just a ZIP file containing NPY data. The rust branch re-implements a writer/parser for NPY files.

> not sure what crate::io::save exactly does, could not find doc page

this is the one https://lab-cosmo.github.io/equistore/latest/reference/rust/equistore/io/fn.save.html, `crate` expands to the current crate (i.e. package in Python terms), here equistore.

>  But for pickling we have to provide the byteobject (best only a view) to the Pickler. 

Yes, this is what we discussed last time. In my mind, the simplest way to do this would be to add an `eqs_tensormap_save_buffer`/`eqs_tensormap_load_buffer` function to the C API, looking like this:

```c
eqs_status_t eqs_tensormap_save_buffer(eqs_tensormap_t* tensor, uint8_t** buffer, size_t* length);
eqs_status_t eqs_tensormap_load_buffer(const uint8_t* buffer, size_t length, create_array_callback);
```

and then the returned buffer can be passed to the Pickler directly!

The `load` function should be pretty easy to write, a simple replacement of https://github.com/lab-cosmo/equistore/blob/6c484592f3532bba97b67b8f88e71b3e47f7939e/equistore-core/src/c_api/io.rs#L105-L107

with something like 

```rust
buffer = std::slice::from_raw_parts(buffer, length);
let tensor = crate::io::load(buffer, create_array)?; 
```

The `save` function is  bit harder, because we need to allocate for the `buffer`, but it is not clear who should free the data. We could either allocate with `malloc` and make python free the data when it is done with `free`, or pass an explicit allocator to the function. I'm happy to help to find a permanent solution here, but a temporary one which would be backward compatible is to do this

```py
def __setstate__(self, buffer):
    with open("tmp.npz", "wb") as fd:
       fd.write(buffer)

    # not sure if one can reset self? Otherwise we need to copy the fields
    self = equistore.io.load("tmp.npz")

def __getstate__(self):
    equistore.io.save("tmp.npz")
    with open("tmp.npz", "rb") as fd:
       buffer = fd.read()

    return buffer
```

agoscinski:
First thanks for the long comment, it makes a lot of things clearer to me. I think the naming confused a lot: python's binary buffer and a byte buffer are very different (maybe first below what I mean with this). There are a

> and then the returned buffer can be passed to the Pickler directly!

This part is not very clear to me. I cannot really create a buffer on the Python side, because I don't know how large the object is.  Okay I can transfer this information to the Python side and create a buffer.
and pass it to Rust where it writes to the buffer.
```python
size = lib.eqs_tensormap_size_as_buffer()
buffer = bytes(b"0"*size)
lib.eqs_tensormap_save_buffer( (ctypes.c_byte * size).from_buffer(buffer), size)
```
Or the Rust function could return a buffer (ptr + size) and on the Python side this can be wrapped directly with the `bytes` function and passed to the pickler
https://discuss.python.org/t/ctypes-how-to-convert-a-c-ubyte-array-to-python-bytes/13994/2
But both cases will result in copies of a byte object.


#### Preventing the byte buffer copy

What I understood from public resources, to avoid any copy one has to support the python buffer protocol (it's an interface to me, don't know why the naming), so python knows how to read from the rust object without copying from it (e.g. making a memoryview).
https://docs.python.org/3/c-api/buffer.html
https://github.com/python/cpython/blob/main/Include/pybuffer.h

Since this requires CPython, this would require another layer of interface we need to implement I think (C-API <-> CPython <-> Python).
http://jakevdp.github.io/blog/2014/05/05/introduction-to-the-python-buffer-protocol/#Adding-the-Buffer-Protocol

There is a crate pyo3 that offers CPython bindings for Rust, but this seems also like a lot of work to restructure the classes in the Rust code which we wrap (is this the right word?) Python.
https://docs.rs/pyo3/0.11.1/x86_64-apple-darwin/pyo3/class/buffer/index.html
Issue talking about the python buffer protocol https://github.com/PyO3/pyo3/issues/1385


I guess performance-wise that one copy does not matter much, so I would start with your suggested temporary solution for this PR (just dumping to a file from Rust side and reading it from python side) and then start a new PR with your suggestion to pass a byte buffer. What do you think

Luthaf:
Another possibility to avoid a copy of the buffer seems to be the [memoryview](https://docs.python.org/3/c-api/memoryview.html) type. In particular, [PyMemoryView_FromMemory](https://docs.python.org/3/c-api/memoryview.html#c.PyMemoryView_FromMemory) might be doing what we need, and it seems like we can call it dynamically from ctypes (instead of linking to the CPython C-API): https://stackoverflow.com/a/72968176/4692076.

We still need to be able to pass pointer + length to Python, and Python needs to be able to free the pointer, so we could have a small wrapper class like this

```py
class MallocBuffer:
    def __init__(self, ptr, size):
        self.ptr = ptr
        self.size = size
    
    def view(self):
        assert self.ptr is not None
        return memoryview(...)  # use PyMemoryView_FromMemory
    
    def __del__(self):
        # free the pointer when we are done with it
        libc.free(self.ptr)
        self.ptr = None
```

The C API can then be a single function, returning pointer + size; where the pointer is allocated with `malloc`/`realloc`. I remember it was a bit problematic to access `free` from ctypes though, so we would want to check this.

Regarding the plan, I would go with something like 

- step 0: write to a temporary files & read it. This allow us to use pickle for other things in a backward compatible way
- step 1: write the `load` function, which should be the easiest
- step 2: write the save function

> What I understood from public resources, to avoid any copy one has to support the python buffer protocol (it's an interface to me, don't know why the naming)

Python protocols are the same as interfaces in other languages.

agoscinski:
Before the rebase it worked, wtih the new master For the script
```python
import pickle
import equistore.io

from utils import tensor_map

tm = tensor_map()

equistore.io.save('tmp.npz', tm)

with open('tmp.npz', 'rb') as f: b = f.read()

with open('tmp.pikcle', 'wb') as f: pickle.dump(b, f)

equistore.io.load('tmp.pikcle')

```
I get
```
  File "/home/alexgo/code/c-api-pickle/equistoresave.py", line 14, in <module>
    equistore.io.load('tmp.pikcle')
  File "/opt/anaconda3/lib/python3.10/site-packages/equistore/io.py", line 34, in load
    return load_custom_array(path, create_numpy_array)
  File "/opt/anaconda3/lib/python3.10/site-packages/equistore/io.py", line 69, in load_custom_array
    return TensorMap._from_ptr(ptr)
  File "/opt/anaconda3/lib/python3.10/site-packages/equistore/tensor.py", line 68, in _from_ptr
    _check_pointer(ptr)
  File "/opt/anaconda3/lib/python3.10/site-packages/equistore/status.py", line 48, in _check_pointer
    raise EquistoreError(last_error())
equistore.status.EquistoreError: serialization format error: invalid header in file: expected a string, got ' '
```

Luthaf:
I don't expect this to work, the pickle format adds some more data around what we give it. Something like this should work though

```py
import pickle
import equistore.io

from utils import tensor_map

tm = tensor_map()

equistore.io.save('tmp.npz', tm)

with open('tmp.npz', 'rb') as f: b = f.read()

with open('tmp.pikcle', 'wb') as f: pickle.dump(b, f)

with open('tmp.pikcle', 'rb') as f: c = pickle.load(f)

with open('tmp-2.npz', 'wb') as f: f.write(c)

equistore.io.load('tmp-2.npz')
```

agoscinski:
> I don't expect this to work, the pickle format adds some more data around what we give it. Something like this should work though

I agree, but it worked before the rebase, so I assumed pickle does not add anything if it gets one binary object. I also tried what you have sent, and this gave me the same error message. So I suspect it the reason lies somewhere else.

Luthaf:
Yes, looks like I forgot to strip a string somewhere in #240. I'll update the code, sorry about that!

python/src/equistore/tensor.py
```diff
@@ -1,12 +1,22 @@
 import copy
 import ctypes
+import os
+import pickle
+import sys
+import tempfile
+
+
+if (sys.version_info.major >= 3) and (sys.version_info.minor >= 7):
+    from pickle import PickleBuffer
```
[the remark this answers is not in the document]
agoscinski: Did now a if-then case. Do you rather want the try-except block?

python/src/equistore/tensor.py
```diff
@@ -1,12 +1,22 @@
 import copy
 import ctypes
+import os
+import pickle
+import sys
+import tempfile
+
+
+if (sys.version_info.major >= 3) and (sys.version_info.minor >= 8):
+    from pickle import PickleBuffer
```
Luthaf, line 10: Could you move this import below the `from typing ...`. This keeps import grouped in stdlib/external/internal

python/src/equistore/tensor.py
```diff
@@ -68,6 +78,23 @@ def _from_ptr(ptr):
         obj._blocks = []
         return obj
 
+    @classmethod
+    def _from_pickle(cls, tm_bytes: bytes):
```
Luthaf, line 82: I had to think a bit on the meaning of `tm_` here, I though it was tmp initiallly

```suggestion
    def _from_pickle(cls, data: bytes):
```

python/src/equistore/tensor.py
```diff
@@ -68,6 +78,23 @@ def _from_ptr(ptr):
...
+        Passed to pickler to reconstruct TensorMap from bytes object
+        """
+        lib = _get_library()
+        path = os.path.join(tempfile.gettempdir(), str(id(tm_bytes)))
+        try:
+            with open(path, "wb") as f:
+                pickle.dump(tm_bytes, f)
+            ptr = lib.eqs_tensormap_load(
+                path.encode("utf8"), eqs_create_array_callback_t(create_numpy_array)
+            )
```
Luthaf, line 93: Why not `tensor = equistiore.io.load(...)` here?

python/src/equistore/tensor.py
```diff
@@ -88,6 +115,24 @@ def copy(self) -> "TensorMap":
...
 
+    def __reduce_ex__(self, protocol: int):
+        """
+        Used by the Pickler to dump TensorMap object to bytes object.
+        When protocol >= 5 it supports PickleBuffer which reduces number of
+        copyies needed
+        """
+        path = os.path.join(tempfile.gettempdir(), str(id(self)))
+        try:
+            self._lib.eqs_tensormap_save(path.encode("utf8"), self._ptr)
```
Luthaf, line 126: Same, this could use `equistore.io.save(path, self)`

Luthaf (commented):
Looks good overall!

python/src/equistore/tensor.py
agoscinski: I think this was because io.py has a dependency on TensorMap, so I could not import it here

python/src/equistore/tensor.py
agoscinski: moved

python/src/equistore/tensor.py
Luthaf: right, you can not import it on the top level, but I think importing inside the function should be fine. We do something similar for `__eq__` and friends

python/src/equistore/tensor.py
agoscinski: I think this was because io.py has a dependency on TensorMap, so I could not import it here

python/src/equistore/tensor.py
agoscinski: And what you think about the `create_numpy_array`  function? I don't need it anymore in the tensor.py for this PR, so I can remove the dependency and also move it back to io.py. But for the step 1, I might need to pass create_numpy_array 
```python
    @classmethod
    def _from_pickle(cls, data: bytes):
        """
        Passed to pickler to reconstruct TensorMap from bytes object
        """
        lib = _get_library()
        return lib.eqs_tensormap_load_from_buffer(data, eqs_create_array_callback_t(create_numpy_array))
```
except if we keep it in the io.py as well
```python
    @classmethod
    def _from_pickle(cls, data: bytes):
        """
        Passed to pickler to reconstruct TensorMap from bytes object
        """
        from equistore.io import load_from_buffer
        return load_from_buffer(data)
```
What do you prefer?

Luthaf: I would leave everything related to io in the `io` module

agoscinski:
wanted to fix warning but did not want to create a separate PR, so I kept 2 commits

python/src/equistore/tensor.py
```diff
@@ -88,6 +112,26 @@ def copy(self) -> "TensorMap":
...
+        """
+        Used by the Pickler to dump TensorMap object to bytes object.
+        When protocol >= 5 it supports PickleBuffer which reduces number of
+        copyies needed
+        """
+        from equistore.io import save
+
+        path = os.path.join(tempfile.gettempdir(), str(id(self)) + ".npz")
+        try:
+            save(path, self)
```
Luthaf, line 125:
```suggestion
        import equistore

        path = os.path.join(tempfile.gettempdir(), str(id(self)) + ".npz")
        try:
            equistore.save(path, self)
```

python/src/equistore/tensor.py
```diff
@@ -68,6 +76,22 @@ def _from_ptr(ptr):
...
+        """
+        Passed to pickler to reconstruct TensorMap from bytes object
+        """
+        from equistore.io import load
+
+        path = os.path.join(tempfile.gettempdir(), str(id(data)))
+        try:
+            with open(path, "wb") as f:
+                pickle.dump(data, f)
+            tensor_map = load(path)
```
Luthaf, line 90:
```suggestion
        import equistore

        path = os.path.join(tempfile.gettempdir(), str(id(data)))
        try:
            with open(path, "wb") as f:
                pickle.dump(data, f)
            tensor_map = equistore.load(path)
```

Luthaf (approved):
Two small changes and it is good to go!

Luthaf (approved):

PR #239, opened 2023-03-30
Labels.arange() function

frostedoyster:
----
:books: Documentation preview :books:: https://equistore--239.org.readthedocs.build/en/239/

### DIFF
--- a/python/src/equistore/labels.py
+++ b/python/src/equistore/labels.py
@@ -252,7 +252,7 @@ class Labels(np.ndarray):
         return self.position(label) is not None
 
     @staticmethod
-    def arange(*args, name="name"):
+    def arange(name, *args):
         """
         Creates a very simple `Labels` object.
 
--- a/python/src/equistore/labels.py
+++ b/python/src/equistore/labels.py
@@ -251,6 +251,30 @@ class Labels(np.ndarray):
     def __contains__(self, label):
         return self.position(label) is not None
 
+    def arange(*args, name="name"):
+        """Description"""
+
+        args_len = len(args)
+        if args_len == 0:
+            raise ValueError(
+                "please provide at least one integer for `Labels.arange()`"
+            )
+        if args_len > 3:
+            raise ValueError(
+                f"the maximum number of integer arguments accepted by `Labels.arange()` is 3. {args_len} were provided"
+            )
+        for arg in args:
+            if not np.issubdtype(type(arg), np.integer):
+                raise ValueError(
+                    "all numbers provided to `Labels.arange()` must be integers"
+                )
+
+        labels = Labels(
+            names=[name], values=np.arange(*args, dtype=np.int32).reshape(-1, 1)
+        )
+
+        return labels
+
 
 def _eqs_labels_view(array):
     """Create a new eqs_label_t where the values are a view inside the array"""
--- a/python/src/equistore/labels.py
+++ b/python/src/equistore/labels.py
@@ -251,8 +251,18 @@ class Labels(np.ndarray):
     def __contains__(self, label):
         return self.position(label) is not None
 
+    @staticmethod
     def arange(*args, name="name"):
-        """Description"""
+        """
+        Creates a very simple `Labels` object.
+
+        The resulting `Labels` only contain one label, whose name
+        is `name`, unless otherwise specified. The values of the
+        labels depend on a set of integer arguments, whose
+        meaning is the same as those of the `numpy.arange` function.
+        As a result, this function only accepts one to three such
+        integer parameters.
+        """
 
         args_len = len(args)
         if args_len == 0:
@@ -261,7 +271,8 @@ class Labels(np.ndarray):
             )
         if args_len > 3:
             raise ValueError(
-                f"the maximum number of integer arguments accepted by `Labels.arange()` is 3. {args_len} were provided"
+                f"the maximum number of integer arguments accepted by \
+                `Labels.arange()` is 3. {args_len} were provided"
             )
         for arg in args:
             if not np.issubdtype(type(arg), np.integer):
--- a/python/src/equistore/labels.py
+++ b/python/src/equistore/labels.py
@@ -254,14 +254,22 @@ class Labels(np.ndarray):
     @staticmethod
     def arange(name, *args):
         """
-        A `Labels` instance with evenly spaced values within a given interval.
+        A `Labels` instance with evenly spaced integers within a given interval.
 
         The resulting `Labels` only contain one label, whose name
-        is `name`, unless otherwise specified. The values of the
-        labels depend on a set of integer arguments, whose
-        meaning is the same as those of the `numpy.arange` function.
-        As a result, this function only accepts one to three such
-        integer parameters.
+        must be specified. The values of the labels depend on a set of
+        integer arguments, whose meaning is the same as those of the
+        `numpy.arange` function.
+        In particular:
+
+        :param `start`: (Optional) Lower bound of the range. If not specified,
+            it defaults to zero
+        :param `stop`: Upper bound of the range.
+        :param `step`: (Optional) Spacing within the range of values. If not
+            specified, it defaults to one
+
+        :returns: New :py:class:`Labels` object with the provided name and
+            values corresponding to the specified range bounds and step.
         """
 
         args_len = len(args)
--- a/python/src/equistore/labels.py
+++ b/python/src/equistore/labels.py
@@ -171,7 +171,7 @@ class Labels(np.ndarray):
 
             labels = Labels(
                 names=["structure", "atom", "center_species"],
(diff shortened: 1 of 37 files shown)


python/src/equistore/labels.py
```diff
@@ -251,6 +251,41 @@ def asarray(self):
...
+                `Labels.arange()` is 3. {args_len} were provided"
+            )
+        for arg in args:
+            if not np.issubdtype(type(arg), np.integer):
+                raise ValueError(
+                    "all numbers provided to `Labels.arange()` must be integers"
+                )
+
+        labels = Labels(
+            names=[name], values=np.arange(*args, dtype=np.int32).reshape(-1, 1)
```
jwa7, line 284: Did we remove the need to specify dtype here @Luthaf ?

jwa7 (changes requested):
Looks mostly good! 

Quick comment: in your docstring examples (i.e. in solve.py) you include calls like:
```py
samples = Labels.arange(2, "sample")
```
But this raises an error as the function tries to interpret "sample" as an int:
```
ValueError: all numbers provided to `Labels.arange()` must be integers
```

These docstring examples will need updating to pass the name as a keyword arg.

Also, in this case, it might be helpful to receive an error message that points the user towards passing the name as a keyword argument instead of a positional one.

python/src/equistore/labels.py
```diff
@@ -251,6 +251,41 @@ def asarray(self):
...
+        )
```
jwa7, line 285: Also, do we want to allow for creation of Labels of > 1 dimension?

Perhaps something like this, passing a list of names and using the length of this to reshape the np.arange array

```suggestion

        if isinstance(names, str):
            names = [names]
        if not np.all([isinstance(n, str) for n in names]):
            raise TypeError("all names provided to `Labels.arange()` must be strings")

        labels = Labels(
            names=names, values=np.arange(*args, dtype=np.int32).reshape(-1, len(names))
        )
```

python/src/equistore/labels.py
```diff
@@ -251,6 +251,41 @@ def asarray(self):
     def __contains__(self, label):
         return self.position(label) is not None
 
+    @staticmethod
+    def arange(*args, name="name"):
```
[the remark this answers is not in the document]
PicoCentauri: And explicit parameters please. Who knows what you want as `*args`.

python/src/equistore/labels.py
```diff
@@ -251,6 +251,41 @@ def asarray(self):
...
+        if args_len > 3:
+            raise ValueError(
+                f"the maximum number of integer arguments accepted by \
+                `Labels.arange()` is 3. {args_len} were provided"
+            )
+        for arg in args:
+            if not np.issubdtype(type(arg), np.integer):
+                raise ValueError(
+                    "all numbers provided to `Labels.arange()` must be integers"
+                )
```
PicoCentauri, line 281: I think you don't need any checks here. Just pass everything to `np.arange`. They will handle it no?

python/tests/labels.py
```diff
@@ -183,3 +183,36 @@ def test_labels_contiguous(self):
...
+    def test_arange_three_arguments(self):
+        labels_arange = Labels.arange(0, 10, 2, name="samples")
+        assert labels_arange.names == ("samples",)
+        assert np.array_equal(
+            labels_arange.asarray().reshape((-1,)), np.arange(0, 10, 2)
+        )
+
+    def test_arange_incorrect_arguments(self):
+        with pytest.raises(ValueError, match="3"):
+            _ = Labels.arange(0, 10, 2, 4, name="dummy")
```
PicoCentauri, line 208: You don't need to assign this.

jwa7:
Re: my suggestions about multi-dimensonal - sorry, just seen the discussion on ML dev about keeping it to one dimension. Happy for it to be that way for now but think it'll be useful to have the multidimensional case eventually.

python/src/equistore/labels.py
PicoCentauri: Maybe use `np.array(range(*args))[:, np.newaxis]`, then you are sure that everything is an integer.

python/src/equistore/labels.py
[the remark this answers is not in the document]
Luthaf: I would force an explicit name instead of having a default:

```
    @staticmethod
    def arange(name, *args):
```

or with explicit arguments

```
    @staticmethod
    def arange(name, start, stop=None, step=None):
```

python/tests/labels.py
```diff
@@ -183,3 +183,36 @@ def test_labels_contiguous(self):
...
+        labels_arange = Labels.arange(10, 42, name="dummy")
+        assert labels_arange.names == ("dummy",)
+        assert np.array_equal(labels_arange.asarray().reshape((-1,)), np.arange(10, 42))
+
+    def test_arange_three_arguments(self):
+        labels_arange = Labels.arange(0, 10, 2, name="samples")
+        assert labels_arange.names == ("samples",)
+        assert np.array_equal(
+            labels_arange.asarray().reshape((-1,)), np.arange(0, 10, 2)
+        )
```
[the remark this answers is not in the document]
frostedoyster: As requested

python/tests/labels.py
frostedoyster: Ok, I've removed that and all the other discarded values in the other tests

python/src/equistore/labels.py
frostedoyster: I feel like our own error messages can be more informative

python/src/equistore/labels.py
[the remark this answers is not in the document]
frostedoyster: I've made the name mandatory and I've documented the three integer parameters

python/src/equistore/labels.py
frostedoyster: I've tried to remove the np.int32 there and everywhere else in equistore

python/src/equistore/labels.py
PicoCentauri: fine for me

PicoCentauri (approved):
Looks good and very nice that your removed the `np.int32` from the other files.

frostedoyster:
@jwa7 Please don't merge... I messed up... There are so many more manual implementations that I missed simply because they weren't using np.arange

python/src/equistore/labels.py
```diff
@@ -251,6 +251,48 @@ def asarray(self):
     def __contains__(self, label):
         return self.position(label) is not None
 
+    @staticmethod
+    def arange(name, *args):
```
DavideTisi, line 255: one comment, nice function. For now the function just create a label with one name and only one arange.
Maybe it would be also helpful to have the possibility to create a label with more names, something like 

```python
arange(names=['a1', 'a2'],start=[0,3],stop=[2,7],step=[1,2])-> Labels(['a1', 'a2'],np.array([[0,3], [1,5]])```

the function should handle the errors if the the two arange are not compatible

File: python/src/equistore/labels.py (hunk 2 above)
Comment (jwa7) [reply python/src/equistore/labels.py]: Decided to keep to one dimension

**Review by jwa7 — COMMENTED** [review]
Looking great, thanks @frostedoyster ! Also thanks for cleaning up the dtype stuff - much cleaner now. I removed a few remaining dtypes and merged master into here so we're ready to merge.

Before we merge, please could you address @DavideTisi 's comment above?

**Comment by DavideTisi** [comment]
@jwa7 but my comment was just to do the multi-dimensional thing. when I wrote the comment, I hadn't read the post on slack channel. 
My idea was basically just what Michele was suggesting there, but since it has been decided to do it only 1D for me it is ok.

For the rest it is super fine, I am not a super fan of having a function that wraps basically a single line, but I do not have much experience as a user, and the frustrations that come with it.

**Review by jwa7 — APPROVED** [review]

**Comment by jwa7** [comment]
OK thanks @DavideTisi - merged :)

File: python/src/equistore/labels.py [hunk 8]
```diff
@@ -251,6 +251,48 @@ def asarray(self):
...
+        integer arguments, whose meaning is the same as those of the
+        `numpy.arange` function.
+        In particular:
+
+        :param `name`: Name of the resulting labels
+        :param `start`: (Optional) Lower bound of the range. If not specified,
+            it defaults to zero
+        :param `stop`: Upper bound of the range
+        :param `step`: (Optional) Spacing within the range of values. If not
+            specified, it defaults to one
```
Luthaf, line 270: This documentation indicate that users can call this function with `Labels.arange(name="s", start=3, stop=4)` which is not the case. We could also have the function take `*args, **kwargs` and check both, I would first check how numpy handles this.

frostedoyster: I tried to check numpy's function arguments, but it seems to be written in C, so I gave up. I agree that using and checking kwargs is a good way to solve this. Probably also an example would help. I will work on this

PR #240, opened 2023-03-31
Remove dependency on py-litteral and pest

Luthaf:
These are used to implement the npz reader/writer for `equistore::io::save`/`equistore::io::load`, but are relatively heavy dependencies for what we do with them. Removing them shaves 20% out of equistore compile time.

Instead of the dependency, the code now uses a small handwritten parser for the couple of Python literals we need to read instead of a generated [pest](https://docs.rs/pest/latest/pest/) parser.

---

I had this branch sitting on my computer for a while, it might be time to merge it =)

----
:books: Documentation preview :books:: https://equistore--240.org.readthedocs.build/en/240/

### DIFF
--- a/equistore-core/src/io/npy_header.rs
+++ b/equistore-core/src/io/npy_header.rs
@@ -1,14 +1,12 @@
-// This file was taken from https://github.com/jturner314/ndarray-npy, version
-// 0.8.1. It is Copyright 2018–2021 Jim Turner and ndarray-npy developers,
-// released under MIT and Apache Licenses.
-
-use byteorder::{ByteOrder, LittleEndian, ReadBytesExt};
-use num_traits::ToPrimitive;
-use py_literal::{
-    FormatError as PyValueFormatError, ParseError as PyValueParseError, Value as PyValue,
-};
+// This file was initially taken from https://github.com/jturner314/ndarray-npy,
+// version 0.8.1. It is Copyright 2018–2021 Jim Turner and ndarray-npy
+// developers, released under MIT and Apache Licenses.
 use std::convert::TryFrom;
 use std::error::Error;
+use std::fmt::Write as FmtWrite;
+use std::io::Write as IoWrite;
+
+use byteorder::{ByteOrder, LittleEndian, ReadBytesExt};
 
 /// Magic string to indicate npy format.
 const MAGIC_STRING: &[u8] = b"\x93NUMPY";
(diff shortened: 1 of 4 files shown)


equistore-core/src/io/npy_header.rs
```diff
@@ -496,7 +632,34 @@ impl From<WriteHeaderError> for crate::Error {
     fn from(error: WriteHeaderError) -> Self {
         match error {
             WriteHeaderError::Io(e) => crate::Error::Io(e),
-            WriteHeaderError::Format(e) => crate::Error::Serialization(e.to_string()),
+            WriteHeaderError::Format(e) => crate::Error::Serialization(e),
         }
     }
 }
+
+#[cfg(test)]
```
PicoCentauri, line 640: Shouldn't there be also a test for the writing?

equistore-core/src/io/npy_header.rs
```diff
@@ -308,79 +256,269 @@ impl From<std::io::Error> for WriteHeaderError {
...
+        self.expects('{')?;
+        loop {
+            let key = self.parse_string()?;
+            self.skip_whitespaces();
+            self.expects(':')?;
+            self.skip_whitespaces();
+
+            if key == "descr" {
+                type_descriptor = Some(self.parse_data_type()?);
+            } else if key == "fortran_order" {
```
PicoCentauri, line 480: What about "c_order"?

PicoCentauri (changes requested):
Less compile time is good. I scanned through the code and it looks good enough for us. I have a couple of minor questions.

equistore-core/src/io/npy_header.rs
Luthaf: the npy format stores `fortran_order: true` or `fortran_order: false` directly

equistore-core/src/io/npy_header.rs
Luthaf: This is the `to_dict_literal()` checks below. I could rename the test function name

PicoCentauri: No its fine.

equistore-core/src/io/npy_header.rs
PicoCentauri: Ahh I see.

PicoCentauri (approved):
I like!

PR #242, opened 2023-04-02
Provide kwargs for Labels.arange()

frostedoyster:
----
:books: Documentation preview :books:: https://equistore--242.org.readthedocs.build/en/242/


python/src/equistore/labels.py
```diff
@@ -271,25 +271,81 @@ def arange(name, *args):
...
+                    "the maximum number of integer arguments accepted by "
+                    f"`Labels.arange()` is 3. {args_len} were provided"
                 )
-
-        labels = Labels(names=[name], values=np.arange(*args).reshape(-1, 1))
+            for arg in args:
+                if not np.issubdtype(type(arg), np.integer):
+                    raise ValueError(
+                        "all numbers provided to `Labels.arange()` must be integers"
+                    )
```
Luthaf, line 326: Do we need to check this ourself? Shouldn't numpy do this for us?

frostedoyster: No, unfortunately `np.arange` also works with floats, and it gives float ranges out in that case, which would not work for us. It works a bit like linspace with slightly different arguments in that case

frostedoyster: Ok, perfect. I've added the comments

PicoCentauri:
Before we merge this can we try to merge the changes of the tests in #235. I think rebasing here is easier than if I have to rebase #235 again.

python/tests/labels.py
```diff
@@ -202,3 +210,11 @@ def test_arange_incorrect_arguments():
...
     with pytest.raises(ValueError, match="integer"):
         Labels.arange("dummy", 0, 5, 0.2)
+    with pytest.raises(ValueError, match="argument"):
+        Labels.arange("dummy", random=10)
+    with pytest.raises(EquistoreError, match="label name"):
+        Labels.arange(0, 1, 2)
+    with pytest.raises(ValueError, match="integer"):
+        Labels.arange(name="dummy", start=0.0, stop=1.0, step=2)
+    with pytest.raises(ValueError, match="integer"):
+        Labels.arange("dummy", start=0, stop=0.2)
```
Luthaf, line 220: Could you update these to use the full error message?

frostedoyster: Sure, done

Luthaf:
The history on this branch is weird ... Did you rebase and then merge? 

Could you squash all commits to a single one to make sure we understand what's being changed here?

frostedoyster:
Yes... How do I do that?

Luthaf:
> How do I do that?

This is a good introduction: https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history. You'll have to fix any potential conflicts as you go through the commits.

If you are stuck, let me know and we can do this together!

Luthaf:
This is now #255!

PR #243, opened 2023-04-02
One-hot encoding of labels

frostedoyster:
@Luthaf @abmazitov
This would be very helpful for a memory-efficient PyTorch implementation of any type of alchemical compression. Please have a look at how the example works and let me know your thoughts.

----
:books: Documentation preview :books:: https://equistore--243.org.readthedocs.build/en/243/


frostedoyster:
It returns a np.array, and not a `Labels` object as advertised. Initially, I thought that would be a good idea, but clearly there can be duplicates, hence `Labels` is out of question

Luthaf:
The code looks good, I'm not sure how one would use it but I'm fine merging this! It needs a test and to be added to the documentation.

Regarding naming, is there some existing library using the `one_hot` name for a similar function?

frostedoyster:
Yes, there is such a similar function: https://pytorch.org/docs/stable/generated/torch.nn.functional.one_hot.html
One-hot encoding allows the one-hot output array to be multiplied by the element weight matrix directly, thereby avoiding having to store all expansion coefficients before the contraction, but directly using each pair contribution to add to the correct pseudo-element channel with its correct weight contribution

Luthaf:
Ok, the behavior is similar enough to use the same name! This just needs tests & doc then

python/tests/operations/one_hot.py
```diff
@@ -0,0 +1,90 @@
+import numpy as np
+import pytest
+
+import equistore
+from equistore import Labels
+
+
+class TestOneHot:
```
Luthaf, line 8: Sorry I missed this on the previous review rounds. You don't need a class here, the test functions can be defined directly in the file.

Luthaf (approved):
One last small changed and this is good to go!

PR #245, opened 2023-04-03
changing the arguments of slice()

### LINKED ISSUE #144: Should `slice()` have the arg `axis`, like operations `split()`, and `join()`?
Currently, the operations `split` and `join` take the `axis` argument to specify whether splitting or joining should be performed over the "samples" or "properties" axis. The `slice` function is written to allow users to slice over both "samples" and "properties" in the same function call, using the args `slice(tensor, samples=Labels(...), properties=Labels(...))`. 

After discussion with @PicoCentauri , we are wondering if the the arguments should be made consistent, such that slice only works over a single axis at once, specifying the axis with the `axis` argument. This would be a breaking change, and mean that users would have to make 2 separate function calls to slice along samples and properties. However, the change feels more numpy-esque.

Old behaviour:
```py
from equistore.operations import slice
...
sliced_tensor = slice(tensor, samples=samples_labels, properties=properties_labels)
```

Proposed new behaviour (with the same result as above):
```py
sliced_tensor_along_samples = slice(tensor, axis="samples", labels=samples_labels)
sliced_tensor_along_properties = slice(sliced_tensor_along_samples, axis="properties", labels=properties_labels)
```

**To Do**

- [ ] Change the args of all functions in `slice.py` to have `axis` and `labels`
- [ ] Update function calls in tests/.../slice.py
- [ ] For the test class TestSliceBoth, update the single function call to now be 2 function calls
- [ ] **Bonus test**: Check that slicing samples then properties gives the same result as slicing properties then samples.

cbenmahm:
fix #144 

----
:books: Documentation preview :books:: https://equistore--245.org.readthedocs.build/en/245/


python/src/equistore/operations/slice.py
```diff
@@ -196,119 +196,135 @@ def _slice_block(
     """
     # Store current values for later modification
     new_values = block.values
-    new_samples = block.samples
-    new_properties = block.properties
+    if axis == "samples":
+        new_samples = block.samples
+    elif axis == "properties":
```
jwa7, line 201:
```suggestion
    else:  # properties
```

python/src/equistore/operations/slice.py
```diff
@@ -196,119 +196,135 @@ def _slice_block(
     """
```
jwa7, line 196: The docstring needs to be updated to the new args here

python/src/equistore/operations/slice.py
```diff
@@ -196,119 +196,135 @@ def _slice_block(
...
+    if axis == "samples":
+        new_samples = block.samples
+    elif axis == "properties":
+        new_properties = block.properties
 
     # Generate arrays of bools indicating which samples indices to keep upon slicing.
-    if samples is not None:
-        all_samples = block.samples[list(samples.names)].tolist()
-        set_samples_to_slice = set(samples.tolist())
+    if axis == "samples" and labels is not None:
```
jwa7, line 205:
```suggestion
    if axis == "samples":
```

jwa7 (changes requested):
Looking mostly good thanks @cbenmahm ! Suggestions are mostly to do with the change of args to be mandatory, but also some docstrings need updating.

One more thing: it would be great if we could change the arg `grouped_idxs` in the `split.py` functions (and tests) to be `grouped_labels`, re: @Luthaf 's comment on the original issue #144 :)

python/src/equistore/operations/slice.py
```diff
@@ -196,119 +196,132 @@ def _slice_block(
...
+                    )
+                    new_grad_samples[:, 0] = sample_map[new_grad_samples[:, 0]]
+
+                    new_grad_samples = Labels(
+                        names=gradient.samples.names,
+                        values=new_grad_samples,
+                    )
+
+                new_grad_data = new_grad_data[grad_samples_filter]
+            else:  # properties
```
Luthaf, line 278:
```suggestion
            else:
                assert axis == "properties"
```

python/src/equistore/operations/slice.py
```diff
@@ -63,44 +67,39 @@ def slice(
...
-    samples: Optional[Labels] = None,
-    properties: Optional[Labels] = None,
+    axis: str,
+    labels: Labels,
 ) -> TensorBlock:
     """
-    Slices an input :py:class:`TensorBlock` along the samples and/or properties
-    dimension(s). ``samples`` and ``properties`` are
+    Slices an input :py:class:`TensorBlock` along either the "samples" or "properties"
+    `axis`. dimension(s). ``labels`` is a
```
jwa7, line 97:
```suggestion
    `axis`.  `labels` is a
```

python/src/equistore/operations/slice.py
```diff
@@ -63,44 +67,39 @@ def slice(
...
     :py:class:`Labels` objects that specify the samples/properties
-    (respectively) names and indices that should be sliced, i.e. kept in the
+    names and indices that should be sliced, i.e. kept in the
     output :py:class:`TensorBlock`.
 
-    Note that either ``samples`` or ``properties``, or both,
+    Note that either ``samples`` or ``properties``
     should be specified as input.
```
jwa7, line 104:
```suggestion
```

python/src/equistore/operations/slice.py
```diff
@@ -157,79 +161,88 @@ def slice_block(
...
+    axis: str,
+    labels: Labels,
 ) -> TensorBlock:
     """
-    Slices an input :py:class:`TensorBlock` along the samples and/or properties
-    dimension(s). ``samples`` and ``properties`` are
-    :py:class:`Labels` objects that specify the samples/properties
-    (respectively) names and indices that should be sliced, i.e. kept in the
+    Slices an input :py:class:`TensorBlock` along either the "samples" or "properties"
+    `axis`. dimension(s). ``labels`` is
```
jwa7, line 180:
```suggestion
    `axis`. `labels` is
```

python/src/equistore/operations/slice.py
```diff
@@ -157,79 +161,88 @@ def slice_block(
...
+    :py:class:`Labels` object that specify the samples/properties
+    names and indices that should be sliced, i.e. kept in the
     output :py:class:`TensorBlock`.
 
     Note that either ``samples`` or ``properties``, or both,
     should be specified as input.
```
jwa7, line 187:
```suggestion
```

python/src/equistore/operations/slice.py
```diff
@@ -275,40 +289,43 @@ def _slice_block(
...
             if name not in s_names:
                 raise ValueError(
                     f"invalid sample name '{name}' which is not part of the input"
                 )
     # Check properties Labels if passed
-    if properties is not None:
-        # Check type
-        if not isinstance(properties, Labels):
-            raise TypeError("properties must be a `Labels` object")
+    else:  # properties
```
jwa7, line 325:
```suggestion
    else:
        assert axis == "properties"
```

python/src/equistore/operations/slice.py
```diff
@@ -157,79 +161,88 @@ def slice_block(
...
     :return new_block: a :py:class:`TensorBlock` that corresponds to the sliced
         input.
     """
     # Store current values for later modification
     new_values = block.values
-    new_samples = block.samples
-    new_properties = block.properties
+    if axis == "samples":
+        new_samples = block.samples
+    else:  # properties
```
jwa7, line 202:
```suggestion
    else:
        assert axis == "properties"
```

python/tests/operations/join.py
```diff
@@ -24,214 +24,205 @@ def tensor():
 
     return tensor
 
+    @pytest.fixture
```
Luthaf, line 27: why are these indented?

Luthaf (commented):
Please also update slice tests to use pytest, you are already changing most of them anyway

python/tests/operations/join.py
jwa7: Good spot, thanks! Fixed.

jwa7:
Made some changes; fixed the indentation issue with the join tests, and changed the split arg from `grouped_idxs` to `grouped_labels`.

Still to do: port the slice tests to pytest

jwa7:
Tests ported to PyTest by @cbenmahm . I made some minor changes, removing the test classes.

Luthaf (approved):

PR #248, opened 2023-04-05
`drop_blocks()` - add a `copy` flag

jwa7:
This flag dictates whether or not the underlying data of the input TensorBlocks (and Gradients) should be copied for the returned TensorMap.

Setting to False, the the new TensorMap is constructed with a reference to the original data. This of course could be unsafe if the user manipulates via one reference and unknowingly changes the other. However, the advantage for the careful user is that a large increase in memory usage is not created.

Default value is set to False.

Also done in this PR (and not in a separate one, for no good reason) was a removal of the test class in the test script unique_metadata.py

----
:books: Documentation preview :books:: https://equistore--248.org.readthedocs.build/en/248/

### DIFF
--- a/python/src/equistore/operations/drop_blocks.py
+++ b/python/src/equistore/operations/drop_blocks.py
@@ -30,9 +30,7 @@ def drop_blocks(tensor: TensorMap, keys: Labels, copy: bool = False) -> TensorMa
             f"input `keys` must be a Labels object, got '{type(keys)}' instead"
         )
     if not isinstance(copy, bool):
-        raise TypeError(
-            f"`copy` flag must be a boolean, got '{type(copy)}' instead"
-        )
+        raise TypeError(f"`copy` flag must be a boolean, got '{type(copy)}' instead")
 
     if not np.all(tensor.keys.names == keys.names):
         raise ValueError(
@@ -57,18 +55,18 @@ def drop_blocks(tensor: TensorMap, keys: Labels, copy: bool = False) -> TensorMa
         for key in new_keys:
             # Create new block
             new_block = TensorBlock(
+                values=tensor[key].values,
                 samples=tensor[key].samples,
                 components=tensor[key].components,
                 properties=tensor[key].properties,
-                values=tensor[key].values,
             )
             # Add gradients
             for parameter, gradient in tensor[key].gradients():
                 new_block.add_gradient(
-                    parameter=param,
-                    samples=obj.samples,
-                    components=obj.components,
-                    data=obj.data,
+                    parameter=parameter,
+                    data=gradient.data,
+                    samples=gradient.samples,
+                    components=gradient.components,
                 )
             new_blocks.append(new_block)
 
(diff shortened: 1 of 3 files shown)


jwa7:
After discussion with @Luthaf , the default flag has been changed to ``False``

python/src/equistore/operations/drop_blocks.py
```diff
@@ -1,30 +1,75 @@
...
 
-    :return: the input :py:class:`TensorMap` with the specified key/block
-        pairs dropped.
+    :return: the input :py:class:`TensorMap` with the specified key/block pairs
+        dropped.
     """
+    # Check arg types
+    if not isinstance(tensor, TensorMap):
+        raise TypeError(
+            "The input `tensor` must be a TensorMap. " f"Got {type(tensor)} instead."
```
Luthaf, line 26:
```suggestion
            f"input `tensor` must be a TensorMap, got '{type(tensor)}' instead"
```

python/src/equistore/operations/drop_blocks.py
```diff
@@ -1,30 +1,75 @@
...
+        )
+    if not isinstance(keys, Labels):
+        raise TypeError(
+            "The input `keys` must be a Labels object. " f"Got {type(keys)} instead."
```
Luthaf, line 30:
```suggestion
            f"input `keys` must be a Labels object, got '{type(keys)}' instead"
```

python/src/equistore/operations/drop_blocks.py
```diff
@@ -1,30 +1,75 @@
...
+        )
+    if not isinstance(copy, bool):
+        raise TypeError(
+            "The input `copy` flag must be a boolean. " f"Got {type(copy)} instead."
```
Luthaf, line 34:
```suggestion
            f"`copy` flag must be a boolean, got '{type(copy)}' instead"
```

python/src/equistore/operations/drop_blocks.py
```diff
@@ -1,30 +1,75 @@
...
+        for key in new_keys:
+            # Create new block
+            new_block = TensorBlock(
+                samples=tensor[key].samples,
+                components=tensor[key].components,
+                properties=tensor[key].properties,
+                values=tensor[key].values,
+            )
+            # Add gradients
+            for param, obj in tensor[key].gradients():
```
Luthaf, line 66:
```suggestion
            for parameter, gradient in tensor[key].gradients():
```

python/src/equistore/operations/drop_blocks.py
```diff
@@ -1,30 +1,75 @@
...
+                new_block.add_gradient(
+                    parameter=param,
+                    samples=obj.samples,
+                    components=obj.components,
+                    data=obj.data,
```
Luthaf, line 71: I see you like this order for the arguments, but I tend to prefer following the order in the function definition, i.e. `parameter`, `data`, `samples`, `components`

jwa7: Updated :)

Luthaf (approved):

Issue #247, opened 2023-04-05
Data type conflict in reduce_over_samples while using torch Tensors


abmazitov:
There is a data type conflict in `reduce_over_samples`, if it is done on a tensormap with `torch.Tensor` type of data inside.

--- 

In a file `equistore.operations.reduce_over_samples.py` 
In a function `_reduce_over_samples_block()`
Lines 87-89:
```
if reduction == "mean" or reduction == "std" or reduction == "variance":
        bincount = _dispatch.bincount(index)
        values_result = values_result / bincount.reshape(
            (-1,) + (1,) * len(other_shape)
        )
```

`values_result` is a `torch.Tensor`, but `bincount` and `index` are numpy arrays.

The `index` object actually comes from lines 71-73, where datatype management is omitted.

```
 new_samples, index = np.unique(
        samples[:, sample_selected], return_inverse=True, axis=0
    )
``` 

---

This cause a conversion of a torch Tensor to a numpy format and ruins the workflow.

abmazitov:
I would suggest to add a `unique()` function into `_dispatch.py` and maintain a datatype inside

DavideTisi:
mmm i see so index is a numpy array becouse it comes from an operation on a sample which is a numpy array

abmazitov:
> 

Exactly. In this case `samples` has a np.int32 type, and we run `np.unique` on top of it. It means that we have 2 potential sources of unexpected dtype conversion.

DavideTisi:
@Luthaf have any idea how to fix it or we just add some hacks?

Luthaf:
Should be fixed by #250

PR #250, opened 2023-04-06
quick fix for type mistmatch

DavideTisi:
fix for #247 

----
:books: Documentation preview :books:: https://equistore--250.org.readthedocs.build/en/250/

### DIFF
--- a/python/src/equistore/operations/_dispatch.py
+++ b/python/src/equistore/operations/_dispatch.py
@@ -65,6 +65,7 @@ def bincount(input, weights=None, minlength=0):
                                         Defaults to None.
         minlength (int, optional): A minimum number of bins for the output array.
                                         Defaults to 0.
+        out_format
     """
     if isinstance(input, np.ndarray):
         if weights is not None:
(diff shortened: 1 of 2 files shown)


DavideTisi:
it should fix the problem but it is not tested yet

python/src/equistore/operations/_dispatch.py
```diff
@@ -65,6 +65,7 @@ def bincount(input, weights=None, minlength=0):
                                         Defaults to None.
         minlength (int, optional): A minimum number of bins for the output array.
                                         Defaults to 0.
+        out_format
```
Luthaf, line 68: looks like a leftover

Luthaf: Could you also add a link to `array_like_data` here? Pointing future devs to this problem =)

Luthaf (commented):
You've checked elsewhere that there are no other usage of bincount, right?

Luthaf:
@abmazitov I'll let you check this fixes the issue!

python/src/equistore/operations/_dispatch.py
DavideTisi: not sure what you meant but i added a comment

DavideTisi:
i rebased and fixed the linter with `blackdoc`. For me it can be merged

python/src/equistore/operations/_dispatch.py
DavideTisi: 👍

PR #251, opened 2023-04-09
added unsafe to join (with tests)

DavideTisi:
the first instance of the `unsafe` flag added to join ( which seems the most problematic).
I haven't benchmarked yet (maybe someone wants to do it). 
So my main concern now is that I can only skip a few checks, but I still left the "mechanics" of constructing a new tensor and then applying `key_to_properties/values`. 
In particular, I cannot do the trivial thing I wanted, which is constricting a new block with values ``np.stack(b1.values,b2.values,axis=axis)`` because then I have to handle the labels, which are not a simple stack since I have to add a new name (`tensor` )with relative values. Constructing new labels would imply making a loop over samples or properties, which is probably. I couldn't find a better way to do it, rather than leave the `key_to_*` mechanics.

Another thing: now the code always assumes that the tensors have the same names for samples and properties. I do not know if it is what we want.

### DIFF
--- a/python/src/equistore/operations/join.py
+++ b/python/src/equistore/operations/join.py
@@ -9,7 +9,7 @@ from ..tensor import TensorBlock, TensorMap
 from .equal_metadata import _check_maps
 
 
-def join(tensor_maps: List[TensorMap], axis: str):
+def join(tensor_maps: List[TensorMap], axis: str, unsafe=False):
     """Join a sequence of :py:class:`TensorMap` along an axis.
 
     The ``axis`` parameter specifies the type join. For example, if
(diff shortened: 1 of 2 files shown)


python/src/equistore/operations/join.py
```diff
@@ -48,36 +53,38 @@ def join(tensor_maps: List[TensorMap], axis: str):
...
+        names_are_same = np.all(
+            len(unique_names) == np.array([len(names) for names in names_list])
         )
 
+        # It's fine to loose metadata on the property axis, less so on the sample axis!
+        if axis == "samples" and not names_are_same:
+            raise ValueError(
+                "Sample names are not the same! Joining along samples with different "
+                "sample names will loose information and is not supported."
+            )
```
PicoCentauri, line 86: These are no tests but are required to construct the new names.

PicoCentauri (commented):
The `join` function got a major speedup by factor with its last update #222. The only place where we could gain a speed up from the Python side is removing the `_check_maps` functions.

But I would vote for altering these check functions: i.e. `_check_maps(..., usafe=True)`. Then we have an affect to all `operations` and do not have to introduce `unsafe` blocks everyhwere. Also wit the latter block we can control that we keep some simple size checks.

python/src/equistore/operations/join.py
DavideTisi: yes i know, the idea here is to avoid assuming that the naming scheme. i do not know if it saves something

PicoCentauri:
Yes, I am also happy if we provide the two API's. Should we do this within this PR or open a new one?

PicoCentauri:
Hey @DavideTisi can you rebase and use the syntax of #254?

DavideTisi:
since the unsafe gave no speed up i would close this pr

PR #252, opened 2023-04-11
Spring cleaning for tests and operations

### LINKED ISSUE #175: Equal/allclose and friends should not hide ValueError
comparing two empty tensor maps (same keys but the blocks are an empty list) using `equistore.equal` gives a `False`

Luthaf:
This PR is better reviewed one commit at the time!

This contains multiple small improvements to:

- operations parameter names
- tests readability
- porting some tests to pytest

This also fix #175 as a side effect

---

I can split the PR in one PR per commit if you prefer!

----
:books: Documentation preview :books:: https://equistore--252.org.readthedocs.build/en/252/

### DIFF
--- a/python/src/equistore/operations/allclose.py
+++ b/python/src/equistore/operations/allclose.py
@@ -17,19 +17,19 @@ def allclose(
     """
     Compare two :py:class:`TensorMap`.
 
-    This function returns ``True`` if the two tensors have the same keys
+    This function returns :py:obj:`True` if the two tensors have the same keys
     (potentially in different order) and all the :py:class:`TensorBlock` have
     the same (and in the same order) samples, components, properties, and their
     values matrices pass the numpy-like ``allclose`` test with the provided
     ``rtol``, and ``atol``.
 
     The :py:class:`TensorMap` contains gradient data, then this function only
-    returns ``True`` if all the gradients also have the same samples,
+    returns :py:obj:`True` if all the gradients also have the same samples,
     components, properties and their data matrices pass the numpy-like
     ``allclose`` test with the provided ``rtol``, and ``atol``.
 
     In practice this function calls :py:func:`allclose_raise`, returning
-    ``True`` if no exception is raised, ``False`` otherwise.
+    :py:obj:`True` if no exception is raised, :py:obj:`False` otherwise.
 
     :param tensor_1: first :py:class:`TensorMap`
     :param tensor_2: second :py:class:`TensorMap`
@@ -252,7 +252,7 @@ def allclose_block(
     """
     Compare two :py:class:`TensorBlock`.
 
-    This function returns ``True`` if the two :py:class:`TensorBlock` have the
+    This function returns :py:obj:`True` if the two :py:class:`TensorBlock` have the
     same samples, components, properties and their values matrices must pass the
     numpy-like ``allclose`` test with the provided ``rtol``, and ``atol``.
 
@@ -262,7 +262,7 @@ def allclose_block(
     provided ``rtol``, and ``atol``.
 
     In practice this function calls :py:func:`allclose_block_raise`, returning
-    ``True`` if no exception is raised, ``False`` otherwise.
+    :py:obj:`True` if no exception is raised, :py:obj:`False` otherwise.
 
     :param block_1: first :py:class:`TensorBlock`
     :param block_2: second :py:class:`TensorBlock`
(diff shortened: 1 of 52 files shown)


python/src/equistore/operations/allclose.py
```diff
@@ -27,13 +29,13 @@ def allclose(
     ``allclose`` test with the provided ``rtol``, and ``atol``.
 
     In practice this function calls :py:func:`allclose_raise`, returning
-    ``True`` if no exception is rased, `False` otherwise.
+    ``True`` if no exception is raised, ``False`` otherwise.
```
PicoCentauri, line 32: We should use be py:obj:`True` instead of ``True``. With this Sphinx will render this as a link to the Python documentation. We could also include [flake8-sphinx-links](https://pypi.org/project/flake8-sphinx-links) which will test this for us.

PicoCentauri (changes requested):
Some fist comments but good cleanup!

python/src/equistore/operations/allclose.py
Luthaf: Do we really need links for True/False? Other than that, sure I can add the links =)

python/src/equistore/operations/unique_metadata.py
```diff
@@ -90,10 +90,10 @@ def unique_metadata(
     :param names: a ``str``, ``list`` of ``str``, or ``tuple`` of ``str``
         corresponding to the name(s) of the indices along the specified ``axis``
         for which the unique values should be found.
-    :param gradient_param: a ``str`` corresponding to the gradient parameter
-        name for the gradient blocks to find the unique indices for. If none
-        (default), the unique indices of the regular :py:class:`TensorBlock`
-        objects will be calculated.
+    :param gradient: a ``str`` corresponding to the gradient parameter name for
+        the gradient blocks to find the unique indices for. If none (default),
```
DavideTisi, line 94:
```suggestion
        the gradient blocks to find the unique indices for. If None (default),
```

DavideTisi (commented):
I didn't look at the allclose/equal thing yet but so far it is ok

Luthaf:
Now with four more commits!

- Remove unnecessary noqa in operations/__init__.py
- Remove "here is an example for this function" from the docstrings
- Rename "parameters"=>"gradients" in XXX_like (and port their tests to pytest)
- Add links to Python constants in docstrings

python/src/equistore/operations/unique_metadata.py
```diff
@@ -90,8 +90,8 @@ def unique_metadata(
     :param names: a ``str``, ``list`` of ``str``, or ``tuple`` of ``str``
         corresponding to the name(s) of the indices along the specified ``axis``
         for which the unique values should be found.
-    :param gradient_param: a ``str`` corresponding to the gradient parameter
-        name for the gradient blocks to find the unique indices for. If none
+    :param gradient: a ``str`` corresponding to the gradient parameter name for
```
PicoCentauri, line 93: `gradient` can also be a list. We should reflect this in the docs.

python/src/equistore/operations/empty_like.py
```diff
@@ -21,14 +21,12 @@ def empty_like(
...
-    Here is an example using this function. First we create a ``TensorMap`` with
-    just one block with two gradients, named ``alpha`` and ``beta``, containing
-    random data for values and gradients.
-
     >>> import numpy as np
     >>> import equistore
     >>> from equistore import TensorBlock, TensorMap, Labels
     >>> np.random.seed(1)
+    >>> # First we create a TensorMap with just one block with two gradients,
+    >>> # named 'alpha' and 'beta', containing random data
```
PicoCentauri, line 35: Can we move this above the example? I think this disturbs a bit reading the code.

PicoCentauri (changes requested):
Super nice. The only thing I am not sure is if putting comments as inline comments to the codes helps with reading the code example.

python/src/equistore/operations/unique_metadata.py
Luthaf: No, `names` can be a list but not `gradient`

python/src/equistore/operations/empty_like.py
Luthaf: I pushed this as a fixup commit f6b511bb3b9a0f0cc7d56433f4e41d8198f8713c, I'll rebase to merge it with the corresponding commit.

PicoCentauri (approved):
Good to gooo!

DavideTisi (approved):
Very nice!!! Great spring cleaning

Issue #56, opened 2022-10-13
Support for DLPack interchange format


Luthaf:
DLPack is an existing standard, already implemented and supported by multiple frameworks, to share arrays between these frameworks without a copy. https://dmlc.github.io/dlpack/latest/index.html

We already support it for values/gradients in Python (since these will return numpy arrays or torch tensors), but we should explore how we can add support for it in the Rust/C/C++ API both as output (transforming `eqs_array_t` to `DLManagedTensor`) and as input (wrapping `DLManagedTensor` into `eqs_array_t`).

Luthaf:
Done in #1045

PR #253, opened 2023-04-12
Gradients as full TensorBlocks

### LINKED ISSUE #96: Can we standardize the access of the arrays gradients and values?
For the implementations in equisolve we are writing a lot of functions that can access the values or the gradients of a tensor block and then do something with it. Because the values and gradients are accessed differently, we have to write this kind of code
```python 
data_keys =  [...] # allowed are 'values' and everything that is in gradients, e.g. 'positions' or 'cell'
for data_key in data_keys:
    if data_key == "values":
        property_len = len(tensor_block.values)
    else:
        property_len = len(tensor_block.gradient(data_key).data)
```

the handling would be simpler if we could access values and gradients the same way

```python
data_keys =  [...] # allowed are 'values' and everything that is in gradients, e.g. 'positions' or 'cell'
for data_key in data_keys:
    property_len = len(tensor_block.data(data_key))
```
This should work similar as `tensor_block.gradient(parameter_key)`. What might be confusing is that data here is a method and in on the gradients level it is member variable. But this is related to the inconsistency between obtaining the raw array values of a tensor block and of a gradient tensor block (one uses `values` but for gradents one uses `data`).

### LINKED ISSUE #219: Add support for higher order derivatives
We currently only support storing first derivatives in equistore, but there are multiple cases where we might want to store higher order derivatives. A simple way to allow that would be to modify the definition of a TensorBlock to make the gradients full tensor blocks as well (meaning they can have their own gradients)

```rs
// current
struct BasicBlock {
    data: Array, 
    samples: Labels,
    components: Vec<Labels>,
    properties: Labels,
}

struct TensorBlock {
    values: BasicBlock,
    gradients: HashMap<String, BasicBlock>
}

// Proposed
struct TensorBlock {
    values: Array, 
    samples: Labels,
    components: Vec<Labels>,
    properties: Labels,

    gradients: HashMap<String, TensorBlock>
}
```

This will require changing the C/C++/Rust API, but the Python API should stay very similar.

Luthaf:
This PR is built on top of #252.

Fixes #96
Fixes #219

----

This changes the core data structures to make gradients of a TensorBlock full tensor blocks, simplifying how we handle them and enabling storage of higher order derivatives.

This also required changing the serialization format to store gradients of gradients. There is a script to convert from the old to the new format, but users needs to be aware of it.

All operations currently throw a `NotImplementedError` if they encounter gradients of gradients, updating them is left to future PR.

TODO:

- [ ] detect old format when loading files and point the user to the update script
- [ ] update keys_to_samples/keys_to_properties to support gradients of gradients. This might be left to a future PR as well


Luthaf:
One point we discussed with @frostedoyster: currently, building gradients of gradient needs to happen in this order:

```py
grad_grad = TensorBlock(...)
grad = TensorBlock(...)
grad.add_gradient("positions", grad_grad)

block = TensorBlock(...)
block.add_gradient("positions", grad)
```

It would be nice to be able to write this code like this:

```py
block = TensorBlock(...)
block.add_gradient("positions", TensorBlock(...))
block.gradient("positions").add_gradient("positions", TensorBlock(...))
```

The issue is that the return value of `block.gradient("positions")` is a view of a TensorBlock inside something else, and we currently do not differentiate between view of blocks inside a TensorMap (for which adding new gradient is invalid) and the view of a gradient inside a standalone block (for which adding gradient is valid). 

I feel like making this difference should not be too hard, but could be left to a future PR.

PicoCentauri:
I agree that the second example would be also nice to have but could be done in a future PR. I don't see this as blocker.

python/src/equistore/operations/join.py
```diff
@@ -101,11 +101,19 @@ def join(tensor_maps: List[TensorMap], axis: str):
...
                 new_block.add_gradient(
                     parameter=parameter,
-                    data=gradient.data,
-                    samples=gradient.samples,
-                    components=gradient.components,
+                    gradient=TensorBlock(
+                        values=gradient.values,
+                        samples=gradient.samples,
+                        components=gradient.components,
+                        properties=gradient.properties,
```
PicoCentauri, line 115: I thought this is not possible with the current API?

PicoCentauri (changes requested):
I started reading the PR but I am a bit confused by the new architecture. I think it makes sense but I can not judge it right now.

@frostedoyster can you write a section to the docs? I would like to read the design first before I can continue here.

Luthaf:
> I started reading the PR but I am a bit confused by the new architecture. 

The core of it is that a TensorBlock is now literally 

```rust
struct TensorBlock {
    values: Array,
    samples: Labels,
    components: Vec<Labels>,
    properties: Labels,
    gradients: HashMap<String, TensorBlock>
}
```

The previous `BasicBlock` (source of some confusion) is now gone.

All the other changes are adapting to this, to add gradients to a block, retrieve gradients, and manipulate the corresponding data.

python/src/equistore/operations/join.py
frostedoyster: Actually could gradient=gradient work in the above?

frostedoyster:
Yes, sorry about the documentation. I'll try to update it today

python/src/equistore/operations/join.py
Luthaf: no, because `gradient` is already owned by another block. We could do `gradient=gradient.copy()`, but that's going to be slower. 

I'll add a comment on the code explaining that.

Luthaf: Actually, this code was slightly wrong, it should be using the same properties as the block, which might be different from the ones in `gradient.properties`.

Luthaf:
The current Rust API does not work for rascaline, since we are using the fact that samples and values are disjoint over there to borrow both of them in separate ways. I'll update it here

equistore-core/src/blocks.rs
```diff
@@ -112,106 +103,39 @@ fn check_component_labels(components: &[Arc<Labels>]) -> Result<(), Error> {
...
-            components: self.components.clone(),
-            properties: Arc::clone(&self.properties),
-        })
-    }
-}
-
-/// A single block in a `TensorMap`, containing both values & optionally
-/// gradients of these values w.r.t. any relevant quantity.
+/// A single block in a `TensorMap`, containing both data & optionally
+/// gradients of these data w.r.t. any relevant quantity.
```
PicoCentauri, line 107: Maybe say that another block can be **linked** as a gradient to this block.

equistore-core/src/io/load.rs
```diff
@@ -0,0 +1,168 @@
...
+    let mut archive = ZipArchive::new(reader).map_err(|e| ("<root>".into(), e))?;
+
+    let path = String::from("keys.npy");
+    let keys = read_npy_labels(archive.by_name(&path).map_err(|e| (path, e))?)?;
+
+    // FIXME before merging: should we add a fallback?
+    assert!(
+        archive.by_name("blocks/0/values/data.npy").is_err(),
+        "this is the old data format, TODO: fallback"
+    );
```
PicoCentauri, line 64: I think a legacy parser is useful. We can also remove the legacy parser quite quickly again.

PicoCentauri (commented):
Okay, I get it now and like this a lot! There are some todos already in the code which should be fixed at some point.

equistore-core/src/io/load.rs
Luthaf: I have a script to convert from old to new format using only numpy, so you don't even need equistore to convert the data. I'll push it ASAP

Luthaf: And this code checks if we have a file in the old format, and point users to the script

Luthaf:
I think I've handled all the TODO from this PR, so is good to go for me!

equistore-core/src/c_api/blocks.rs
```diff
@@ -218,14 +204,67 @@ pub unsafe extern fn eqs_block_labels(
...
+/// The gradient memory is still managed by the block, the returned
+/// `eqs_block_t*` should not be freed. The gradient pointer is invalidated if
+/// more gradients are added to the parent block, or if the parent block is
+/// freed with `eqs_block_free`.
 ///
-/// The `values_gradients` parameter controls whether this function looks up
-/// labels for `"values"` or one of the gradients in this block.
+/// @param block pointer to an existing block @param parameter the name of the
+/// gradient to be extracted @param gradient pointer to an empty `eqs_block_t`
+/// pointer that will be set to point to the requested gradient
```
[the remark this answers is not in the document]
Luthaf: Whoops, thanks!

equistore-core/include/equistore.hpp
```diff
@@ -1210,7 +1090,7 @@ class TensorBlock final {
 
     /// Get a view in the values in this block
     NDArray<double> values() {
```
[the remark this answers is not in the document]
Luthaf: No, we only support `double` to get data through the C API. #56 would allow using other types.

The Python API does not go through the same mechanism to get the arrays back, so it is able to pass other data types.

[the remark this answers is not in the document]
frostedoyster: Ok then cool

equistore-core/src/c_api/blocks.rs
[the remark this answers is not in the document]
frostedoyster: I see that this hasn't changed (?) Otherwise it's all good for me

PicoCentauri (approved):
Yes, I am happy with this simplification!

PR #254, opened 2023-04-13
Add infrastructure for conveniently manage checks

PicoCentauri:
Adding the functions `checks_enabled` and the classes/functions `enable_checks` and `disable_checks`. The latter two can be used two either set the check status globally or within a context manager. I also add a parser for an env variable `EQUISTORE_DISABLE_CHECKS`.


python/src/equistore/operations/unsafe.py
```diff
@@ -0,0 +1,136 @@
...
+
+.. autofunction:: equistore.checks_enabled
+
+.. autoclass:: equistore.enable_checks
+
+.. autoclass:: equistore.disable_checks
+
+"""
+
+from equistore.utils import _parse_bool_env_var
```
Luthaf, line 39: Since this is not used anywhere else, should we define this function in this module?

PicoCentauri: Yes, I don't mind where this lives for now. I just thought it might be useful in the future.

python/src/equistore/operations/checks.py
```diff
@@ -0,0 +1,133 @@
...
+Equistore performs `shape` as well as `consistency` checks for the metadata during each
+operation. The check status can be verified with :py:func:`equistore.checks_enabled()`
+and controlled by :py:class:`equistore.unsafe_enable_checks()` and
+:py:class:`equistore.unsafe_disable_checks()`.
+
+.. warning::
+
+    Using equistore without checks is unsafe and can lead to unwanted results and may
+    produce cryptic error messages. Therefore, disabling checks are recommended for
+    nobody and should only be applied by advanced users!
```
Luthaf, line 11:
```suggestion
.. warning::

    Using equistore without checks is unsafe, can lead to unwanted results, and may
    produce cryptic error messages. Therefore, disabling checks is not recommended
    and should only be used by advanced users! If you see strange results after disabling 
    checks, try running the code again with checks enabled.
```

PicoCentauri:
Thanks for the comments. If the CI passes we can merge and move on.

PR #255, opened 2023-04-13
Provide kwargs for `Labels.arange()`

frostedoyster:


python/src/equistore/labels.py
```diff
@@ -12,46 +12,35 @@
 class Labels(np.ndarray):
     """
     A set of labels used to carry metadata associated with a tensor map.
-
```
Luthaf, line 15: Why remove these lines?

Luthaf (approved):

PR #256, opened 2023-04-14
Split the Python package in three

### LINKED ISSUE #108: Translate setup.cfg into pyproject.toml
Once we dropped support for Python 3.6 and 3.7 we can move everything from the setup.cfg into the pyproject.toml as suggested in #88.

Luthaf:
~This builds on top of #253~

Fixes #108 (it seems that we can do this in Python 3.7)

---

There is now an `equistore-core` package which can have a very stable version number; an `equistore-operations` package which can evolve fast and break everything and an `equistore` package which re-exports functions and classes from the previous two.

----
:books: Documentation preview :books:: https://equistore--256.org.readthedocs.build/en/256/


tox.ini
```diff
@@ -3,17 +3,23 @@ min_version = 4.0
...
 
 [testenv]
 package = external
-package_env = build-equistore
+package_env = build-equistore-core
 lint_folders = "{toxinidir}/python" "{toxinidir}/setup.py"
 
+commands =
+    # error if the user gives a wrong testenv name in `tox -e`
+    python -c "import sys; print('environement {env_name} does not exist'); sys.exit(1)"
```
PicoCentauri, line 19: ohh yes this is very useful!

python/equistore/equistore/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+    HAS_EQUISTORE_OPERATIONS = True
+except ImportError:
+    HAS_EQUISTORE_OPERATIONS = False
+
+
+from .core import *  # noqa
+
+if HAS_EQUISTORE_OPERATIONS:
+    from .operations import *  # noqa
```
PicoCentauri, line 30: Why do we need a double import here?

python/equistore/equistore/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+from . import core  # noqa
+
+try:
+    from . import operations  # noqa
+
+    HAS_EQUISTORE_OPERATIONS = True
+except ImportError:
+    HAS_EQUISTORE_OPERATIONS = False
```
PicoCentauri, line 24: I thought equistore is a meta package only packaging the core and opearations together. Why aren't the only lines here:

```python
import equistore-core as core
import equistore-operations as operations

from equistore-core import TensorMap, TensorBlock, Labls
from equistore-operations import *
```

PicoCentauri (changes requested):
I like the split. This will simplify the development for us and due to the meta package the users will not notice anything. I have some questions though.

python/equistore/equistore/__init__.py
Luthaf: Because equistore-core and equistore-operations are the names of the **distributions** in PEP language, and correspond to namespace **packages** installing their code in `<site-packages>/equistore/core` and `<site-packages>/equistore/operations`.

The weird imports is more about separating the import of sub-packages from the changes to the equistore namespace, but this could be changed.

.github/workflows/build-wheels.yml
```diff
@@ -99,10 +119,19 @@ jobs:
...
+          python setup.py sdist --dist-dir ../../dist/
+      - name: build equistore-operations sdist
+        run: |
+          cd python/equistore-operations
+          python setup.py sdist --dist-dir ../../dist/
+      - name: build equistore sdist
         run: |
-          python -m pip install wheel
-          python setup.py sdist
+          python setup.py sdist --dist-dir ./dist/
```
PicoCentauri, line 134: same as above

.github/workflows/build-wheels.yml
```diff
@@ -114,3 +143,36 @@ jobs:
...
+
+      - name: install dependencies
+        run: python -m pip install wheel
+      - name: build equistore-operations wheel
+        run: |
+          cd python/equistore-operations
+          python setup.py bdist_wheel --dist-dir ../../dist/
+      - name: build equistore wheel
+        run: |
+          python setup.py bdist_wheel --dist-dir ./dist/
```
PicoCentauri, line 167: same as above

python/equistore-core/README.rst
```diff
@@ -0,0 +1,6 @@
+equistore-core
+==============
+
+This package contains the Python bindings to the core C API of equistore. Most
+users will want to use the ``equistore`` package instead, which also includes
```
PicoCentauri, line 5:
```suggestion
users will want to use the ``equistore`` meta package instead, which also includes
```

python/equistore-core/equistore/core/_c_lib.py
```diff
@@ -1,16 +1,26 @@
...
 
 
 _HERE = os.path.realpath(os.path.dirname(__file__))
 
+Version = namedtuple("Version", ["major", "minor", "patch"])
+
+
+def parse_version(version):
+    match = re.match(r"(\d+)\.(\d+)\.(\d+).*", version)
+    if match:
```
PicoCentauri, line 19: Only one version drop until we can do 🥳

```python
    if match := re.match(r"(\d+)\.(\d+)\.(\d+).*", version)
```

pyproject.toml
```diff
@@ -1,11 +1,55 @@
...
-    "setuptools >=44",
-    "wheel >=0.36",
-    "cmake",
+    "setuptools >=61",
+    "wheel >=0.38",
 ]
 build-backend = "setuptools.build_meta"
 
+[tool.setuptools]
+zip-safe = true
```
PicoCentauri, line 44: I think this is the default, but we can leave it if you like.

python/equistore-operations/setup.py
```diff
@@ -0,0 +1,120 @@
+import os
```
PicoCentauri, line 1: Cant't we use code from the setup.py of equistore-core? Since operations should depend on the core it should be there.

.github/workflows/build-wheels.yml
```diff
@@ -32,22 +32,23 @@ jobs:
...
+      - name: install dependencies
+        run: python -m pip install wheel
+      - name: build equistore-core wheel
         env:
           RUST_BUILD_TARGET: ${{ matrix.rust-target }}
         run: |
-          python -m pip install wheel
-          python setup.py bdist_wheel --plat-name ${{ matrix.platform-name }}
+          cd python/equistore-core
+          python setup.py bdist_wheel --plat-name ${{ matrix.platform-name }} --dist-dir ../../dist/
```
PicoCentauri, line 51: and why are we not using tox for building the binary here. Seems like duplicated.

PicoCentauri (commented):
I mainly have a comment if we should not use `build` instead of `python setup.py`, everything else is looking good!

pyproject.toml
Luthaf: yes, I like that this is explicit here (since we also have cases with zip-safe=false)

python/equistore-operations/setup.py
Luthaf: done!

Luthaf:
Wheels and sdist from the CI looks alright, most of them are able to use the `build` module, except for the one where we manually cross-compile to M1/arm64 for macOS.

.github/workflows/build-wheels.yml
```diff
@@ -99,18 +110,28 @@ jobs:
...
-      - name: build sdist
-        run: |
-          python -m pip install wheel
-          python setup.py sdist
+      - name: install dependencies
+        run: python -m pip install wheel build
+      - name: package vendored code
+        run: ./scripts/package-core.sh
+      - name: build equistore-core sdist
+        run: python -m build python/equistore-core --sdist --outdir=dist/
```
PicoCentauri, line 118: super nice! I like it

.github/workflows/tests.yml
```diff
@@ -104,56 +102,71 @@ jobs:
...
   python-build:
     runs-on: ubuntu-20.04
     name: check Python build
-    strategy:
-      matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
     steps:
       - uses: actions/checkout@v3
 
       - name: set up Python ${{ matrix.python-version }}
```
PicoCentauri, line 128: `${{ matrix.python-version }} `does not exist anymore.

docs/src/devdoc/architecture.rst
```diff
@@ -3,15 +3,20 @@
 Code organization
 =================
 
-The code is organized in three main modules, each in a separate directory:
+The code is organized in multiple modules, each in a separate directory:
```
PicoCentauri, line 6: I see what you are planning here 😝

python/equistore-core/pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+python_files = *.py
+testpaths = tests
```
PicoCentauri, line 3: Why not putting this into the `pyproject.toml`?

python/equistore-operations/pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+python_files = *.py
+testpaths = tests
```
PicoCentauri, line 3: same here as in the core.

PicoCentauri (changes requested):
Almost there. One question and a slight change in a title.

docs/src/devdoc/architecture.rst
Luthaf: 😁

python/equistore-core/pytest.ini
Luthaf: I can try, I think there was some issue with pytest rootdir without this file.

Luthaf: no, it works!

.github/workflows/tests.yml
Luthaf: what would you think of testing only on the min & max supported Python version? Min to ensure we don't use too modern things, max to ensure we don't use deprecated things

PicoCentauri: Sounds reasonable. I am just wondering why all other packages test the full chain.

Luthaf: what do you mean by all other? All other cosmo packages? Then the reason is we added all version because it was easy and did not though too much about it ^^

DavideTisi (commented):
In my opinion the pr is fine, but i am not the best one to review this. Many of the things concerning the building i learn how to do it reading the PR ahahah, but the idea is super nice

PR #258, opened 2023-04-18
Add public `equal_metadata_block` function

### LINKED ISSUE #257: Add a public `equal_metadata_block()` function
After discussion with @DivyaSuman14 its clear that there are use cases (i.e. Hamiltonian learning) where one block in a descriptor TensorMap is used to learn against multiple blocks in a target properties TensorMap. In such a case, the keys of the input and output TensorMaps will not be the same, but the metadata (at least, along samples and components) of the pairs of blocks that are learned together should still be the same.

At the moment, only `equal_metadata` function that acts on TensorMaps is publicly available. There should be a public-facing `equal_metadata_block` function, that essentially wraps the private `_check_blocks` function that is already there.

jwa7:
Closes #257 

----
:books: Documentation preview :books:: https://equistore--258.org.readthedocs.build/en/258/

### DIFF
--- a/python/src/equistore/operations/equal_metadata.py
+++ b/python/src/equistore/operations/equal_metadata.py
@@ -5,6 +5,8 @@ from typing import List, Optional
 
 import numpy as np
 
+import equistore
+
 from ..block import TensorBlock
 from ..labels import Labels
 from ..tensor import TensorMap
@@ -107,7 +109,7 @@ def equal_metadata(
     # Check equivalence in keys
     try:
         _check_same_keys(tensor_1, tensor_2, "equal_metadata")
-    except ValueError:
+    except equistore.NotEqualError:
         return False
 
     # Loop over the blocks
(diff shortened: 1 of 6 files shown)


python/src/equistore/operations/equal_metadata.py
```diff
@@ -129,6 +129,86 @@ def equal_metadata(
...
 
 
+def equal_metadata_block(
+    block_1: TensorBlock, block_2: TensorBlock, check: Optional[List] = None
+) -> bool:
+    """
+    Checks if two :py:class:`TensorBlock` objects have the same metadata,
+    returning a bool.
+
+    If `check` is none (the default), all metadata (i.e. the samples,
```
Luthaf, line 139:
```suggestion
    If ``check`` is :py:obj:`None` (the default), all metadata (i.e. the samples,
```

All the check should have two backticks

python/src/equistore/operations/equal_metadata.py
```diff
@@ -129,6 +129,86 @@ def equal_metadata(
...
+    # Check metadata of the blocks
+    try:
+        _check_blocks(block_1, block_2, check, "equal_metadata")
+    except ValueError:
+        return False
+
+    # Check metadata of the gradients
+    try:
+        _check_same_gradients(block_1, block_2, check, "equal_metadata")
+    except ValueError:
```
Luthaf, line 207: I don't love that we still use ValueError for the `_check` functions, since that could hide issues in the same way as #175. Should we also use `NotEqualError` for these?

jwa7:
Thanks for the review @Luthaf ! I've made the requested changes, using `NotEqualError`s in the `_check*` functions. Let me know if they're ok!

Luthaf (approved):

Luthaf:
Do you want to squash the first two commits, or should I squash-merge the whole PR?

PR #259, opened 2023-04-19
When converting between npz formats, save files uncompressed

Luthaf:
This is a follow-up to #253. I used `savez_compressed` there to update the tests data, but this should not be the default since equistore-core does not know how to read such files.

----
:books: Documentation preview :books:: https://equistore--259.org.readthedocs.build/en/259/


agoscinski:
So ZipArchive on the equistore-core side is not using any compression?

agoscinski (approved):
I see, thanks.

PR #260, opened 2023-04-24
Fix CMakeLists.txt for Ninja generator

Luthaf:
We need to declare that cargo-build-equistore is the command to use to create static/shared equistore library, or ninja gets confused.

----
:books: Documentation preview :books:: https://equistore--260.org.readthedocs.build/en/260/


agoscinski (approved):
Looks fine now.

PR #262, opened 2023-04-25
Fix all bugs

### LINKED ISSUE #65: Unhelpful error message when trying to write a ``TensorBlock`` to file
When trying to write an ``TensorBlock`` object to file using the ``equistore.io.save()`` function, the following error message is given:
```py
ArgumentError: argument 2: <class 'TypeError'>: expected LP_eqs_tensormap_t instance instead of LP_eqs_block_t
```

**Example**
```py
import numpy as np

from equistore import io, TensorBlock

# Create an example TensorBlock
block = TensorBlock(values=np.random.random((2, 2)),
                    samples=Labels(names=('structure','center'), values=np.array([[0, 1], [0, 2]])),
                    components=[],
                    properties=Labels(names=('n',), values=np.array([(0,), (1,),])),
                   )

# Attempt to save to file
io.save('tensorblock.npz', block)
>>> ArgumentError: argument 2: <class 'TypeError'>: expected LP_eqs_tensormap_t instance instead of LP_eqs_block_t
```

**Desired Behaviour**

The error message could be more helpful by explaining more explicitly / less cryptically that only ``TensorMap`` objects, and not ``TensorBlock`` objects, can be saved.

### LINKED ISSUE #174: Accessing blocks in empty tensor map
Hey! I think this is a bug. I am able to create an empty tensor map, but then I am not allowed to access the blocks as the code raises a ValueError `ValueError: Couldn't find any block matching the selection {}`

Snippet to reproduce the bug

```py
empty_tensor = TensorMap(keys=Labels.empty("key"), blocks=[])
empty_tensor.blocks()
```

### LINKED ISSUE #218: Make gradient blocks with samples that are not contained in the values invalid
This code works, even though sample 1 (referenced in the gradient block) is not in the values of the tensor block.
```python
block_1 = TensorBlock(
    values=np.full((1, 1), 1.0),
    samples=Labels(["samples"], np.array([[0]], dtype=np.int32)),
    components=[],
    properties=Labels(["properties"], np.array([[0]], dtype=np.int32)),
)
block_1.add_gradient(
    "parameter",
    samples=Labels(
        ["sample"], np.array([[1]], dtype=np.int32)
    ),
    data=np.full((1, 1), 11.0),
    components=[],
)
```

Code that needs to be modified 
https://github.com/lab-cosmo/equistore/blob/b8a65fa205ffc0542b0f34a58ace607e883a3271/equistore-core/src/blocks.rs#L304-L309

Luthaf:
Fix #65
Fix #174 
Fix #218 

----
:books: Documentation preview :books:: https://equistore--262.org.readthedocs.build/en/262/

### DIFF
--- a/python/equistore-core/equistore/core/labels.py
+++ b/python/equistore-core/equistore/core/labels.py
@@ -1,6 +1,6 @@
 import ctypes
 from collections import namedtuple
-from typing import List, Optional, Union
+from typing import Optional, Sequence, Tuple, Union
 
 import numpy as np
 
@@ -57,7 +57,7 @@ class Labels(np.ndarray):
             ...
     """
 
-    def __new__(cls, names: Union[List[str], str], values: np.ndarray, **kwargs):
+    def __new__(cls, names: Union[Sequence[str], str], values: np.ndarray, **kwargs):
         """
         :param names: names of the dimensions in the new labels, in the case of a single
                       name also a single string can be given: ``names = "name"``
@@ -141,7 +141,7 @@ class Labels(np.ndarray):
             self._lib.eqs_labels_free(self._eqs_labels_t)
 
     @property
-    def names(self) -> List[str]:
+    def names(self) -> Tuple[str]:
         """Names of the columns/dimensions used for these labels"""
         return self.dtype.names or []
 
--- a/python/equistore-core/equistore/core/labels.py
+++ b/python/equistore-core/equistore/core/labels.py
@@ -73,7 +73,7 @@ class Labels(np.ndarray):
             if len(names) == 0:
                 names = tuple()
             else:
-                names = tuple(names)
+                names = (names,)
         else:
             names = tuple(str(n) for n in names)
 
--- a/python/equistore-core/tests/tensor.py
+++ b/python/equistore-core/tests/tensor.py
@@ -5,6 +5,8 @@ import numpy as np
 import pytest
 from numpy.testing import assert_equal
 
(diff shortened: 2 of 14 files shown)


python/equistore-core/equistore/core/labels.py
```diff
@@ -57,7 +57,7 @@ class Labels(np.ndarray):
             ...
     """
 
-    def __new__(cls, names: Union[List[str], str], values: np.ndarray, **kwargs):
+    def __new__(cls, names: Union[Sequence[str], str], values: np.ndarray, **kwargs):
```
jwa7, line 60: Why the change from List to Sequence here?

python/equistore-core/tests/tensor.py
```diff
@@ -361,3 +363,35 @@ def test_components_to_properties(tensor):
...
+    message = "invalid parameter: 'not_a_key' is not part of the keys for this tensor"
+    with pytest.raises(EquistoreError, match=message):
+        empty_tensor.blocks_matching(not_a_key=3)
+
+    message = (
+        "invalid parameter: block index out of bounds: we have "
+        "0 blocks but the index is 3"
+    )
+    with pytest.raises(EquistoreError, match=message):
+        empty_tensor.block(3)
```
jwa7, line 390: Would be good to have a test on `blocks()` that also raises an error

Luthaf: done!

PR #263, opened 2023-04-27
Initial TorchScript version of the core classes

Luthaf:
This is still missing a bunch of magic functions (`__getitem__`, …) and other convenience which exist on equistore-core, but should be at a good starting point for review!

----
:books: Documentation preview :books:: https://equistore--263.org.readthedocs.build/en/263/

### DIFF
--- a/docs/src/reference/torch/index.rst
+++ b/docs/src/reference/torch/index.rst
@@ -5,6 +5,33 @@ TorchScript API reference
 
 .. py:currentmodule:: equistore.torch
 
+We provide a special PyTorch C++ extension exporting all of the core equistore
+types in a way compatible with the TorchScript compiler, allowing users to save
+and load models based on equistore everywhere TorchScript is supported. In
+particular, this allow to define, train and save a model from Python, and then
+load it with pure C++ code, without requiring a Python interpreter. Please refer
+to the :ref:`installation instructions <install-torch-script-lib>` to know how
+to install the Python and C++ sides of this library.
+
+The classes and functions in the TorchScript API are kept as close as possible
+to the classes and functions of the pure Python API, with the explicit goal that
+changing from
+
+.. code-block:: python
+
+    import equistore
+    from equistore import TensorMap, TensorBlock, Labels
+
+to
+
+.. code-block:: python
+
+    import equistore.torch as equistore
+    from equistore.torch import TensorMap, TensorBlock, Labels
+
+should be 80% of the work required to make a model developed in Python with
+:py:mod:`equistore.operations` compatible with TorchScript.
+
 The documentation for the three usual core classes of equistore can be found in
 the following pages:
 
@@ -27,3 +54,38 @@ in the following documentation:
     :maxdepth: 1
 
     cxx/index
+
+.. _python-vs-torch:
+
+Differences between Python and TorchScript API
+----------------------------------------------
+
+The Python and TorchScript interfaces to equistore are very similar, and have
+some level of overlap (it is possible to store data in torch ``Tensor`` with the
+Python API), so a big question is when/why should you use one or the other, and
+what's the difference between them.
+
+First, the Python and TorchScript API are separate to allow using the Python API
+without torch, for example in a pure numpy workflow, or with other array types
+such as `Jax`_ arrays, `cupy`_ arrays, *etc.*.
+
+While this works great for a lot of use cases based on torch (defining models,
+training them with autograd, …), the TorchScript compiler is more restrictive in
+what code it accepts and does not support calling into arbitrary native code
+like the Python API does. The TorchScript version of equistore is here to fill
+this gap, teaching the TorchScript compiler what's going on inside equistore and
+how to translate code using equistore to TorchScript.
+
+Another different is that while the Python API supports storing data in multiple
+ways, including storing it on GPU through torch Tensor, the metadata is always
+stored on CPU, inside the equistore shared library. The TorchScript API changes
+this to enable storing both the data and metadata on GPU, minimizing data
+transfer and making the models faster.
+
+If you only care about PyTorch, we would recommend using the TorchScript API
+from the start, to make sure you will be able to export your code to
+TorchScript. If you are not using PyTorch, or if you want to write code in an
+engine agnostic way, we recommend using the Python API.
+
+.. _Jax: https://jax.readthedocs.io/en/latest/
+.. _cupy: https://cupy.dev
--- a/equistore-core/include/equistore.hpp
+++ b/equistore-core/include/equistore.hpp
@@ -26,6 +26,7 @@
 namespace equistore_torch {
     class LabelsHolder;
     class TensorBlockHolder;
+    class TensorMapHolder;
 }
 
 namespace equistore {
@@ -408,7 +409,6 @@ public:
     Labels(const std::vector<std::string>& names, std::vector<std::initializer_list<int32_t>> values):
         Labels(details::labels_from_cxx(names, NDArray(std::move(values), names.size()))) {}
 
-
     /// Create an empty set of Labels with the given names
     Labels(const std::vector<std::string>& names):
         Labels(details::labels_from_cxx(
@@ -523,11 +523,15 @@ private:
         }
     }
 
+    Labels(const std::vector<std::string>& names, const int32_t* values, size_t count):
+        Labels(details::labels_from_cxx(names, NDArray(values, {count, names.size()}))) {}
+
     friend Labels details::labels_from_cxx(const std::vector<std::string>&, NDArray<int32_t>);
     friend class TensorMap;
     friend class TensorBlock;
 
     friend class equistore_torch::LabelsHolder;
+    friend class equistore_torch::TensorMapHolder;
 
     std::vector<const char*> names_;
     eqs_labels_t labels_;
--- a/equistore-core/include/equistore.hpp
+++ b/equistore-core/include/equistore.hpp
@@ -25,6 +25,7 @@
 
 namespace equistore_torch {
     class LabelsHolder;
+    class TensorBlockHolder;
 }
 
 namespace equistore {
@@ -1205,7 +1206,7 @@ public:
     ///
     /// @param parameter check for gradients with respect to this `parameter`
     ///                  (e.g. `"positions"`, `"cell"`, ...)
-    TensorBlock gradient(std::string parameter) const {
+    TensorBlock gradient(const std::string& parameter) const {
         eqs_block_t* gradient_block = nullptr;
         details::check_status(
             eqs_block_gradient(block_, parameter.c_str(), &gradient_block)
@@ -1318,6 +1319,7 @@ private:
     }
 
     friend class TensorMap;
+    friend class equistore_torch::TensorBlockHolder;
 
     eqs_block_t* block_;
     bool is_view_;
--- a/equistore-core/include/equistore.hpp
+++ b/equistore-core/include/equistore.hpp
@@ -1240,7 +1240,7 @@ public:
     eqs_block_t* release() {
          if (is_view_) {
             throw Error(
-                "can not call non-const TensorBlock::release on this "
+                "can not call TensorBlock::release on this "
                 "block since it is a view inside a TensorMap"
             );
         }
@@ -1318,7 +1318,6 @@ private:
         return array;
     }
 
-    friend class TensorMap;
     friend class equistore_torch::TensorBlockHolder;
 
     eqs_block_t* block_;
@@ -1347,10 +1346,9 @@ public:
     TensorMap(Labels keys, std::vector<TensorBlock> blocks) {
         auto c_blocks = std::vector<eqs_block_t*>();
         for (auto& block: blocks) {
-            c_blocks.push_back(block.as_eqs_block_t());
-            // We will move the data inside the new map, let's put the
-            // TensorBLock in its moved-from state now
-            block.block_ = nullptr;
+            // We will move the data inside the new map, let's release the
+            // pointers out of the TensorBlock now
+            c_blocks.push_back(block.release());
         }
 
         tensor_ = eqs_tensormap(
@@ -1518,13 +1516,13 @@ public:
         return TensorMap(ptr);
     }
 
-    /// This function calls `keys_to_properties` with an empty set of `Labels`
+    /// This function calls `keys_to_samples` with an empty set of `Labels`
     /// with the dimensions defined in `keys_to_move`
     TensorMap keys_to_samples(const std::vector<std::string>& keys_to_move, bool sort_samples = true) const {
         return keys_to_samples(Labels(keys_to_move), sort_samples);
     }
 
-    /// This function calls `keys_to_properties` with an empty set of `Labels`
+    /// This function calls `keys_to_samples` with an empty set of `Labels`
     /// with a single dimension: `key_to_move`
     TensorMap keys_to_samples(const std::string& key_to_move, bool sort_samples = true) const {
         return keys_to_samples(std::vector<std::string>{key_to_move}, sort_samples);
@@ -1535,7 +1533,7 @@ public:
     ///
     /// @param dimensions name of the component dimensions to move to the
     ///                  properties
-    TensorMap  components_to_properties(const std::vector<std::string>& dimensions) const {
+    TensorMap components_to_properties(const std::vector<std::string>& dimensions) const {
         auto c_dimensions = std::vector<const char*>();
         for (const auto& v: dimensions) {
             c_dimensions.push_back(v.c_str());
--- a/equistore-core/include/equistore.hpp
+++ b/equistore-core/include/equistore.hpp
@@ -23,6 +23,10 @@
 /// convenient, but also allow to drop back to the C API if required, by
 /// providing functions to extract the C API handles (named `as_eqs_XXX`).
 
+namespace equistore_torch {
+    class LabelsHolder;
+}
+
 namespace equistore {
 class Labels;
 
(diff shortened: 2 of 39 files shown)


equistore-core/include/equistore.hpp
```diff
@@ -1294,7 +1318,7 @@ class TensorBlock final {
         return array;
     }
 
-    friend class TensorMap;
+    friend class equistore_torch::TensorBlockHolder;
```
agoscinski, line 1321: I don't understand the change here. Is this correct?

* doesn't this require to have equistore-torch also installed?
* why is the friend class TensorMap? Isn't TensorMap still needed as friend class?

equistore-core/tests/cpp/CMakeLists.txt
```diff
@@ -1,9 +1,10 @@
 cmake_minimum_required(VERSION 3.10)
 project(equistore-tests)
 
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_NO_SYSTEM_FROM_IMPORTED ON)
+if (${CMAKE_VERSION} VERSION_EQUAL 3.10)
```
agoscinski, line 4: why is this needed? What if I have a higher cmake version is used? Don't you not need to set the CXX standard?

equistore-core/tests/c-api-install.rs
```diff
@@ -45,22 +41,19 @@ fn check_c_api_build_install() {
...
+    assert!(status.success(), "failed to run test project cmake configuration");
+
+    // build the code, linking to equistore
+    let mut cmake_build = utils::cmake_build(&build_dir);
+    let status = cmake_build.status().expect("could not run cmake");
+    assert!(status.success(), "failed to run test project cmake build");
+
+    // run the executables
+    let mut ctest = utils::ctest(&build_dir);
+    let status = ctest.status().expect("could not run ctest");
```
agoscinski, line 57: note for me: arguments have been moved to utils/mod.rs

scripts/package-torch.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
```
agoscinski, line 2: Maybe some doc what this script does?

equistore-torch/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
...
+    endif()
+endif()
+
+find_package(equistore 0.1 CONFIG REQUIRED)
+
+# FindCUDNN.cmake distributed with PyTorch is a bit broken, so we have a
+# fixed version in `cmake/FindCUDNN.cmake`
+set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake;${CMAKE_MODULE_PATH}")
+
+find_package(Torch 1.11 REQUIRED)
```
agoscinski, line 36: here you require 1.11 and in the toml you require 1.13

agoscinski (commented):
What is the idea for people to install equistore-torch later on? Using a extra_require?

.github/workflows/python-tests.yml
```diff
@@ -0,0 +1,59 @@
+name: Python tests
```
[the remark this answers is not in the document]
Luthaf: yes, I'm slowly rebasing all the branches as more of the required changes get in master

equistore-core/tests/cpp/CMakeLists.txt
Luthaf: We now set the C++ standard using target properties (which is only available since cmake 3.11). See `target_compile_features(catch PUBLIC cxx_std_11)` and similar for all targets

scripts/package-torch.sh
Luthaf: 👍

equistore-torch/CMakeLists.txt
Luthaf: 👍

equistore-core/tests/cpp/CMakeLists.txt
Luthaf: Sorry, I though I was answering to a different PR ^^ This commit is actually part of #233, and will be replaced by the version in #267 once we merge it.

scripts/package-torch.sh
Luthaf: Done in #233

equistore-torch/CMakeLists.txt
Luthaf: Done in #233

equistore-torch/src/labels.cpp
```diff
@@ -51,32 +51,35 @@ static torch::Tensor initializer_list_to_tensor(
     return tensor.reshape({static_cast<int64_t>(count), static_cast<int64_t>(size)});
 }
 
-static std::vector<std::string> normalize_names(torch::IValue names) {
+std::vector<std::string> equistore_torch::details::normalize_names(torch::IValue names, std::string context) {
```
agoscinski, line 54: Will there be doc for this? To what do the names need to be normalized? First thought the context has to do something with the context manager of torch, but it is just for error context. Maybe we can also name it this way (e.g. `error_msg_prefix`)?

equistore-torch/include/equistore/torch/labels.hpp
```diff
@@ -76,24 +89,51 @@ class EQUISTORE_TORCH_EXPORT LabelsHolder: public torch::CustomClassHolder {
...
+        return !labels_.has_value();
     }
+
+    /// Transform a view of Labels into owned Labels, which can be further given
+    /// to equistore functions. This does nothing if the Labels are already
+    /// owned.
+    LabelsHolder to_owned() const;
+
 private:
+    struct CreateView {};
```
agoscinski, line 123: what is this struct `CreateView` doing? Could not find any usage of this

equistore-torch/src/labels.cpp
```diff
@@ -158,60 +159,272 @@ TorchLabels LabelsHolder::create(
...
+    if (!labels_.has_value()) {
+        C10_THROW_ERROR(ValueError,
+            "can not call this function on Labels view, call to_owned first"
+        );
+    }
+
+    return labels_.value();
+}
+
+LabelsHolder LabelsHolder::to_owned() const {
```
agoscinski, line 208: Not exactly sure what this does. So `has_value` checks if is a "None" type (part from Optional type)? If it is not "None", then return itself. Otherwise it creates an LabelHolder with the names and values. I am not sure when this else case is used.

equistore-torch/include/equistore/torch/labels.hpp
Luthaf: leftover from development! I'll remove it

equistore-torch/src/labels.cpp
Luthaf: If `LabelsHolder::labels_` is an `equistore::Labels`, we return that. If `LabelsHolder::labels_` is a nullopt (i.e. None), then we create a new equistore::Labels, and a new LabelsHolder wrapping it. 

`LabelsHolder::labels_` will be a nullopt when we create a column view for some Labels, with something like `labels[["center", "structure"]]`

I'll add some more explanations around this

equistore-torch/include/equistore/torch/labels.hpp
Luthaf: Actually, this is used in the constructor just below, to differentiate it from the main one. I've added a comment about that.

Luthaf:
This should be ready for review and merge! For anyone wanting to give this a look, this PR is better reviewed one commit at the time.

agoscinski:
Hello Guillaume,

Thanks for the work! I tried it out with the spherical expansion code from @frostedoyster and it worked after some changes.

The only problem I had with the core package is that the `__iter__` function is not available which is convenient to use. Do I understand correct that to include the `__iter__` function, we would need to do another change like this one https://github.com/pytorch/pytorch/pull/100724/ ?
In this case it is probably is not even possible because the output type is generic? I am also confused about where this `items()` function is implemented. I could only find the declaration.

Another point the `__repr__` behavior is different
```python
In [6]: tm_ref # equistore
Out[6]: 
TensorMap with 6 blocks
keys: ['a_i' 'lam' 'sigma']
         1     0      1
         6     0      1
         8     0      1
         1     1      1
         6     1      1
         8     1      1
```

```python
In [125]: tm_ref #equistore.torch
Out[125]: <torch.ScriptObject at 0x7f47cd423e70>
```
---
I am not sure if this is something that is part of this PR, but I wanted to mention it in case it is: When I tried to add equistore-torch and equistore-operations as setup requirements,
https://github.com/agoscinski/torch_spex/blob/cdac1549948250cd3ff407465bdc7478882d16e8/setup.py#L19-L32
I get a nonsensical dependency conflict on the same version.
```
ERROR: Cannot install equistore-operations==0.1.0.dev286 and equistore-torch==0.1.0.dev286 because these package versions have conflicting dependencies.
ERROR: Resolution Impossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
The conflict is caused by:
    equistore-operations 0.1.0.dev286 depends on equistore-core 0.1.0.dev286 (from /tmp/pip-install-6b03c972/equistore-operations_e14fe4088fca448786f5b9b3654f02b8/python/equistore-core)
    equistore-torch 0.1.0.dev286 depends on equistore-core 0.1.0.dev286 (from /tmp/pip-install-6b03c972/equistore-torch_a4828d7d7a304d0eb6b2b79c1aceaca7/python/equistore-core)
```
https://github.com/agoscinski/torch_spex/actions/runs/5171050634/jobs/9314351880

In the end I made it work by installing them separately in the tox file, and it worked fine.

Luthaf:
> The only problem I had with the core package is that the __iter__ function is not available which is convenient to use. Do I understand correct that to include the __iter__ function, we would need to do another change like this one https://github.com/pytorch/pytorch/pull/100724 ?

No, we need much more than a simple change in torch: the `__iter__` function is returning a generator (i.e. it uses `yield` statements), but these are not supported by TorchScript, so we would have to add a significant feature to TorchScript. The plan is to replace `for key, block in tensor` with a more explicit `for key, block in tensor.items()`.

> Another point the __repr__ behavior is different

https://github.com/pytorch/pytorch/pull/100724 fixes this, but will only be available in torch 2.1

> I am not sure if this is something that is part of this PR, but I wanted to mention it in case it is: When I tried to add equistore-torch and equistore-operations as setup requirements,
https://github.com/agoscinski/torch_spex/blob/cdac1549948250cd3ff407465bdc7478882d16e8/setup.py#L19-L32
I get a nonsensical dependency conflict on the same version.

This is a known issue, but it is getting very annoying. I'll check if there is a solution, but I think the only way out is to make a release on PyPI

agoscinski:
Thanks for the answers.

I encountered a problem when scripting LabelsView which you get when accessing a Labels with square brackets `__getitem__`
```python
import torch
from equistore.torch import Labels

@torch.jit.script
def foo():
    samples=Labels(
        names=["s_1", "s_2"], values=torch.IntTensor([[0, 0], [1, 1], [2, 2]])
    )
    return samples["s_1"].values
```
Error
```
  File "/home/alexgo/code/torch_spex/labelsview-scripting.py", line 5, in <module>
    def foo():
  File "/opt/anaconda3/envs/equitorch/lib/python3.10/site-packages/torch/jit/_script.py", line 1341, in script
    fn = torch._C._jit_script_compile(
RuntimeError: 
'Any' object has no attribute or method 'values'.:
  File "/home/alexgo/code/torch_spex/labelsview-scripting.py", line 9
        names=["s_1", "s_2"], values=torch.IntTensor([[0, 0], [1, 1], [2, 2]])
    )
    return samples["s_1"].values
           ~~~~~~~~~~~~~~~~~~~~ <--- HERE

```
This works
```python
import torch
from equistore.torch import Labels

@torch.jit.script
def foo():
    samples=Labels(
        names=["s_1", "s_2"], values=torch.IntTensor([[0, 0], [1, 1], [2, 2]])
    )
    return samples.values[:, samples.names.index("s_1")]
```

Luthaf:
Yes, this is a known limitation: https://github.com/lab-cosmo/equistore/blob/f995ba2c479941cc4feff19f666ea73c83679a89/python/equistore-torch/equistore/torch/documentation.py#L251-L254

You can either use the `view()` function explicitly

```py
samples.view("s_1").values
```

or help the compiler refine types:

```py
view = samples["s_1"]
assert isinstance(view, Labels)
```

equistore-torch/src/array.cpp
```diff
@@ -1,44 +1,95 @@
...
-#include <torch/torch.h>
+#include <torch/script.h>
 
 #include <equistore.hpp>
 
 #include "equistore/torch/array.hpp"
 
 using namespace equistore_torch;
 
+eqs_data_origin_t equistore_torch::TORCH_DATA_ORIGIN = 0;
```
agoscinski, line 12: Can this get a bit more explanation? This is not clear to me how this constants decided?

equistore-torch/src/array.cpp
```diff
@@ -1,44 +1,95 @@
...
+
+// eqs_data_origin registration in a thread-safe way through C++11 static
+// initialization of a class with a constructor
+struct EqsOriginRegistration {
```
agoscinski, line 16: Can this get a bit more explanation? I don't understand why this thread-safe.

equistore-torch/src/tensor.cpp
```diff
@@ -1,19 +1,432 @@
...
+        return {};
+    }
+
+    auto block = this->block_by_id(0);
+    auto n_dimensions = block->values().sizes().size();
+
+    return labels_names(block->as_equistore(), n_dimensions - 1);
+}
+
+std::vector<std::tuple<TorchLabelsEntry, TorchTensorBlock>> TensorMapHolder::items() {
```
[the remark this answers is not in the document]
Luthaf: Since this is a simple wrapper, I would rather re-implement it on the Python side directly. This reduces the number of functions/types in the shared library, reducing the cost to add new wrappers

equistore-torch/src/array.cpp
Luthaf: this is just initialized to 0 here (meaning "no origin"), and later allocated by the code in `EqsOriginRegistration`

docs/src/reference/torch/index.rst
```diff
@@ -0,0 +1,91 @@
...
+    from equistore import TensorMap, TensorBlock, Labels
+
+to
+
+.. code-block:: python
+
+    import equistore.torch as equistore
+    from equistore.torch import TensorMap, TensorBlock, Labels
+
+is 80% of the work required to make a model developed in Python --- without
```
agoscinski, line 32: If you only use equistore operations from core and equistore-operations, that will be True, but if you have custom equistore operations then not.

```suggestion
makes all the handling of equistore objects through equistore-operations compatible with TorchScript.
```

docs/src/reference/torch/index.rst
```diff
@@ -0,0 +1,91 @@
...
+such as `Jax`_ arrays, `cupy`_ arrays, *etc.*.
+
+While this works great for a lot of use cases based on torch (defining models,
+training them with autograd, …), the TorchScript compiler is more restrictive in
+what code it accepts and does not support calling into arbitrary native code
+like the Python API does. The TorchScript version of equistore is here to fill
+this gap, teaching the TorchScript compiler what's going on inside equistore and
+how to translate code using equistore to TorchScript.
+
+Another different is that while the Python API supports storing data in multiple
```
agoscinski, line 79:
```suggestion
Another difference is that while the Python API supports storing data in multiple
```

equistore-torch/src/array.cpp
Luthaf: I'll add some more comment around this

docs/src/reference/torch/index.rst
```diff
@@ -0,0 +1,91 @@
...
+ways, including storing it on GPU through torch Tensor, the metadata is always
+stored on CPU, inside the equistore shared library. The TorchScript API changes
+this to enable storing both the data and metadata on GPU, minimizing data
+transfer and making the models faster.
+
+If you only care about PyTorch, we would recommend using the TorchScript API
```
agoscinski, line 85:
```suggestion
If your code only needs to work with PyTorch, we would recommend using the TorchScript API
```

docs/src/reference/torch/index.rst
```diff
@@ -0,0 +1,91 @@
...
+some level of overlap (it is possible to store data in torch ``Tensor`` with the
+Python API), so a big question is when/why should you use one or the other, and
+what's the difference between them.
+
+First, the Python and TorchScript API are separate to allow using the Python API
+without torch, for example in a pure numpy workflow, or with other array types
+such as `Jax`_ arrays, `cupy`_ arrays, *etc.*.
+
+While this works great for a lot of use cases based on torch (defining models,
+training them with autograd, …), the TorchScript compiler is more restrictive in
```
agoscinski, line 73:
```suggestion
training them with autograd, …), the `TorchScript`_ compiler is more restrictive in
```
added reference suggestion at the end

docs/src/reference/torch/index.rst
```diff
@@ -0,0 +1,91 @@
...
+this to enable storing both the data and metadata on GPU, minimizing data
+transfer and making the models faster.
+
+If you only care about PyTorch, we would recommend using the TorchScript API
+from the start, to make sure you will be able to export your code to
+TorchScript. If you are not using PyTorch, or if you want to write code in an
+engine agnostic way, we recommend using the Python API.
+
+.. _Jax: https://jax.readthedocs.io/en/latest/
+.. _cupy: https://cupy.dev
```
agoscinski, line 91:
```suggestion
.. _cupy: https://cupy.dev
.. _TorchScript: https://pytorch.org/docs/stable/jit_language_reference.html#language-reference
```

PR #266, opened 2023-05-02
Check that we can build the whole project for Windows

Luthaf:
This builds on top of #267


PicoCentauri (approved):
The cmake config looks horrible 🤣.

Seems to work though. Very nice. Lets wait for our actual first windows users.

PR #267, opened 2023-05-03
Upgrade minimal CMake version to 3.16

Luthaf:
This is the default for Ubuntu 20.04, and it is relatively easy to get a newer version if someone needs it on a different system.

----
:books: Documentation preview :books:: https://equistore--267.org.readthedocs.build/en/267/

### DIFF
--- a/equistore-core/CMakeLists.txt
+++ b/equistore-core/CMakeLists.txt
@@ -3,7 +3,7 @@
 # This API is implemented in Rust, in the equistore-core crate, but Rust users
 # of the API should use the equistore crate instead, wrapping equistore-core in
 # an easier to use, idiomatic Rust API.
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.16)
 
 # Is equistore the main project configured by the user? Or is this being used
 # as a submodule/subdirectory?
@@ -24,9 +24,6 @@ if (POLICY CMP0077)
     # use variables to set OPTIONS
     cmake_policy(SET CMP0077 NEW)
 endif()
-
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 file(STRINGS "Cargo.toml" CARGO_TOML_CONTENT)
 foreach(line ${CARGO_TOML_CONTENT})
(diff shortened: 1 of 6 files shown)


equistore-core/CMakeLists.txt
```diff
@@ -209,27 +206,18 @@ set_target_properties(equistore::shared PROPERTIES
...
-if (${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.11)
-    if (BUILD_SHARED_LIBS)
-        add_library(equistore ALIAS equistore::shared)
-    else()
-        add_library(equistore ALIAS equistore::static)
-    endif()
+if (BUILD_SHARED_LIBS)
+    add_library(equistore ALIAS equistore::shared)
 else()
-    # CMake 3.10 (default on Ubuntu 20.04) does not support ALIAS for IMPORTED
```
agoscinski, line 225: Was that comment wrong? Was this referring to Ubuntu 18.04?

equistore-core/tests/cpp/CMakeLists.txt
```diff
@@ -63,6 +59,7 @@ foreach(_file_ ${ALL_TESTS})
         # cmake does not find the library on the filesystem and does not add the
         # RPATH to executables linking to it
         BUILD_RPATH ${EQUISTORE_DIR}
+        NO_SYSTEM_FROM_IMPORTED ON
```
agoscinski, line 62: For what is this needed? I found a description, but I don't understand when this is useful.

> If, for some reason, a consumer should not treat header search paths from imported targets as system search paths, one can set the consumer’s NO_SYSTEM_FROM_IMPORTED target property to true.

equistore-core/tests/cpp/external/CMakeLists.txt
```diff
@@ -1,33 +1,3 @@
-function(unpack_library _name_)
```
agoscinski, line 1: Is the removal related to the cmake update? I just could not function used in the repo

equistore-core/tests/cpp/CMakeLists.txt
Luthaf: So we get warnings from these headers when compiling the tests

equistore-core/CMakeLists.txt
Luthaf: yes, this should have been ubuntu 18.04

equistore-core/tests/cpp/external/CMakeLists.txt
Luthaf: yes, the function was never used ^_^ I copy-pasted it from another project thinking I'll need it, but ended up doing things differently

agoscinski (approved):
Looks good

PR #268, opened 2023-05-04
Cancel running CI when a new commit is pushed to the branch

Luthaf:
Taken from https://stackoverflow.com/a/70972844


Luthaf:
@PicoCentauri should we do this on all lab-cosmo repos?

PicoCentauri:
Yes sounds like a very useful addition.

PicoCentauri (approved):

PR #269, opened 2023-05-08
Fix all function for torch array when axis is None

agoscinski:
dim=None results in an error, because torch uses a different function when dim is not given.

I would like to add a test, but it seems like one has to to adapt the tests that they also run for torch arrays and that is a bit too much for me now.

----
:books: Documentation preview :books:: https://equistore--269.org.readthedocs.build/en/269/

### DIFF
--- a/python/equistore-operations/equistore/operations/_dispatch.py
+++ b/python/equistore-operations/equistore/operations/_dispatch.py
@@ -32,7 +32,12 @@ def all(a, axis=None):
     if isinstance(a, np.ndarray):
         return np.all(a=a, axis=axis)
     elif isinstance(a, TorchTensor):
-        return torch.all(input=a, dim=axis)
+        # torch.all has two implementation, and picks one depending if more than one
+        # parameter is given. The second one does not supports setting dim to `None`
+        if axis is None:
+            return torch.all(input=a)
+        else:
+            return torch.all(input=a, dim=axis)
     else:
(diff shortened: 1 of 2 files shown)


agoscinski:
Okay added now a new test file that will use the torch version if torch is available. I am not sure if we have at the moment a github action running with torch, but at least this way, we have the fix covered in the future when we add such a github action.

python/equistore-operations/tests/all.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
```
Luthaf, line 1: I would put this in `tests/dispatch.py` instead of `all.py`

python/equistore-operations/equistore/operations/_dispatch.py
```diff
@@ -32,7 +32,10 @@ def all(a, axis=None):
     if isinstance(a, np.ndarray):
         return np.all(a=a, axis=axis)
     elif isinstance(a, TorchTensor):
-        return torch.all(input=a, dim=axis)
+        if axis is None:
```
Luthaf, line 35:
```suggestion
        # torch.all has two implementation, and picks one depending if more than one parameter 
        # is given. The second one does not supports setting dim to `None`
        if axis is None:
```

Luthaf:
Yes, we run the operations tests with torch installed!

DavideTisi (commented):
once the comments from @Luthaf are solved for  me it is fine

PR #270, opened 2023-05-09
Create shared and static libequistore directly

Luthaf:
Previously we would create these as libequistore_core.so, and rename the file to libequistore.so. This is much harder to do on Windows, where the `.dll.lib` file contains an hardcoded `equistore_core.dll`, and modifying the `.dll.lib` file is non-trivial (it typically requires an additional python script to do this).

Since we actually don't need to build the [rlib](https://doc.rust-lang.org/reference/linkage.html) for equistore-core, we can set lib.name to `equistore` and everything works just fine!

----
:books: Documentation preview :books:: https://equistore--270.org.readthedocs.build/en/270/


Luthaf:
Another proof that this works is that we can link and load to equistore.dll (through equistore_torch.dll) in https://github.com/lab-cosmo/equistore/pull/233

equistore-core/Cargo.toml
```diff
@@ -11,7 +11,8 @@ exclude = [
 [lib]
 # When our minimal Rust version becomes 1.64, we can pass these options directly
 # to Cargo. Until then, we build all the crate-type we need.
-crate-type = ["cdylib", "staticlib", "rlib"]
+name = "equistore"
```
PicoCentauri, line 14: `equistore` or `equistore-core`. Maybe the first becuae there is no `equistore-core` is the only rust package, no?

Luthaf: so that's the main change here: by setting the name of the lib to `equistore`, rustc will create equistore.dll and equistore.dll.lib directly. Before it was using the default `equistore_core` as the name, and we had to rename the files

PR #271, opened 2023-05-15
Make `LabelsBuilder::new` return an error for invalid names

Luthaf:
It previously used an assert, resulting in this error message `internal error (this is likely a bug, please report it): invalid labels: the same name is used multiple times` when the user gave invalid names, but this is not an internal error.

Now if the user gives invalid name they will get `invalid parameter: labels names must be unique, got 'foo' multiple times`


equistore-core/src/blocks.rs
```diff
@@ -447,7 +447,7 @@ mod tests {
             let mut block = TensorBlock::new(values, samples, vec![], properties.clone()).unwrap();
             assert!(block.gradients().is_empty());
 
-            let mut gradient_samples = LabelsBuilder::new(vec!["sample", "foo"]);
+            let mut gradient_samples = LabelsBuilder::new(vec!["sample", "foo"]).expect("invalide names");
```
jwa7, line 450:
```suggestion
            let mut gradient_samples = LabelsBuilder::new(vec!["sample", "foo"]).expect("invalid names");
```

equistore-core/src/labels.rs
```diff
@@ -332,3 +342,17 @@ impl std::ops::Index<usize> for Labels {
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn valid_names() {
+        let e = LabelsBuilder::new(vec!["not an ident"]).err().unwrap();
+        assert_eq!(e.to_string(), "invalid parameter: all labels names must be valid identifiers, 'not an ident' is not");
+
+        let e = LabelsBuilder::new(vec!["not", "there", "not"]).err().unwrap();
+        assert_eq!(e.to_string(), "invalid parameter: labels names must be unique, got 'not' multiple times");
```
jwa7, line 356: Is this definitely tested correctly?

When I try:
```
Labels(names=["not", "there", "not"], values=np.array([[0, 1, 2],]))
```
with this build I get a different error:
`ValueError: field 'not' occurs more than once`

jwa7 (approved):
Looks good!

PR #279, opened 2023-06-06
Fix version specification for Python packages

Luthaf:
`~=` means "not this version"

----
:books: Documentation preview :books:: https://equistore--279.org.readthedocs.build/en/279/


Luthaf:
I realised this when trying to install from https://github.com/Luthaf/nightly-wheels

PR #280, opened 2023-06-07
Removing try-catch blocks from equistore-operations

agoscinski:
To support TorchScript for the equistore-operations we need to remove all try-catch block. This includes these files
* equal.py
* equal_metadata.py
* allclose.py

We remove the try-catch block by implementing a `_impl_operation` function, that returns an error message, `operation_raise` will raise, `operation` will just check if the error message is empty and depending on this return True or False.

I started with equal.py. The test for this one should pass. **I will continue with the other operations, once I get feedback that the changes in `equal.py` are good**. There is just one small change here worth mentioning: We cannot return simply a backtrace of errors, so I just concatenate the error messages
```python
-        try:
-            equal_block_raise(block_1, tensor_2.block(key))
-        except NotEqualError as e:
-            raise NotEqualError(f"blocks for key '{key}' are different") from e
+        message = _equal_block_impl(block_1, tensor_2.block(key))
+        if message != "":
+            return f"blocks for key '{key}' are different: " + message 
```

### DIFF
--- a/python/equistore-operations/equistore/operations/_utils.py
+++ b/python/equistore-operations/equistore/operations/_utils.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Union
 
 import numpy as np
 
@@ -11,7 +11,17 @@ class NotEqualError(Exception):
     pass
 
 
-def _check_same_keys(a: TensorMap, b: TensorMap, fname: str):
+def _check_same_keys(a: TensorMap, b: TensorMap, fname: str) -> bool:
+    return not bool(_check_same_keys_impl(a, b, fname))
+
+
+def _check_same_keys_raise(a: TensorMap, b: TensorMap, fname: str) -> None:
+    message = _check_same_keys_impl(a, b, fname)
+    if message != "":
+        raise NotEqualError(message)
+
+
+def _check_same_keys_impl(a: TensorMap, b: TensorMap, fname: str) -> str:
     """Check if metadata between two TensorMaps is consistent for an operation.
 
     The functions verifies that
@@ -22,28 +32,46 @@ def _check_same_keys(a: TensorMap, b: TensorMap, fname: str):
 
     :param a: first :py:class:`TensorMap` for check
     :param b: second :py:class:`TensorMap` for check
+    :param fname: name of the function where the check is performed. The input will be
+                  used to generate a meaningful error message.
     """
 
     keys_a = a.keys
     keys_b = b.keys
 
     if keys_a.names != keys_b.names:
-        raise NotEqualError(
+        return (
             f"inputs to {fname} should have the same keys names, "
             f"got '{keys_a.names}' and '{keys_b.names}'"
         )
 
     if len(keys_a) != len(keys_b):
-        raise NotEqualError(
+        return (
             f"inputs to {fname} should have the same number of blocks, "
             f"got {len(keys_a)} and {len(keys_b)}"
         )
 
     if not np.all([key in keys_a for key in keys_b]):
-        raise NotEqualError(f"inputs to {fname} should have the same keys")
+        return f"inputs to {fname} should have the same keys"
+
+    return ""
 
 
-def _check_blocks(a: TensorBlock, b: TensorBlock, props: List[str], fname: str):
+def _check_blocks(a: TensorBlock, b: TensorBlock, props: List[str], fname: str) -> str:
+    return not bool(_check_blocks_impl(a, b, props, fname))
+
+
+def _check_blocks_raise(
+    a: TensorBlock, b: TensorBlock, props: List[str], fname: str
+) -> None:
+    message = _check_blocks_impl(a, b, props, fname)
+    if message != "":
+        raise NotEqualError(message)
+
+
+def _check_blocks_impl(
+    a: TensorBlock, b: TensorBlock, props: List[str], fname: str
+) -> str:
     """Check if metadata between two TensorBlocks is consistent for an operation.
 
     The functions verifies that that the metadata of the given props is the same
@@ -63,42 +91,45 @@ def _check_blocks(a: TensorBlock, b: TensorBlock, props: List[str], fname: str):
 
         if prop == "samples":
             if not len(a.samples) == len(b.samples):
-                raise NotEqualError(err_msg + err_msg_len)
+                return err_msg + err_msg_len
             if not a.samples.names == b.samples.names:
-                raise NotEqualError(err_msg + err_msg_names)
+                return err_msg + err_msg_names
             if not np.all(a.samples == b.samples):
-                raise NotEqualError(err_msg + err_msg_1)
+                return err_msg + err_msg_1
 
         elif prop == "properties":
             if not len(a.properties) == len(b.properties):
-                raise NotEqualError(err_msg + err_msg_len)
+                return err_msg + err_msg_len
             if not a.properties.names == b.properties.names:
-                raise NotEqualError(err_msg + err_msg_names)
+                return err_msg + err_msg_names
             if not np.all(a.properties == b.properties):
-                raise NotEqualError(err_msg + err_msg_1)
+                return err_msg + err_msg_1
 
         elif prop == "components":
             if len(a.components) != len(b.components):
-                raise NotEqualError(err_msg + err_msg_len)
+                return err_msg + err_msg_len
 
             for c1, c2 in zip(a.components, b.components):
                 if not (c1.names == c2.names):
-                    raise NotEqualError(err_msg + err_msg_names)
+                    return err_msg + err_msg_names
 
                 if not (len(c1) == len(c2)):
-                    raise NotEqualError(err_msg + err_msg_1)
+                    return err_msg + err_msg_1
 
                 if not np.all(c1 == c2):
-                    raise NotEqualError(err_msg + err_msg_1)
+                    return err_msg + err_msg_1
 
         else:
             raise ValueError(
                 f"{prop} is not a valid property to check, "
                 "choose from ['samples', 'properties', 'components']"
             )
+    return ""
 
 
-def _check_same_gradients(a: TensorBlock, b: TensorBlock, props: List[str], fname: str):
+def _check_same_gradients(
+    a: TensorBlock, b: TensorBlock, props: List[str], fname: str
+) -> bool:
     """
     Check if metadata between two gradients's TensorBlocks is consistent
      for an operation.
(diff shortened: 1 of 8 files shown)


python/equistore-operations/equistore/operations/_utils.py
```diff
@@ -11,7 +11,7 @@ class NotEqualError(Exception):
     pass
 
 
-def _check_same_keys(a: TensorMap, b: TensorMap, fname: str):
+def _check_same_keys(a: TensorMap, b: TensorMap, fname: str) -> str:
```
Luthaf, line 14: these functions are used elsewhere in code that assumes they raise. So there should also be a `_raise` and `_impl` version for these functions.

Luthaf (commented):
this looks good overall, with a small issue related to the code in `_utils`

python/equistore-operations/equistore/operations/equal.py
```diff
@@ -51,16 +51,20 @@ def equal_raise(tensor_1: TensorMap, tensor_2: TensorMap):
     :param tensor_1: first :py:class:`TensorMap`.
     :param tensor_2: second :py:class:`TensorMap`.
     """
-    try:
-        _check_same_keys(tensor_1, tensor_2, "equal")
-    except ValueError as e:
```
agoscinski, line 56: Was this actually a bug that a `ValueError` was caught here and not a `NotEqualError`? Because `_check_same_keys` only raises `NotEqualError` errors. and `NotEqualError` inherits from Exception

agoscinski (commented):
Do you think it makes now more sense to just merge the functions and add a `raise_error` option? It mainly is beneficial to reduce redundancy in the documentation of the raising and non raising functions. It reads a bit weirdly that the raising function links to the non raising function for more doc.

I checked with TorchScript if you can have both
```
import torch
from typing import Union

@torch.jit.script
def foo(a: int) -> Union[None, bool]: # does compile
    if a == 5:
        raise ValueError("a should not be 5")
    else:
        return True
```
At the moment the equal checks can anyway raise a ValueError or NotImplemented error. So equal check could also return None in any case.

Luthaf:
You mean merge `equal` and `equal_raise`? I could see a point for it from a developer point of view, but from a user point of view it feels a bit weird to have a function that either returns `True`/`False` or raises an error. Might be worth to ask for more feedback on this point @ceriottm @jwa7 @PicoCentauri @DavideTisi what do you think?

ceriottm:
What is the difference between having an equal_raise and using assert(allequal())?

Luthaf:
The idea was that `equal_raise` will include what's different between the arguments in the error message

agoscinski:
so we are talking about these two options
```python
def operation(..., raise_error=False):
    ...
```
and
```python
def operation(...):
    ...
def operation_raise(...):
    ...
```
I mainly stumbled on it because it duplicates documentation a bit and I will have to refactor all these `operation_raise` functions for this PR anyway. I don't have strong opinions.

agoscinski:
I just keep the pattern we use at the moment
```python
def operation(...):
    ...
def operation_raise(...):
    ...
```


I updated the `_utils.py` function. Is this okay like this? Because I need to update all other operations where they are used. Could you check the usage in `equal.py` which passes the tests @Luthaf?

python/equistore-operations/equistore/operations/equal.py
```diff
@@ -81,14 +85,14 @@ def equal_block(block_1: TensorBlock, block_2: TensorBlock) -> bool:
...
     :param block_2: second :py:class:`TensorBlock`.
     """
-    try:
-        equal_block_raise(block_1=block_1, block_2=block_2)
+    message = _equal_block_impl(block_1=block_1, block_2=block_2)
+    if message == "":
         return True
-    except NotEqualError:
+    else:
         return False
```
Luthaf, line 92: Since a string is contextually convertible to bool in Python, this could also be `return not message`

python/equistore-operations/equistore/operations/equal.py
```diff
@@ -51,16 +51,20 @@ def equal_raise(tensor_1: TensorMap, tensor_2: TensorMap):
...
 
     for key, block_1 in tensor_1:
-        try:
-            equal_block_raise(block_1, tensor_2.block(key))
-        except NotEqualError as e:
-            raise NotEqualError(f"blocks for key '{key}' are different") from e
+        message = _equal_block_impl(block_1, tensor_2.block(key))
+        if message != "":
+            return f"blocks for key '{key}' are different: " + message
+    return ""
```
Luthaf, line 67:
```suggestion
        if message:
            return f"blocks for key '{key}' are different: " + message
    return ""
```

Luthaf: Also, we could have a `NO_ERROR = ""` variable, and return that for better readability

python/equistore-operations/tests/equal_metadata.py
```diff
@@ -197,28 +196,28 @@ def test_single_nonexisting_meta(test_tensor_map_1, test_tensor_map_2):
     # wrong metadata key alone
     wrong_meta = "species"
     error_message = f"Invalid metadata to check: {wrong_meta}"
-    with pytest.raises(ValueError, match=error_message):
-        equistore.equal_metadata(
```
PicoCentauri, line 201: I don't know how was this passing the tests before 🤣

PicoCentauri:
@agoscinski I continued your idea and removed all `try-except` blocks from the operations. During this I also unified the `equal` and `allclose` operations and did some more syntax highlighting and structuring within the docs of the touched functions.

python/equistore-operations/equistore/operations/_utils.py
```diff
@@ -192,3 +275,90 @@ def _check_gradient_presence(block: TensorBlock, parameters: List[str], fname: s
...
+    """
+    # They can only be equivalent if the same length
+    if len(a) != len(b):
+        return False
+    # Check the names
+    if not np.all(a.names == b.names):
+        return False
+    if exact_order:
+        return np.all(np.array(a == b))
+    return np.all([a_i in b for a_i in a])
```
PicoCentauri, line 295: We have to remove all numpy operations here for torch script right? But maybe we do this in an upcoming PR....

agoscinski: yes we need to dispatch them all

python/equistore-operations/equistore/operations/_utils.py
```diff
@@ -192,3 +275,90 @@ def _check_gradient_presence(block: TensorBlock, parameters: List[str], fname: s
...
+            block_2=tensor_2.block(key),
+            **extra_args,
+        )
+        if message != NO_ERROR:
+            return f"blocks for key '{key}' are different: {message}"
+    return NO_ERROR
+
+
+def _comparison_block_impl(
+    operation: str, block_1: TensorBlock, block_2: TensorBlock, **extra_args
```
agoscinski, line 319: TorchScript does not support kwargs. Also does not support lambda functions and function definitions inside functions. I gave it a thought how to get something similar as you do with the comparison function, but I could not come up with anything that is compatible with TorchScript. The problem is that you want arbitrary callables **and** arbitrary key word arguments. I could not find a solution that can do both

python/equistore-operations/equistore/operations/_utils.py
PicoCentauri: okay!

python/equistore-operations/equistore/operations/_utils.py
PicoCentauri: Okay I can restructure this. But passing functions is fine? Then we can just factor out this logic.

PicoCentauri:
I put back the explicit functions for the implementations of `equal` and `allclose` 😞. Let me know if this is okay.

PicoCentauri:
Global variables are not allowed in torch script. I had to remove your nice suggestion of `NO_ERROR` @Luthaf

python/equistore-operations/equistore/operations/_utils.py
agoscinski: Is this function still needed or useful for some future PR? I could not find a usage in this PR

python/equistore-operations/tests/equal_metadata.py
agoscinski: I think it was raising a ValueError before the PR. In my first commit I changed it to NotEqualError

python/equistore-operations/equistore/operations/equal.py
agoscinski: yes maybe one could do `bool(message)` I don.t have big preferences. This way it is easier to read, but takes more lines to express

agoscinski (commented):
Looks like terrible TorchScript code. Good job!

python/equistore-operations/equistore/operations/_utils.py
jwa7: I think this function is a relic of the past; I wrote it ages ago to help with metadata checks when you couldn't just do `labels_a == labels_b`. So fairly sure it's not needed anymore, but not 100 %. @Luthaf ?

agoscinski: Then it was removed already and was somehow reintroduced on a rebase here? Would just remove it

PicoCentauri: Okay I will remove it!

python/equistore-operations/equistore/operations/equal.py
PicoCentauri: I think bool(message) is a good idea! This will remove some code as well.

python/equistore-operations/tests/equal_metadata.py
PicoCentauri: Ahh okay!

python/equistore/equistore/__init__.py
```diff
@@ -31,7 +31,8 @@
     from . import operations  # noqa
 
     HAS_EQUISTORE_OPERATIONS = True
-except ImportError:
+except ImportError as e:
```
agoscinski, line 34:
```suggestion
except ImportError:
```

PicoCentauri:
Thanks for the hint @agoscinski. Code is now a bit readable I would say.

python/equistore-operations/equistore/operations/allclose.py
```diff
@@ -1,10 +1,81 @@
...
+    for key, block_1 in tensor_1.items():
+        message = _allclose_block_impl(
+            block_1=block_1,
+            block_2=tensor_2.block(key),
+            rtol=rtol,
+            atol=atol,
+            equal_nan=equal_nan,
+        )
+        if message != "":
+            return f"blocks for key '{key}' are different: {message}"
```
Luthaf, line 29:
```suggestion
            return f"blocks for {key.print()} are different: {message}"
```

This will produce `blocks for (aa=3, bb=4) are different` instead of `blocks for key LabelsEntry(aa=3, bb=4) are different`

Luthaf: (we can not use different `__repr__` and `__str__` for this, since torch did not support customizing `__repr__`, and this will only be available in torch >2.1)

Luthaf (commented):
This looks good to me, I'm leaving the final approval to @agoscinski!

agoscinski:
Approve (its my PR cannot approve using Github)

PicoCentauri (approved):
LETS MERGE!

PR #281, opened 2023-06-08
Implementation of sort operation

agoscinski:
Still a draft, but some comments on the signature of the sort function are already welcomed. I marked questions for review with `PR COMMENT`

Missing:
* sorting of gradients
* dispatching np.take function
* tests

----
:books: Documentation preview :books:: https://equistore--281.org.readthedocs.build/en/281/

### DIFF
--- a/python/metatensor-operations/metatensor/operations/_dispatch.py
+++ b/python/metatensor-operations/metatensor/operations/_dispatch.py
@@ -614,7 +614,7 @@ def where(array):
 
 def take(array, indices, axis: int):
     """
-    See torch take_along_dim or numpy take as reference.
+    See :py:func:`torch.index_select` or :py:func:`numpy.take` as reference.
 
     Because numpy and torch have different APIs we went for the more limited one
     in torch not supporting ``mode`` argument. The argument ``out`` is not supported
@@ -635,68 +635,31 @@ def take(array, indices, axis: int):
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
-# PR COMMENT I implemented argsort because I thought I need it, but it turned out not
-#            but I would nevertheless keep it, might be useful for something else
-def argsort(array, axis=-1, descending: bool = False, stable: bool = False):
+def argsort_labels_values(labels_values, reverse: bool = False):
     """
-    See torch argsort as reference.
+    Similar to :py:func:`np.argsort`, but sort the rows as one aggregated
+    tuple.
 
-    Because numpy and torch have different APIs we went for the more limited one
-    in torch. Numpy supports more options for sorting that are not here available.
+    :param labels_values: numpy.array or torch.Tensor
+    :param reverse: if true, order is descending
 
-    :param array: the array that is argsorted
-    :param axis: axis of array to argsort
-    :param stable: if true, the order of duplicate elements stays the same an the array
-    :param descending: if false, the order is ascending
-
-    :return: indices of the sorted array wrt. the unsorted one
-    """
-    if isinstance(array, TorchTensor):
-        return torch.argsort(array, dim=axis, descending=descending, stable=stable)
-    elif isinstance(array, np.ndarray):
-        if stable:
-            idx_ascending = np.argsort(array, axis=axis, kind="stable")
-        else:
-            idx_ascending = np.argsort(array, axis=axis, kind="quicksort")
-
-        if descending:
-            return np.flip(idx_ascending, axis=axis)
-        else:
-            return idx_ascending
-    else:
-        raise TypeError(UNKNOWN_ARRAY_TYPE)
-
-
-def argsort_metadata_values(metadata_values, reverse: bool = False):
-    """
-    In this function we append an index at the end to get the indices of the
-    corresponding to the sorted values (argsort).  Because we append the indices at the
-    end and since metadata is unique, we do not affect the sorted order.
-
-    We are highly limited by TorchScript with the sorting functions: torch.argsort does
-    not work because it cannot sort over the rows considering each row as one aggregated
-    element The ´´reverse´´ keyword argument is only supported by TorchScript using sort
-    like l.sort(reverse=...), but not using sorted(reverse=...) the ``key`` keyword
-    argument is not supported by TorchScript at all
-
-    :param metadata_values: numpy.array or torch.Tensor :param reverse: if true, order
-    is descending
-
-    :return: indices corresponding to the sorted values in py:obj:`metadata_values`
+    :return: indices corresponding to the sorted values in ``labels_values``
 
     """
-    if isinstance(metadata_values, TorchTensor):
-        # PR COMMENT torchscript does not support sort of List[List[int]]
-        #            the algorithm will run into numerical problems for high max ints
-        #            for future PR would do a unique before to map to a lower range
-        max_int = torch.max(metadata_values)
+    if isinstance(labels_values, TorchTensor):
+        # torchscript does not support sorted for List[List[int]]
+        # so we temporary do this trick. this will be fixed with issue #366
+        max_int = torch.max(labels_values)
         idx = torch.sum(
-            max_int ** torch.arange(metadata_values.shape[1]) * metadata_values, dim=1
+            max_int ** torch.arange(labels_values.shape[1]) * labels_values, dim=1
         )
         return torch.argsort(idx, dim=-1, descending=reverse)
-    elif isinstance(metadata_values, np.ndarray):
-        list_tuples: List[List[int]] = metadata_values.tolist()
-        for i in range(len(metadata_values)):
+    elif isinstance(labels_values, np.ndarray):
+        # Index is appended at the end to get the indices corresponding to the
+        # sorted values. Because we append the indices at the end and since metadata
+        # is unique, we do not affect the sorted order.
+        list_tuples: List[List[int]] = labels_values.tolist()
+        for i in range(len(labels_values)):
             list_tuples[i].append(i)
         list_tuples.sort(reverse=reverse)
         return np.array(list_tuples)[:, -1]
--- a/python/metatensor-operations/metatensor/operations/_dispatch.py
+++ b/python/metatensor-operations/metatensor/operations/_dispatch.py
@@ -610,3 +610,95 @@ def where(array):
         return np.where(array)
     else:
         raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def take(array, indices, axis: int):
+    """
+    See torch take_along_dim or numpy take as reference.
+
+    Because numpy and torch have different APIs we went for the more limited one
+    in torch not supporting ``mode`` argument. The argument ``out`` is not supported
+    by TorchScript, or at least it is nontrivial to add
+
+    :param array: the array the elements are returned from
+    :param indices: the indices to take long an axes
+    :param axis: axis of array to take
+
+    :return: the elments at the indices in the array along dimension specified by the
+    axis
+    """
+    if isinstance(array, TorchTensor):
+        return torch.index_select(array, dim=axis, index=indices)
+    elif isinstance(array, np.ndarray):
+        return np.take(array, indices, axis=axis)
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+# PR COMMENT I implemented argsort because I thought I need it, but it turned out not
+#            but I would nevertheless keep it, might be useful for something else
+def argsort(array, axis=-1, descending: bool = False, stable: bool = False):
+    """
+    See torch argsort as reference.
+
+    Because numpy and torch have different APIs we went for the more limited one
+    in torch. Numpy supports more options for sorting that are not here available.
+
+    :param array: the array that is argsorted
+    :param axis: axis of array to argsort
+    :param stable: if true, the order of duplicate elements stays the same an the array
+    :param descending: if false, the order is ascending
+
+    :return: indices of the sorted array wrt. the unsorted one
+    """
+    if isinstance(array, TorchTensor):
+        return torch.argsort(array, dim=axis, descending=descending, stable=stable)
+    elif isinstance(array, np.ndarray):
+        if stable:
+            idx_ascending = np.argsort(array, axis=axis, kind="stable")
+        else:
+            idx_ascending = np.argsort(array, axis=axis, kind="quicksort")
+
+        if descending:
+            return np.flip(idx_ascending, axis=axis)
+        else:
+            return idx_ascending
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def argsort_metadata_values(metadata_values, reverse: bool = False):
+    """
+    In this function we append an index at the end to get the indices of the
+    corresponding to the sorted values (argsort).  Because we append the indices at the
+    end and since metadata is unique, we do not affect the sorted order.
+
+    We are highly limited by TorchScript with the sorting functions: torch.argsort does
+    not work because it cannot sort over the rows considering each row as one aggregated
+    element The ´´reverse´´ keyword argument is only supported by TorchScript using sort
+    like l.sort(reverse=...), but not using sorted(reverse=...) the ``key`` keyword
+    argument is not supported by TorchScript at all
+
+    :param metadata_values: numpy.array or torch.Tensor :param reverse: if true, order
+    is descending
+
+    :return: indices corresponding to the sorted values in py:obj:`metadata_values`
+
+    """
+    if isinstance(metadata_values, TorchTensor):
+        # PR COMMENT torchscript does not support sort of List[List[int]]
+        #            the algorithm will run into numerical problems for high max ints
+        #            for future PR would do a unique before to map to a lower range
+        max_int = torch.max(metadata_values)
+        idx = torch.sum(
+            max_int ** torch.arange(metadata_values.shape[1]) * metadata_values, dim=1
+        )
+        return torch.argsort(idx, dim=-1, descending=reverse)
+    elif isinstance(metadata_values, np.ndarray):
+        list_tuples: List[List[int]] = metadata_values.tolist()
+        for i in range(len(metadata_values)):
+            list_tuples[i].append(i)
+        list_tuples.sort(reverse=reverse)
+        return np.array(list_tuples)[:, -1]
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
--- a/python/metatensor-operations/metatensor/operations/_dispatch.py
+++ b/python/metatensor-operations/metatensor/operations/_dispatch.py
@@ -113,19 +113,6 @@ def bincount(input, weights: Optional[TorchTensor] = None, minlength: int = 0):
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
-def bool_array_like(bool_list: List[bool], like):
-    """
-    Converts the input list of bool to a numpy array or torch tensor
-    based on the type of `like`.
-    """
-    if isinstance(like, TorchTensor):
-        return torch.tensor(bool_list, dtype=torch.bool, device=like.device)
-    elif isinstance(like, np.ndarray):
-        return np.array(bool_list).astype(bool)
-    else:
-        raise TypeError(UNKNOWN_ARRAY_TYPE)
-
-
 def copy(array):
     """Returns a copy of ``array``.
     The new data is not shared with the original array"""
@@ -151,15 +138,28 @@ def eye_like(array, size: int):
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
-def int_array_like(shape: List[int], fill_value: int, like):
+def bool_array_like(bool_list: List[bool], like):
     """
-    Fills an array with a value. This will be a numpy array or a torch tensor based on
-    the `like` argument.
+    Converts the input list of bool to a numpy array or torch tensor
+    based on the type of `like`.
     """
     if isinstance(like, TorchTensor):
-        return torch.full(shape, fill_value, dtype=like.dtype, device=like.device)
+        return torch.tensor(bool_list, dtype=torch.bool, device=like.device)
     elif isinstance(like, np.ndarray):
-        return np.full(shape, fill_value)
+        return np.array(bool_list).astype(bool)
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def int_array_like(int_list: List[int], like):
+    """
+    Converts the input list of int to a numpy array or torch tensor
+    based on the type of `like`.
+    """
+    if isinstance(like, TorchTensor):
+        return torch.tensor(int_list, dtype=torch.int64, device=like.device)
+    elif isinstance(like, np.ndarray):
+        return np.array(int_list).astype(np.int64)
     else:
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
--- a/python/metatensor-operations/metatensor/operations/sort.py
+++ b/python/metatensor-operations/metatensor/operations/sort.py
@@ -72,8 +72,8 @@ def sort_block(
 
     :param axes: axes to sort. The labels entries along these axes will be sorted in
         lexicographic order, and the arrays values will be reordered accordingly.
-        Possible values are ``'keys'``, ``'samples'``, ``'components'``,
-        ``'properties'`` and ``'all'`` to sort everything.
+        Possible values are ``'samples'``, ``'components'``, ``'properties'`` and
+        ``'all'`` to sort everything.
 
     :param descending: if false, the order is ascending
 
@@ -193,8 +193,8 @@ def sort(
 
     :param axes: axes to sort. The labels entries along these axes will be sorted in
         lexicographic order, and the arrays values will be reordered accordingly.
-        Possible values are ``'samples'``, ``'components'``, ``'properties'`` and
-        ``'all'`` to sort everything.
+        Possible values are ``'keys'``, ``'samples'``, ``'components'``,
+        ``'properties'`` and ``'all'`` to sort everything.
     :param descending: if false, the order is ascending
     :return: sorted tensor map
 
--- a/python/metatensor-operations/metatensor/operations/sort.py
+++ b/python/metatensor-operations/metatensor/operations/sort.py
@@ -27,20 +27,18 @@ def _sort_single_block(
 
     values = block.values
     if "samples" in axes:
-        sorted_idx = _dispatch.argsort_metadata_values(
-            samples_values, reverse=descending
-        )
+        sorted_idx = _dispatch.argsort_labels_values(samples_values, reverse=descending)
         samples_values = samples_values[sorted_idx]
         values = values[sorted_idx]
     if "components" in axes:
         for i, _ in enumerate(block.components):
-            sorted_idx = _dispatch.argsort_metadata_values(
+            sorted_idx = _dispatch.argsort_labels_values(
                 components_values[i], reverse=descending
             )
             components_values[i] = components_values[i][sorted_idx]
             values = _dispatch.take(values, sorted_idx, axis=i + 1)
     if "properties" in axes:
-        sorted_idx = _dispatch.argsort_metadata_values(
+        sorted_idx = _dispatch.argsort_labels_values(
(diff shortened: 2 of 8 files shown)


python/equistore-operations/equistore/operations/sort.py
```diff
@@ -0,0 +1,183 @@
...
+    return TensorMap(tensor.keys, blocks)
+
+
+def sort_block(
+    block: TensorBlock,
+    block_dims: List[str] = None,
+    sort_block_dims_names: bool = True, # sort each block's labels' names lexicographical
+    sort_block_dims_values: bool = True, # sort each block's values' according to the labels
+    descending: bool = False,
+    stable: bool = False
```
Luthaf, line 86: We should not need `stable` here since this is always applied on Labels, which do not contain duplicated rows.

python/equistore-operations/equistore/operations/sort.py
```diff
@@ -0,0 +1,183 @@
...
+                block=block, block_dims=block_dims, sort_metadata=sort_metadata, descending=descending, stable=stable
+            )
+        )
+    return TensorMap(tensor.keys, blocks)
+
+
+def sort_block(
+    block: TensorBlock,
+    block_dims: List[str] = None,
+    sort_block_dims_names: bool = True, # sort each block's labels' names lexicographical
```
Luthaf, line 83: I don't see a use case for `sort_block_dims_names = True`, and it looks like it will make everything more complex. Do we really need it?

python/equistore-operations/equistore/operations/sort.py
```diff
@@ -0,0 +1,183 @@
...
+    sort_block_dims_values: bool = True, # sort each block's values' according to the labels
+    descending: bool = False,
+    stable: bool = False
+) -> TensorBlock:
+    """
+    TODO
+    """
+    if block_dims is None:
+        block_dims = ["samples", "components", "properties"]
```
Luthaf, line 92: We should validate that `block_dims` only contains allowed values

Luthaf: Also, I don't see the point of sorting `components`, since each component can only contain a single column (and I would be fine enforcing that this column is sorted in equistore-core, I can see no reason for it not to be).

If this is about re-ordering components, I would be against since components carry more meaning (in particular, gradients and values components are linked)

python/equistore-operations/equistore/operations/sort.py
agoscinski: I was thinking about the equal check considering permutations. If the names are in a different order the Label entries have also a different order. We can say that this operations sorts everything always and remove the option, but we also need to consider somewhere two TensorBlocks having same labels with different order in the names.

python/equistore-operations/equistore/operations/sort.py
agoscinski: yes!

python/equistore-operations/equistore/operations/sort.py
agoscinski: In this code each component is sorted separately (within the component). So the components order is not changed. I also intuitively would not do it, since this is outside of the scope of the Labels object.

python/equistore-operations/equistore/operations/sort.py
Luthaf: That's a fair point. I think I would prefer having name re-ordering be part of a different operation, closer to the metadata renaming one.

python/equistore-operations/equistore/operations/sort.py
Luthaf: >  I also intuitively would not do it, since this is outside of the scope of the Labels object.

What's `it` here? Sorting them, or enforcing that they are sorted? We already enforce some limitation on the Labels used for components (the limitations are checked when constructing a TensorBlock, not when constructing a Labels), so we could easily add more limitations.

https://github.com/lab-cosmo/equistore/blob/b9bdcf5b433b17746a8a12fc3a0ba65295d5a55c/equistore-core/src/blocks.rs#L87-L104

agoscinski: I wouldn't sort them in this function in the sense changing the order between components (first component remains first component).

I dont have an opinion about enforcing sorting on components in equistore-core. Seems to me a bit arbitrary without any context.

python/equistore-operations/equistore/operations/sort.py
agoscinski: I can split it into three functions
`sort_metadata_names`
`sort_metadata_entries`
`sort` does everything (I would find this convenient)

agoscinski: After a small discussion with @Luthaf We decided to only keep the `sort_metadata_entries` functionality in `sort` and implement in another PR a `reorder_labels` function that allows to reorder the label columns/names. Because the lexicographic ordering of the names is not really the important feature , but rather having an option to reorder the names (to compare two blocks with different order in the names)

agoscinski:
I had problem to sort the rows
- argsort cannot sort properly a list of Tuple[int], it transforms it to a dense array and then sort it along one axis.
- `sorted` is not well supported by TorchScript (see `PR COMMENT` in current version) using keys

My only idea at the moment is to remap each label column to a dense range [0, N], then a set of Label columns `[0, N_1]_i, ..., [0, N_n]` can be mapped to one index with the sum `sum_i N_max^n_i`.

Luthaf:
What's the issue with adding a `sort` operation to `_dispatch`; and using either `np.sort` or `torch.sort` to sort the values of labels? Both can return the indices of the sorted values in the initial array, which can then be used to re-order the data.

agoscinski:
You need to sort over the rows treating each row as aggregated element in the sorting. torch.sort and numpy.sort do not work with like this (for numpy there would be a workaround with structured arrays). Try to sort `torch.sort(torch.tensor([(2,1),(1,2)]))` with `dim=0` and `dim=1`. The rows or columns are sorted completed independently.

I have now a workaround with TorchScript: `sorted` works, but not the `key` argument which I used to sort a range according to the value of the metadata elements. This is basically an argsort on Tuples. Now I append that an enumeration as last element to the tuple.  Since it is the last element and all Tuples are unique it does not influence the sorting. Then I discovered that tuples are also not TorchScriptable, so I switched to Lists, that seems to also work

Now evrything should be compatible with TorchScript, but I need #309 and #310 merged first

agoscinski:
> Instead of hacking around bad support for sorting in TorchScript, should we add Labels.sort(return_indices=False) -> (Labels, indices) in core, and expose it to Python & TorchScript? This feels much cleaner ...

I think these 10 lines of code are super acceptable. It was just hard to figure out what way is supported https://github.com/lab-cosmo/equistore/blob/988e49c3b876b3cad1096715edebc5c1fa874c66/python/equistore-operations/equistore/operations/_dispatch.py#L400-L410

I dont think dealing with Labels on CPU and GPU when implementing sorting in Rust and Torch API will result in something more understandable than these 10 lines of code

Luthaf:
That's fair enough, although I don't think the code above handles labels values on GPU at all (since everything happens with `List[int]`, which live on CPU).

jwa7:
What's the status of this PR? I would find these functions useful!

agoscinski:
It just needs tests, and format fixes. Sorry am a bit overwhelmed at the moment. Feel free to finish this feature. Or to use my current solution I use in my code

```python
def sort_tm(tm):
    blocks = []
    for _, block in tm.items():
        values = block.values

        samples_values = block.samples.values
        sorted_idx = native_list_argsort([tuple(row.tolist()) for row in block.samples.values])
        samples_values = samples_values[sorted_idx]
        values = values[sorted_idx]

        components_values = []
        for i, component in enumerate(block.components):
            component_values = component.values
            sorted_idx = native_list_argsort([tuple(row.tolist()) for row in component.values])
            components_values.append( component_values[sorted_idx] )
            values = np.take(values, sorted_idx, axis=i+1)

        properties_values = block.properties.values
        sorted_idx = native_list_argsort([tuple(row.tolist()) for row in block.properties.values])
        properties_values = properties_values[sorted_idx]
        values = values[..., sorted_idx]

        blocks.append(
            TensorBlock(
                values=values,
                samples=Labels(values=samples_values, names=block.samples.names),
                components=[Labels(values=components_values[i], names=component.names) for i, component in enumerate(block.components)],
                properties=Labels(values=properties_values, names=block.properties.names)
            )
        )
    return TensorMap(keys=tm.keys, blocks=blocks)
def native_list_argsort(native_list):
    return sorted(range(len(native_list)), key=native_list.__getitem__)
```

agoscinski:
Will continue on this after the renaming

agoscinski:
I made two comments about design decision in the code, marked with `PR COMMENT`

python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -610,3 +610,95 @@ def where(array):
         return np.where(array)
     else:
         raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def take(array, indices, axis: int):
+    """
+    See torch take_along_dim or numpy take as reference.
```
Luthaf, line 617:
```suggestion
    See :py:func:`torch.take_along_dim` or :py:func:`numpy.take` as reference.
```

Luthaf: you mention take_along_dim, but then use index_select. Any reason why?

python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -610,3 +610,95 @@ def where(array):
...
+    if isinstance(array, TorchTensor):
+        return torch.index_select(array, dim=axis, index=indices)
+    elif isinstance(array, np.ndarray):
+        return np.take(array, indices, axis=axis)
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+# PR COMMENT I implemented argsort because I thought I need it, but it turned out not
+#            but I would nevertheless keep it, might be useful for something else
```
Luthaf, line 639: I would not keep code we don't need around, but you can post this function as a comment to the PR so we can find it easily later if we need it

python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -610,3 +610,95 @@ def where(array):
...
+    end and since metadata is unique, we do not affect the sorted order.
+
+    We are highly limited by TorchScript with the sorting functions: torch.argsort does
+    not work because it cannot sort over the rows considering each row as one aggregated
+    element The ´´reverse´´ keyword argument is only supported by TorchScript using sort
+    like l.sort(reverse=...), but not using sorted(reverse=...) the ``key`` keyword
+    argument is not supported by TorchScript at all
+
+    :param metadata_values: numpy.array or torch.Tensor :param reverse: if true, order
+    is descending
```
Luthaf, line 683:
```suggestion
    :param metadata_values: numpy.array or torch.Tensor 
    :param reverse: if true, order is descending
```

python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -610,3 +610,95 @@ def where(array):
...
+        if descending:
+            return np.flip(idx_ascending, axis=axis)
+        else:
+            return idx_ascending
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def argsort_metadata_values(metadata_values, reverse: bool = False):
+    """
```
Luthaf, line 671:
```suggestion
    """
    Similar to np.argsort, but intended for labels values only. This will sort the rows of the labels values.
```

python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -610,3 +610,95 @@ def where(array):
...
+
+        if descending:
+            return np.flip(idx_ascending, axis=axis)
+        else:
+            return idx_ascending
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def argsort_metadata_values(metadata_values, reverse: bool = False):
```
Luthaf, line 670: If my understanding is correct (see function summary below), then we can also do something like this
```suggestion
def argsort_labels_values(labels: Labels, reverse: bool = False):
```

python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -610,3 +610,95 @@ def where(array):
...
+    We are highly limited by TorchScript with the sorting functions: torch.argsort does
+    not work because it cannot sort over the rows considering each row as one aggregated
+    element The ´´reverse´´ keyword argument is only supported by TorchScript using sort
+    like l.sort(reverse=...), but not using sorted(reverse=...) the ``key`` keyword
+    argument is not supported by TorchScript at all
+
+    :param metadata_values: numpy.array or torch.Tensor :param reverse: if true, order
+    is descending
+
+    :return: indices corresponding to the sorted values in py:obj:`metadata_values`
```
Luthaf, line 685:
```suggestion
    :return: indices corresponding to the sorted values in ``metadata_values``
```

python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -610,3 +610,95 @@ def where(array):
...
+        )
+        return torch.argsort(idx, dim=-1, descending=reverse)
+    elif isinstance(metadata_values, np.ndarray):
+        list_tuples: List[List[int]] = metadata_values.tolist()
+        for i in range(len(metadata_values)):
+            list_tuples[i].append(i)
+        list_tuples.sort(reverse=reverse)
+        return np.array(list_tuples)[:, -1]
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
```
Luthaf, line 704: I don't love this implementation. As you mention, it is very hacky and can fail in a lot of ways. Should we do a `Labels.sort(reverse=...) -> (Labels, indices)` in metatensor-core instead? Like we did the union/intersections/...

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -0,0 +1,237 @@
...
+                    f"components' or 'properties' but contains '{axis}'"
+                )
+        axes_list = axes
+    else:
+        raise TypeError("input paramater 'axes' should be of type str or list of str")
+
+    result_block = _sort_single_block(block, axes_list, descending)
+
+    # issue #68
+    # for parameter, gradient in block.gradients().items():
```
Luthaf, line 168:
```suggestion
```

Not sure what this means? Where you referring to https://github.com/lab-cosmo/metatensor/pull/309?

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -0,0 +1,237 @@
...
+    return result_block
+
+
+def sort(
+    tensor: TensorMap,
+    axes: Union[str, List[str]] = "all",
+    descending: bool = False,
+) -> TensorMap:
+    """
+    Sorts each block separately, see :py:func:`sort_block` for more information
```
Luthaf, line 187: I would add a short summary of what the function does before this sentence.

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -0,0 +1,237 @@
...
+    block: TensorBlock,
+    axes: Union[str, List[str]] = "all",
+    descending: bool = False,
+) -> TensorBlock:
+    """
+    Rearanges the values of an block according to the order given by the sorted metadata
+    of the given axes.
+
+    This function creates copies of the metadata on the CPU to sort the metadata.  For
+    that each row have to be interpreted as tuple which is not supported by torch.
```
Luthaf, line 74:
```suggestion
    This function creates copies of the metadata on the CPU to sort the metadata.
```

users of the function should not have to care about this (but this could be a comment inside the function, or we can implement sort in metatensor-core)

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -0,0 +1,237 @@
...
+
+    :param axes: axis of array to argsort :param stable: if true, the order of duplicate
+    elements stays the same an the array :param descending: if false, the order is
+    ascending :return: sorted tensor block
```
Luthaf, line 78:
```suggestion
    :param axes: axis of array to argsort 
    :param stable: if true, the order of duplicate elements stays the same an the array 
    :param descending: if false, the order is ascending 
    
    :return: sorted tensor block
```

Luthaf: Do we actually support `stable` here?

python/metatensor-operations/metatensor/operations/_dispatch.py
agoscinski: ah thanks for the catch, forgot to change the doc, `take_along_dim` did not work with TorchScript (had some assumptions on the shape of input and indices)

python/metatensor-operations/metatensor/operations/_dispatch.py
agoscinski: okay

python/metatensor-operations/metatensor/operations/_dispatch.py
agoscinski: I would like to change it in another PR. I can reference here an issue. but I don't think I can add this core function in any time soon. But yes that would be probably the best solution. Maybe it is better to split the API into a `Labels.sort(Labels) -> Labels` and `Labels.argsort(Labels) -> indices` function, but I don't have strong opinions and we can move this discussion into an issue.

python/metatensor-operations/metatensor/operations/sort.py
agoscinski: no, because everything is unique stable does not make sense, forgot to remove it from the doc, thanks for the catch

python/metatensor-operations/metatensor/operations/sort.py
agoscinski: yes, exactly I forgot about this, adapted

python/metatensor-operations/metatensor/operations/sort.py
agoscinski: changed to
```
    Sort the tensor map according to the key values and the blocks for each
    specified axis in :param axes: according to the label values in the axis.

    Each block is sorted separately, see :py:func:`sort_block` for more information
```
----
Is the 
```
:py:func:`sort_block`
```
actually working? I don't know sphinx so well and I saw that you removed it for a class in another place

agoscinski (commented):
thanks for the catches, the PR is so long that I forgot to adapt the doc at several places.

```python
def argsort(array, axis=-1, descending: bool = False, stable: bool = False):
    """
    See torch argsort as reference.

    Because numpy and torch have different APIs we went for the more limited one
    in torch. Numpy supports more options for sorting that are not here available.

    :param array: the array that is argsorted
    :param axis: axis of array to argsort
    :param stable: if true, the order of duplicate elements stays the same an the array
    :param descending: if false, the order is ascending

    :return: indices of the sorted array wrt. the unsorted one
    """
        return torch.argsort(array, dim=axis, descending=descending, stable=stable)
        if stable:
            idx_ascending = np.argsort(array, axis=axis, kind="stable")
        else:
            idx_ascending = np.argsort(array, axis=axis, kind="quicksort")

        if descending:
            return np.flip(idx_ascending, axis=axis)
        else:
            return idx_ascending
    else:
```

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -0,0 +1,232 @@
...
+    ...     properties=Labels(["n", "l"], np.array([[2, 0], [3, 0], [1, 0]])),
+    ... )
+    >>> print(block)
+    TensorBlock
+        samples (3): ['structures', 'centers']
+        components (): []
+        properties (3): ['n', 'l']
+        gradients: None
+    >>> block_sorted_stepwise = metatensor.sort_block(block, axes=["properties"])
+    >>> # properties are are sorted which are columns in this case
```
Luthaf, line 96:
```suggestion
    >>> # properties (last dimension of the array) are sorted
```

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -0,0 +1,232 @@
...
+    >>> block_sorted_stepwise.values
+    array([[2, 0, 1],
+           [5, 3, 4],
+           [8, 6, 7]])
+    >>> block_sorted_stepwise = metatensor.sort_block(
+    ...     block_sorted_stepwise, axes=["samples"]
+    ... )
+    >>> # samples are are sorted which are rows in this case
```
Luthaf, line 104:
```suggestion
    >>> # samples (first dimension of the array) are sorted
```

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -0,0 +1,232 @@
...
+    >>> block_sorted_stepwise.values
+    array([[5, 3, 4],
+           [8, 6, 7],
+           [2, 0, 1]])
+    >>> sorted_block = metatensor.sort_block(block)
+    >>> np.all(
+    ...     sorted_block.values == block_sorted_stepwise.values
+    ... )  # doing everything at once and verify correctness
+    True
```
Luthaf, line 113:
```suggestion
    >>> # sorting both samples and properties at the same time
    >>> sorted_block = metatensor.sort_block(block)
    >>> np.all(
    ...     sorted_block.values == block_sorted_stepwise.values
    ... )
    True
```

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -0,0 +1,232 @@
...
+    ...     components=[],
+    ...     properties=Labels(["n", "l"], np.array([[2, 0], [3, 0], [1, 0]])),
+    ... )
+    >>> print(block)
+    TensorBlock
+        samples (3): ['structures', 'centers']
+        components (): []
+        properties (3): ['n', 'l']
+        gradients: None
+    >>> block_sorted_stepwise = metatensor.sort_block(block, axes=["properties"])
```
Luthaf, line 95:
```suggestion
    >>> # sorting axes one by one
    >>> block_sorted_stepwise = metatensor.sort_block(block, axes=["properties"])
```

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -0,0 +1,232 @@
...
+    >>> block_sorted_stepwise.values
+    array([[5, 3, 4],
+           [8, 6, 7],
+           [2, 0, 1]])
+    >>> sorted_block = metatensor.sort_block(block)
+    >>> np.all(
+    ...     sorted_block.values == block_sorted_stepwise.values
+    ... )  # doing everything at once and verify correctness
+    True
+    >>> # Example for sorting the gradient
```
Luthaf, line 114:
```suggestion
    >>> # This function can also sort gradients:
```

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -0,0 +1,232 @@
...
+    ...     ),
+    ... )
+    >>> sorted_grad_block = metatensor.sort_block(sorted_block)
+    >>> sorted_grad_block.gradient("g").samples == Labels.range("sample", 3)
+    True
+    >>> sorted_grad_block.gradient("g").properties == sorted_block.properties
+    True
+    >>> sorted_grad_block.gradient(
+    ...     "g"
+    ... ).values  # sorting of components can be clearly seen
```
Luthaf, line 131:
```suggestion
    >>> # the components (middle dimensions) are also sorted:
    >>> sorted_grad_block.gradient("g").values
```

Luthaf (commented):
The functions needs to be added to the API reference for the operations, but apart from a couple of doc comments this should be good to go!

agoscinski:
I put sort into the set operations in the doc. I checked my local of the doc and it the page looked okay.

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+def sort(
+    tensor: TensorMap,
+    axes: Union[str, List[str]] = "all",
+    descending: bool = False,
+) -> TensorMap:
+    """
+    Sort the tensor map according to the key values and the blocks for each
+    specified axis in :param axes: according to the label values in the axis.
```
Luthaf, line 183:
```suggestion
    specified axis in ``axes`` according to the label values along these axes.
```

this does not render on the docs

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -0,0 +1,231 @@
...
+
+    Each block is sorted separately, see :py:func:`sort_block` for more information
+
+    Note: This function duplicates metadata on the CPU for the purpose of sorting.
+
+    :param axes: axis of array to argsort
```
Luthaf, line 189:
```suggestion
    :param axes: axes to sort. The labels entries along these axes will be sorted in lexicographic order, and the arrays values will be reordered accordingly. Possible values are ``'samples'``, ``'components'``, ``'properties'`` and ``'all'`` to sort everything.
```

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -0,0 +1,231 @@
...
+    ... )
+    >>> metatensor.sort(tm)
+    TensorMap with 2 blocks
+    keys: species
+             0
+             1
+    """
+    blocks: List[TensorBlock] = []
+
+    sorted_idx = _dispatch.argsort_labels_values(tensor.keys.values, reverse=descending)
```
Luthaf, line 221: Should not sort the keys by default? and allow the user to select if they want to sort it or only sort the blocks

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -0,0 +1,231 @@
...
+    axes: Union[str, List[str]] = "all",
+    descending: bool = False,
+) -> TensorBlock:
+    """
+    Rearanges the values of a block according to the order given by the sorted metadata
+    of the given axes.
+
+    This function creates copies of the metadata on the CPU to sort the metadata.
+
+    :param axes: axis of array to argsort
```
Luthaf, line 73:
```suggestion
    :param axes: axes to sort. The labels entries along these axes will be sorted in lexicographic order, and the arrays values will be reordered accordingly. Possible values are ``'samples'``, ``'components'``, ``'properties'`` and ``'all'`` to sort everything.
```

Luthaf (commented):
just realized that we might not want to always sort the keys in `sort()`. What do you think?

agoscinski (commented):
I am not sure if this super useful. I think if someone requests it as a feature, we can add it. One can apply the`sort_block` on the blocks individually with the current implementation and get this effect, so its a functionality we in principle already cover, just not so convenient

Luthaf:
> I am not sure if this super useful. I think if someone requests it as a feature, we can add it. One can apply thesort_block on the blocks individually with the current implementation and get this effect, so its a functionality we in principle already cover, just not so conveniently

It would be quite surprising to me as a user that `sort(map, axes="samples")` also modifies the keys.

I'd be fine if we never sort the keys and don't have a `keys` options, but if we modify the keys I think we need to provide a way to disable it. I can take care of the code if you agree with the idea!

agoscinski:
> It would be quite surprising to me as a user that sort(map, axes="samples") also modifies the keys.

Okay I see your point. For me `axes` was never a name that I identified with the keys of the tensor map, but I just saw that we use `keys` as valid axis in the manipulate dimension operations.

python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -151,15 +138,28 @@ def eye_like(array, size: int):
...
+    """
+    if isinstance(like, TorchTensor):
+        return torch.tensor(bool_list, dtype=torch.bool, device=like.device)
+    elif isinstance(like, np.ndarray):
+        return np.array(bool_list).astype(bool)
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def int_array_like(int_list: List[int], like):
```
Luthaf, line 154: I changed this function to be closer to bool_array_like and use it in sort

agoscinski:
Looks good to me, should be just a change in the tests of the0asserted error message.

Luthaf:
I'll merge once CI is green!

PR #282, opened 2023-06-12
Fix dependency constrains for pre-release equistore-core/...

Luthaf:
This allows to build & install wheels in https://github.com/Luthaf/nightly-wheels

----
:books: Documentation preview :books:: https://equistore--282.org.readthedocs.build/en/282/


python/equistore-torch/setup.py
```diff
@@ -198,6 +215,16 @@ def git_extra_version():
...
 if __name__ == "__main__":
+    if os.path.exists("git_extra_version"):
+        # we are building from a sdist, without git available, but the git
+        # version was recorded in a git_extra_version file
+        with open("git_extra_version") as fd:
+            extra_version = fd.read()
+    else:
+        extra_version = git_extra_version()
+
+    version = "0.1.0" + extra_version
```
PicoCentauri, line 226: Can't we pull the version from somwhere else? I am a bit afraid that we forget all the spots when we bump the version.

PicoCentauri (commented):
Just a question if should have the specific version so many times across the library.

python/equistore-torch/setup.py
Luthaf: We could also have a `VERSION` file in the C++ `equistore-torch` folder, and use that. This way we can have the same version for C++ and Python. But if we want different version numbers for -core, -torch and -operations, we will need to update the version in multiple places anyway.

PicoCentauri: I think a single version is fine. If we start shipping different versions for -core, -torch and -operations this becomes maybe a bit more confusing.

Luthaf: I've added a VERSION file for equistore-torch version

PR #287, opened 2023-06-23
Implement Labels::union and Labels::intersection

Luthaf:
These function perform the corresponding set operation over Labels entries, and optionally return the mapping from the positions in the inputs to the position in the output.

Still TODO: 
- [x] update operations to use these functions where needed. 

----
:books: Documentation preview :books:: https://equistore--287.org.readthedocs.build/en/287/

### DIFF
--- a/equistore-core/src/labels.rs
+++ b/equistore-core/src/labels.rs
@@ -1,8 +1,10 @@
 #![allow(clippy::default_trait_access, clippy::module_name_repetitions)]
 
 use std::ffi::CString;
-use std::collections::{BTreeSet, HashMap};
-use std::collections::hash_map::Entry;
+use std::collections::BTreeSet;
+
+use hashbrown::HashMap;
+use hashbrown::hash_map::RawEntryMut;
 
 use smallvec::SmallVec;
 
@@ -93,12 +95,14 @@ impl LabelValue {
     }
 }
 
+type DefaultHasher = std::hash::BuildHasherDefault<ahash::AHasher>;
+
 /// Builder for `Labels`, this should be used to construct `Labels`.
 pub struct LabelsBuilder {
     // cf `Labels` for the documentation of the fields
-    names: Vec<String>,
+    names: Vec<ConstCString>,
     values: Vec<LabelValue>,
-    positions: HashMap<SmallVec<[LabelValue; 4]>, usize, ahash::RandomState>,
+    positions: HashMap<SmallVec<[LabelValue; 4]>, usize, DefaultHasher>,
 }
 
 impl LabelsBuilder {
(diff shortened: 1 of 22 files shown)


python/equistore-core/equistore/core/labels.py
```diff
@@ -456,6 +456,105 @@ def position(self, entry: Union[LabelsEntry, Sequence[int]]) -> Optional[int]:
...
+
+        return Labels._from_eqs_labels_t(output)
+
+    def intersection_and_mapping(
+        self, other: "Labels"
+    ) -> Tuple["Labels", np.ndarray, np.ndarray]:
+        """
+        Take the intersection of these :py:class:`Labels` with ``other``, and the
+        mapping from the positions of entries in the inputs and the positions of entries
+        in the output.
```
PicoCentauri, line 528: An example would be very useful here.

equistore-core/src/labels.rs
```diff
@@ -355,4 +466,82 @@ mod tests {
         let e = LabelsBuilder::new(vec!["not", "there", "not"]).err().unwrap();
         assert_eq!(e.to_string(), "invalid parameter: labels names must be unique, got 'not' multiple times");
     }
+
+    #[test]
+    fn union() {
```
PicoCentauri, line 471: Are you handling the case when one of the labels is an empty label?

python/equistore-core/equistore/core/labels.py
```diff
@@ -456,6 +456,105 @@ def position(self, entry: Union[LabelsEntry, Sequence[int]]) -> Optional[int]:
         else:
             return None
 
+    def union(self, other: "Labels") -> "Labels":
+        """Take the union of these :py:class:`Labels` with ``other``."""
```
PicoCentauri, line 460: I would add an see also here to link to the `union_and_mapping` if one also wants to have the mapping. Same for the intersection.

PicoCentauri (changes requested):
The function logic and the wrapping code looks good. We might want to work a bit on the description of the functions. Also, some examples for the Python methods might be helpful. 

Especially for the distinction between `union` and `union_and_mapping` which is a bit unclear. I am also not super happy with the name `union_and_mapping` and  `intersection_and_mapping`.

python/equistore-core/equistore/core/labels.py
Luthaf: done!

equistore-core/src/labels.rs
Luthaf: added a test!

PicoCentauri (approved):
Awesome! Merge when you like.

PR #292, opened 2023-07-03
Fix numpy deprecation warning when using array with single entry as a scalar

Luthaf:
This also removes some `try/catch` blocks, which will help with TorchScript support (#280 is another step in this direction).

----
:books: Documentation preview :books:: https://equistore--292.org.readthedocs.build/en/292/

### DIFF
--- a/python/equistore-operations/tests/multiply.py
+++ b/python/equistore-operations/tests/multiply.py
@@ -266,14 +266,11 @@ class TestMultiply(unittest.TestCase):
         keys = Labels(names=["key_1", "key_2"], values=np.array([[0, 0], [1, 0]]))
         A = TensorMap(keys, [block_1, block_2])
         B = 5.1
-        C = np.array([5.1])
 
         tensor_sum = equistore.multiply(A, B)
-        tensor_sum_array = equistore.multiply(A, C)
         tensor_result = TensorMap(keys, [block_res_1, block_res_2])
 
         self.assertTrue(equistore.allclose(tensor_result, tensor_sum))
-        self.assertTrue(equistore.allclose(tensor_result, tensor_sum_array))
 
     def test_self_multiply_error(self):
         block_1 = TensorBlock(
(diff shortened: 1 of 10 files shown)


python/equistore-operations/tests/multiply.py
```diff
@@ -266,14 +266,11 @@ def test_self_multiply_scalar_gradient(self):
         keys = Labels(names=["key_1", "key_2"], values=np.array([[0, 0], [1, 0]]))
         A = TensorMap(keys, [block_1, block_2])
         B = 5.1
-        C = np.array([5.1])
```
DavideTisi, line 269: is this "feature" removed now?

python/equistore-operations/tests/pow.py
```diff
@@ -111,14 +111,10 @@ def test_self_pow_scalar_gradient(self):
 
         keys = Labels(names=["key_1", "key_2"], values=np.array([[0, 0], [1, 0]]))
 
-        C = np.array([2.0])
```
DavideTisi, line 114: same as multiply

python/equistore-operations/tests/subtract.py
```diff
@@ -264,14 +264,11 @@ def test_self_subtract_scalar_gradient(self):
         keys = Labels(names=["key_1", "key_2"], values=np.array([[0, 0], [1, 0]]))
         A = TensorMap(keys, [block_1, block_2])
         B = -5.1
-        C = np.array([-5.1])
```
DavideTisi, line 267: same as before

DavideTisi (commented):
Everything is fine, did we remove the feature that we can use `np.array([[scalar]])` ?

Luthaf:
> Everything is fine, did we remove the feature that we can use np.array([[scalar]]) ?


Yes, because it is deprecated in numpy 1.25:

```
>>> import numpy as np
>>> float(np.array([3.2]))
<stdin>:1: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
3.2
```

PR #295, opened 2023-07-05
Add `Labels::column` to access a single column in the values

Luthaf:
And make `Labels::__getitem__` use this instead of view.

This mean that code trying to compute some gradient will go from looking like this:

```py
data = some_function(block.values[gradient.samples["sample"].values.reshape(-1)] )
```

to this, which is much closer to the previous version

```py
data = some_function(block.values[gradient.samples["sample"]] )
```

----
:books: Documentation preview :books:: https://equistore--295.org.readthedocs.build/en/295/

### DIFF
--- a/equistore-torch/include/equistore/torch/labels.hpp
+++ b/equistore-torch/include/equistore/torch/labels.hpp
@@ -78,6 +78,10 @@ public:
         return values_;
     }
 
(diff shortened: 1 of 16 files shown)


jwa7 (approved):
Nice thanks! Makes the labels API much cleaner :)

Issue #133, opened 2023-02-21
The path to TorchScript


Luthaf:
Here is what I currently see as the path to enabling the use of equistore within TorchScript. 

For reference, TorchScript is a small, Python-like language that is implemented inside the torch library (in C++). It can be extended with new operations/classes in Python (if the new operation/class only use the TorchScript-compatible subset of Python); or in C++ (completely custom operations/classes are possible). 

To enable using equistore with TorchScript, we need to add a couple of custom classes (TensorMap, TensorBlock, Labels) to TorchScript, and we need to implement these in C++ (since we need to call into the core equistore library). This mean there will be two version of each class available from Python: the current, pure Python ones, and the TorchScript ones. See point 4 below for how to handle this. 

Once we have the basic classes and member functions (such as `keys_to_properties`) available from TorchScript, we'll have to make sure everything in `equistore.operations` is available from 

### 0) Clean separation of the core shared library

A pre-requisite to the TorchScript version of equistore (and the ability to use this version together with rascaline) is to ensure only one version of the core functions/structs exists, by splitting it into a standalone shared library that can then be loaded by both rascaline and TorchScript.

- [x] pull request: #85

### 1) Write the new TorchScript classes

We can then start writing C++ code to create new classes inheriting from `torch::CustomClassHolder`, and registering these classes with TorchScript. This means re-wrapping all classes and exposing all functions in a TorchScript-compatible way (e.g. all data uses `torch::Tensor`, …). This will first leave in a separate package (equistore_torch) in this repository. This can happen as a set of PR to a long-lived branch if we want to parallelize the work

- [X] create a long-lived branch with basic infrastructure to build and install the code #233 
- [x] implement the new classes:  #263
    - [x] Labels
    - [x] Implement [`equistore::DataArrayBase`](https://github.com/lab-cosmo/equistore/blob/3327e1b5ddd3ee9c0441e3b82a36c67044c6933f/include/equistore.hpp#L573-L582) for `torch::Tensor`
    - [x] TensorBlock/Proxy for gradients
    - [x] TensorMap
- [x] merge the long-lived branch

- [x] Implement serialization for the torch version, this requires the same core functions as #94.

### 2) Change how `Labels` works

[...]

PR #298, opened 2023-07-11
Save and load TensorMap from in-memory buffers

### LINKED ISSUE #94: Implement Python's pickle protocol for TensorMap
This would give us a better integration with the overall Python ecosystem. 

This should be done using `equistore.io.load`/`equistore.io.save`, but the functions themselves might require a bit of rework to write to (resp. load from) in-memory buffer instead of writing to (resp. loading from) the filesystem.

Luthaf:
Part of #133 
Fix #94

This allow implementing the Pickle protocol without writing to a file first, and to reduce the number of copies when loading data/sharing data with other python processes.

----
:books: Documentation preview :books:: https://equistore--298.org.readthedocs.build/en/298/

### DIFF
--- a/equistore-torch/src/misc.cpp
+++ b/equistore-torch/src/misc.cpp
@@ -1,0 +1,41 @@
+#include <torch/torch.h>
+
+#include <equistore.hpp>
+#include <torch/types.h>
+
+#include "equistore/torch/array.hpp"
+#include "equistore/torch/misc.hpp"
+
+using namespace equistore_torch;
+
+
+eqs_status_t equistore_torch::details::create_torch_array(
+    const uintptr_t* shape_ptr,
+    uintptr_t shape_count,
+    eqs_array_t* array
+) {
+    auto sizes = std::vector<int64_t>();
+    for (size_t i=0; i<shape_count; i++) {
+        sizes.push_back(static_cast<int64_t>(shape_ptr[i]));
+    }
+
+    auto options = torch::TensorOptions().device(torch::kCPU).dtype(torch::kF64);
+    auto tensor = torch::zeros(sizes, options);
+
+    auto cxx_array = std::unique_ptr<equistore::DataArrayBase>(new TorchDataArray(tensor));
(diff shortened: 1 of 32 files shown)


equistore-torch/src/misc.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include <equistore.hpp>
+#include <torch/types.h>
+
+#include "equistore/torch/array.hpp"
+#include "equistore/torch/misc.hpp"
+
+using namespace equistore_torch;
+
+
+eqs_status_t equistore_torch::details::create_torch_array(
```
Luthaf, line 16: Note for myself: check that we enclose this function in a try-catch block in the C++ API, or do it here

PicoCentauri (commented):
I took a look. The c-part is a bit hard to follows for me but it looks reasonable.

equistore-torch/src/misc.cpp
Luthaf: Done

equistore-torch/src/register.cpp
```diff
@@ -180,15 +181,38 @@ TORCH_LIBRARY(equistore, m) {
...
-        //     },
-        //     // __setstate__
-        //     [](std::string state) -> torch::intrusive_ptr<TorchCalculator> {
-        //         // TODO
-        //     })
+        .def_pickle(
+            // __getstate__
+            [](const TorchTensorMap& self) -> torch::Tensor {
+                auto buffer = equistore::TensorMap::save_buffer(self->as_equistore());
+                // move the buffer to the heap so it can escape this function
```
agoscinski, line 188: can we explain this a bit more? It was nontrivial for me that from_blob does not take ownership, that is why we need to put on the heap.

```suggestion
                // move the buffer to the heap so it can escape this function
                // since from_blob will not take ownership over the data
```

agoscinski (commented):
Looks good. Just small comment

equistore-torch/src/register.cpp
Luthaf: done!

PR #299, opened 2023-07-13
Add operations for manipulating meta data

PicoCentauri:
This PR adds an operations to manipulate Labels as well as the Labels columns stored in TensorMaps as `keys`, `samples` or `properties`.

----
:books: Documentation preview :books:: https://equistore--299.org.readthedocs.build/en/299/

### DIFF
--- a/equistore-torch/include/equistore/torch/labels.hpp
+++ b/equistore-torch/include/equistore/torch/labels.hpp
@@ -79,6 +79,18 @@ public:
         return values_;
     }
 
+    /// Append a new dimension with the given `name` and `values`  to the end of these `Labels`
+    TorchLabels append(std::string name, torch::Tensor values);
+
+    /// Insert a new dimension with the given `name` and `values` before `index` in these `Labels`
+    TorchLabels insert(int64_t index, std::string name, torch::Tensor values);
+
+    /// Remove `name` from the dimensions of these `Labels`
+    TorchLabels remove(std::string name);
+
+    /// Rename `old_name` dimensions in these Labels to `new_name`
+    TorchLabels rename(std::string old_name, std::string new_name);
+
     /// Move the values for these Labels to the given `device`
     void to(torch::Device device);
 
--- a/equistore-torch/src/labels.cpp
+++ b/equistore-torch/src/labels.cpp
@@ -265,6 +265,73 @@ LabelsHolder LabelsHolder::to_owned() const {
     }
 }
 
+TorchLabels LabelsHolder::append(std::string name, torch::Tensor values) {
+    return this->insert(this->size(), std::move(name), std::move(values));
+}
+
+
+TorchLabels LabelsHolder::insert(int64_t index, std::string name, torch::Tensor values) {
+    auto names = this->names();
+
+    auto it = std::begin(names) + index;
+    names.insert(it, name);
+
+    if (values.sizes().size() != 1) {
+        C10_THROW_ERROR(
+            ValueError,
+            "`values` must be a 1D tensor"
+        );
+    }
+
+    auto old_values = this->values();
+    auto first = old_values.index({torch::indexing::Slice(), torch::indexing::Slice(0, index)});
+    auto second = old_values.index({torch::indexing::Slice(), torch::indexing::Slice(index)});
+
+    auto new_values = torch::hstack({first, values.reshape({values.size(0), 1}), second});
+
+    return torch::make_intrusive<LabelsHolder>(std::move(names), std::move(new_values));
+}
+
+
+TorchLabels LabelsHolder::remove(std::string name) {
+    auto names = this->names();
+
+    auto it = std::find(std::begin(names), std::end(names), name);
+    if (it == std::end(names)) {
+        C10_THROW_ERROR(
+            ValueError,
+            "'" + name + "' not found in the dimensions of these Labels"
+        );
+    }
+
+    auto column_index = it - std::begin(names);
+    names.erase(it);
+
+    auto values = this->values();
+    auto first = values.index({torch::indexing::Slice(), torch::indexing::Slice(0, column_index)});
+    auto second = values.index({torch::indexing::Slice(), torch::indexing::Slice(column_index + 1)});
+    auto new_values = torch::hstack({first, second});
+
+    return torch::make_intrusive<LabelsHolder>(std::move(names), std::move(new_values));
+}
+
+
+TorchLabels LabelsHolder::rename(std::string old_name, std::string new_name) {
+    auto names = this->names();
+
+    auto it = std::find(std::begin(names), std::end(names), old_name);
+    if (it == std::end(names)) {
+        C10_THROW_ERROR(
+            ValueError,
+            "'" + old_name + "' not found in the dimensions of these Labels"
+        );
+    }
+    auto column_index = it - std::begin(names);
+
+    names[column_index] = new_name;
+    return torch::make_intrusive<LabelsHolder>(std::move(names), std::move(this->values()));
+}
+
 void LabelsHolder::to(torch::Device device) {
     // first move the values
     values_ = values_.to(device);
--- a/python/equistore-core/equistore/core/labels.py
+++ b/python/equistore-core/equistore/core/labels.py
@@ -436,6 +436,144 @@ class Labels:
         """
         return self._values
 
+    def append(self, name: str, values: np.ndarray) -> "Labels":
+        """Append a new dimension to the end of the :py:class:`Labels`.
+
+        :param name: name of the new dimension
+        :param values: 1D array of values for the new dimension
+
+        >>> import numpy as np
+        >>> from equistore import Labels
+        >>> label = Labels("foo", np.array([[42]]))
+        >>> label
+        Labels(
+            foo
+            42
+        )
+        >>> label.append(name="bar", values=np.array([10]))
+        Labels(
+            foo  bar
+            42   10
+        )
+        """
+        return self.insert(index=len(self), name=name, values=values)
+
+    def insert(self, index: int, name: str, values: np.ndarray) -> "Labels":
+        """Insert a new dimension before ``index`` in the :py:class:`Labels`.
+
+        :param index: index before the new dimension is inserted
+        :param name: name of the new dimension
+        :param values: 1D array of values for the new dimension
+
+        >>> import numpy as np
+        >>> from equistore import Labels
+        >>> label = Labels("foo", np.array([[42]]))
+        >>> label
+        Labels(
+            foo
+            42
+        )
+        >>> label.insert(0, name="bar", values=np.array([10]))
+        Labels(
+            bar  foo
+            10   42
+        )
+        """
+        new_names = self.names.copy()
+        new_names.insert(index, name)
+
+        if not isinstance(values, np.ndarray):
+            raise ValueError("`values` must be a numpy ndarray")
+
+        if len(values.shape) != 1:
+            raise ValueError("`values` must be a 1D array")
+
+        new_values = np.insert(self.values, index, values, axis=1)
+
+        return Labels(names=new_names, values=new_values)
+
+    def remove(self, name: str) -> "Labels":
+        """Remove ``name`` from the dimensions of the :py:class:`Labels`.
+
+        Removal can only be performed if the resulting :py:class:`Labels` instance will
+        be unique.
+
+        :param name: name to be removed
+        :raises ValueError: if the name is not present.
+
+        >>> import numpy as np
+        >>> from equistore import Labels
+        >>> label = Labels(["foo", "bar"], np.array([[42, 10]]))
+        >>> label
+        Labels(
+            foo  bar
+            42   10
+        )
+        >>> label.remove(name="bar")
+        Labels(
+            foo
+            42
+        )
+
+        If the new :py:class:`Labels` is not unique an error is raised.
+
+        >>> from equistore import EquistoreError
+        >>> label = Labels(["foo", "bar"], np.array([[42, 10], [42, 11]]))
+        >>> label
+        Labels(
+            foo  bar
+            42   10
+            42   11
+        )
+        >>> try:
+        ...     label.remove(name="bar")
+        ... except EquistoreError as e:
+        ...     print(e)
+        ...
+        invalid parameter: can not have the same label value multiple time: [42] is already present at position 0
+        """  # noqa E501
+        if name not in self.names:
+            raise ValueError(f"'{name}' not found in the dimensions of these Labels")
+
+        new_names = self.names.copy()
+        new_names.remove(name)
+
+        index = self.names.index(name)
+        new_values = np.delete(self.values, index, axis=1)
+
+        return Labels(names=new_names, values=new_values)
+
+    def rename(self, old: str, new: str) -> "Labels":
+        """Rename the ``old`` dimension to ``new`` in the :py:class:`Labels`.
+
+        :param old: name to be replaced
+        :param new: name after the replacement
+        :raises ValueError: if old is not present.
+
+        >>> import numpy as np
+        >>> from equistore import Labels
+        >>> label = Labels("foo", np.array([[42]]))
+        >>> label
+        Labels(
+            foo
+            42
+        )
+        >>> label.rename("foo", "bar")
+        Labels(
+            bar
+            42
+        )
+        """
+
+        if old not in self.names:
+            raise ValueError(f"'{old}' not found in the dimensions of these Labels")
+
+        names = self.names.copy()
+        index = names.index(old)
+        names[index] = new
+
+        return Labels(names=names, values=self.values)
+
     def to(self, device):
         """
         Move the values for these Labels to the given ``device``.
--- a/python/equistore-core/tests/labels.py
+++ b/python/equistore-core/tests/labels.py
@@ -111,6 +111,44 @@ def test_custom_constructors():
         Labels.range(0, 1)
 
 
+def test_dimensions_manipulation():
+    label = Labels("foo", np.array([[42]]))
+
+    # Labels.insert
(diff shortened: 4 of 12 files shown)


PicoCentauri:
Sorry for the many force pushes. The three functions should be there now.

python/equistore-operations/equistore/operations/manipulate_metadata.py
```diff
@@ -0,0 +1,335 @@
...
+def _remove_from_list(a: List, b: List):
+    """Remove all items in ``b`` from ``a``"""
+    for item in b:
+        if item not in a:
+            raise ValueError(f"'{item}' not in list")
+        a.remove(item)
+
+
+def append_metadata(
+    tensor: TensorMap, name: str, values: Union[float, np.ndarray], axis: str
```
PicoCentauri, line 53: I currently only allow for only one new label name. We can extend this to allow also a list of names. However, this will makes verifying the values more complex...

jwa7: Yes I think a single name is fine for now

jwa7 (changes requested):
Thanks for this @PicoCentauri ! 

3 things on this:

1. I think the name of the functions would be better as specifying explicitly that we are adding/removing columns. 
2. As a user I would like to have the option to insert the column at an index of my choosing. Or, at the very least, have also the option to prepend as well as append (though controlling the index is preferred).
3. From an UX-consistency point of view I would order the args `axis, name, values`. It fits better with other operations that use the `axis, values` or `axis, labels` ordering.

Can I therefore suggest a function with a signature like:
```py
def insert_metadata_column(
    tensor: TensorMap, axis: str, name: str, values: Union[float, np.ndarray], col_idx: Optional[int] = None
) -> TensorMap:
```
where for the default behaviour with `col_idx=None` the column is appended to the end, i.e. as is currently done by this function.

Thoughts?

python/equistore-operations/equistore/operations/__init__.py
```diff
@@ -44,6 +44,7 @@
 )
 from .join import join
 from .lstsq import lstsq
+from .manipulate_metadata import append_metadata, rename_metadata
```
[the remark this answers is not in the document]
PicoCentauri: Hmm But then we can not use them as toplevel imports because calling `equistore.append` is not clear...

PicoCentauri:
I renamed the functions and introduced the Labels methods that I needed.

Let me know what you think.

python/equistore-core/equistore/core/labels.py
```diff
@@ -303,6 +303,141 @@ def range(name: str, end: int) -> "Labels":
             values=np.arange(end, dtype=np.int32).reshape(-1, 1),
         )
 
+    def append(self, name: str, values: np.ndarray) -> "Labels":
+        """Append a new column to the end.
+
+        :param name: name of the new column
+        :param values: 1D array of values for the new column
```
PicoCentauri, line 310: I demand this to be a 1D array because we are stacking a new column to a 2D array. If you think this seems inconsistent w.r.t. to creating a new Labels object (where we demand a 2D array as input).

python/equistore-core/equistore/core/labels.py
```diff
@@ -303,6 +303,144 @@ def range(name: str, end: int) -> "Labels":
             values=np.arange(end, dtype=np.int32).reshape(-1, 1),
         )
 
+    def append(self, name: str, values: np.ndarray) -> "Labels":
```
Luthaf, line 306: If we add function to the Labels in equistore-core, we need to add the same ones in equistore-torch. 

The API is nicer if we add member functions, but that's more work to do. I don't have strong feelings between adding functions twice or keeping them in equistore-operations

PicoCentauri: Yes, I will add them to `equistore-torch` if we agree on the API. I think the fit quite good here. I only haven't touched this to avoid changing lines twice...

python/equistore-operations/equistore/operations/manipulate_labels_column.py
```diff
@@ -0,0 +1,310 @@
...
+    >>> equistore.remove_labels_column(tensor, axis="keys", name="extra")
+    TensorMap with 1 blocks
+    keys: keys
+           0
+    """
+    _check_axis(axis)
+
+    keys = tensor.keys
+    if axis == "keys":
+        keys = keys.remove(name=name)
```
jwa7, line 204: If we're left with an empty Labels object, an error is created when using this function but the message is not helpful. I'd suggest we check for this and give a better error message:

```suggestion
        keys = keys.remove(name=name)
        if len(keys) == 0:
            raise ValueError("Cannot remove the only key name present.")
```

python/equistore-operations/equistore/operations/manipulate_labels_column.py
```diff
@@ -0,0 +1,310 @@
...
+
+    blocks = []
+    for block in tensor:
+        samples = block.samples
+        properties = block.properties
+
+        if axis == "samples":
+            samples = samples.remove(name)
```
jwa7, line 212: And maybe here too with the samples and properties
```suggestion
            samples = samples.remove(name)
            if len(samples) == 0:
                raise ValueError("Cannot remove the only samples name present.")
```

python/equistore-operations/equistore/operations/manipulate_labels_column.py
```diff
@@ -0,0 +1,310 @@
...
+        elif axis == "properties":
+            properties = properties.remove(name)
```
jwa7, line 214:
```suggestion
            properties = properties.remove(name)
            if len(properties) == 0:
                raise ValueError("Cannot remove the only properties name present.")
```

python/equistore-operations/tests/manipulate_labels_column.py
```diff
@@ -0,0 +1,242 @@
...
+
+    for block in new_tensor:
+        assert_equal(block.properties.values[:, -1], values)
+
+        for _, gradient in block.gradients():
+            assert gradient.properties.names[-1] == "foo"
+            assert_equal(gradient.properties.values[:, -1], values)
+
+
+def test_append_unknonw_axis():
```
jwa7, line 82:
```suggestion
def test_append_unknown_axis():
```

python/equistore-operations/tests/manipulate_labels_column.py
```diff
@@ -0,0 +1,242 @@
...
+
+    for block in new_tensor:
+        assert_equal(block.properties.values[:, index], values)
+
+        for _, gradient in block.gradients():
+            assert gradient.properties.names[index] == "foo"
+            assert_equal(gradient.properties.values[:, index], np.arange(80))
+
+
+def test_insert_unknonw_axis():
```
jwa7, line 140:
```suggestion
def test_insert_unknown_axis():
```

python/equistore-operations/tests/manipulate_labels_column.py
```diff
@@ -0,0 +1,242 @@
...
+        new="foo",
+    )
+    assert new_tensor.property_names[0] == "foo"
+
+    for block in new_tensor:
+        for _, gradient in block.gradients():
+            assert gradient.properties.names[0] == "foo"
+
+
+def test_rename_unknonw_axis():
```
jwa7, line 187:
```suggestion
def test_rename_unknown_axis():
```

python/equistore-operations/tests/manipulate_labels_column.py
```diff
@@ -0,0 +1,242 @@
...
+        name="extra",
+    )
+    assert new_tensor.property_names == ["properties"]
+
+    for block in new_tensor:
+        for _, gradient in block.gradients():
+            assert gradient.properties.names == ["properties"]
+
+
+def test_remove_unknonw_axis():
```
jwa7, line 226:
```suggestion
def test_remove_unknown_axis():
```

python/equistore-operations/tests/manipulate_labels_column.py
```diff
@@ -0,0 +1,242 @@
...
+    for block in new_tensor:
+        for parameter, gradient in block.gradients():
+            if parameter == "positions":
+                gradient_sample_names = ["sample", "foo", "atom"]
+            elif parameter == "cell":
+                gradient_sample_names = ["sample"]
+            assert gradient.samples.names == gradient_sample_names
+
+
+def test_renames_properties():
```
jwa7, line 171:
```suggestion
def test_rename_properties():
```

jwa7 (commented):
Super super useful thanks @PicoCentauri !!

PicoCentauri:
Thanks for the review @jwa7 ! I put the length check in the beginning of the function. We don't have to do this for every block.

I will now do the port of the Labels method for equsitore-torch.

equistore-torch/include/equistore/torch/labels.hpp
```diff
@@ -79,6 +79,18 @@ class EQUISTORE_TORCH_EXPORT LabelsHolder: public torch::CustomClassHolder {
         return values_;
     }
 
+    /// Append a new dimension to the end
```
Luthaf, line 82:
```suggestion
    /// Append a new dimension with the given `name` and `values`  to the end of these `Labels`
```

equistore-torch/include/equistore/torch/labels.hpp
```diff
@@ -79,6 +79,18 @@ class EQUISTORE_TORCH_EXPORT LabelsHolder: public torch::CustomClassHolder {
...
+    TorchLabels append(std::string name, torch::Tensor values);
+
+    /// Insert a new dimension before index
```
Luthaf, line 85:
```suggestion
    /// Insert a new dimension with the given `name` and `values` before `index` in these `Labels`
```

equistore-torch/include/equistore/torch/labels.hpp
```diff
@@ -79,6 +79,18 @@ class EQUISTORE_TORCH_EXPORT LabelsHolder: public torch::CustomClassHolder {
...
+    TorchLabels insert(int64_t index, std::string name, torch::Tensor values);
+
+    /// Remove `name` from the dimensions of these Labels
```
Luthaf, line 88:
```suggestion
    /// Remove `name` from the dimensions of these `Labels`
```

equistore-torch/include/equistore/torch/labels.hpp
```diff
@@ -79,6 +79,18 @@ class EQUISTORE_TORCH_EXPORT LabelsHolder: public torch::CustomClassHolder {
...
+    TorchLabels remove(std::string name);
+
+    /// Remove `name` from the dimensions of these Labels
```
Luthaf, line 91:
```suggestion
    /// Rename `old_name` dimensions in these Labels to `new_name`
```

equistore-torch/src/labels.cpp
```diff
@@ -265,6 +265,73 @@ LabelsHolder LabelsHolder::to_owned() const {
...
+            ValueError,
+            "`values` must be a 1D tensor"
+        );
+    }
+
+    auto old_values = this->values();
+    auto first = old_values.index({torch::indexing::Slice(), torch::indexing::Slice(0, index)});
+    auto second = old_values.index({torch::indexing::Slice(), torch::indexing::Slice(index)});
+
+    auto new_values = torch::hstack({first, values.reshape({1, static_cast<int64_t>(values.size(0))}), second});
```
Luthaf, line 290:
```suggestion
    auto new_values = torch::hstack({first, values.reshape({values.size(0), 1}), second});
```

`values.size(0)` should already be a 64-bit int. Also, I think the shape was the wrong way around?

python/equistore-core/equistore/core/labels.py
```diff
@@ -436,6 +436,144 @@ def values(self) -> np.ndarray:
         """
         return self._values
 
+    def append(self, name: str, values: np.ndarray) -> "Labels":
+        """Append a new dimension to the end.
```
Luthaf, line 440:
```suggestion
        """Append a new dimension to the end of the :py:class:`Labels`.
```

python/equistore-core/equistore/core/labels.py
```diff
@@ -436,6 +436,144 @@ def values(self) -> np.ndarray:
...
+        >>> label.append(name="bar", values=np.array([10]))
+        Labels(
+            foo  bar
+            42   10
+        )
+        """
+        return self.insert(index=len(self), name=name, values=values)
+
+    def insert(self, index: int, name: str, values: np.ndarray) -> "Labels":
+        """Insert a new dimension before ``index``.
```
Luthaf, line 462:
```suggestion
        """Insert a new dimension before ``index`` in the :py:class:`Labels`.
```

python/equistore-core/equistore/core/labels.py
```diff
@@ -436,6 +436,144 @@ def values(self) -> np.ndarray:
...
+
+        if len(values.shape) != 1:
+            raise ValueError("`values` must be a 1D array")
+
+        new_values = np.insert(self.values, index, values, axis=1)
+
+        return Labels(names=new_names, values=new_values)
+
+    def remove(self, name: str) -> "Labels":
+        """Remove ``name`` from the dimensions of these Labels.
```
Luthaf, line 496:
```suggestion
        """Remove ``name`` from the dimensions of the :py:class:`Labels`.
```

python/equistore-core/equistore/core/labels.py
```diff
@@ -436,6 +436,144 @@ def values(self) -> np.ndarray:
...
+        new_names = self.names.copy()
+        new_names.remove(name)
+
+        index = self.names.index(name)
+        new_values = np.delete(self.values, index, axis=1)
+
+        return Labels(names=new_names, values=new_values)
+
+    def rename(self, old: str, new: str) -> "Labels":
+        """Rename the ``old`` dimension to ``new``.
```
Luthaf, line 547:
```suggestion
        """Rename the ``old`` dimension to ``new`` in the :py:class:`Labels`.
```

python/equistore-core/tests/labels.py
```diff
@@ -111,6 +111,44 @@ def test_custom_constructors():
         Labels.range(0, 1)
 
 
+def test_custom_manipulation():
```
Luthaf, line 114:
```suggestion
def test_dimensions_manipulation():
```

python/equistore-operations/equistore/operations/__init__.py
```diff
@@ -44,6 +44,12 @@
 )
 from .join import join
 from .lstsq import lstsq
+from .manipulate_labels_column import (
+    append_labels_column,
+    remove_labels_column,
+    rename_labels_column,
+    insert_labels_column,
```
Luthaf, line 51: Naming-wise do we want `labels_column` or `_dimension`? Both are fine for me, with a slight preference for `_dimension`.


```suggestion
    append_dimension,
    remove_dimension,
    rename_dimension,
    insert_dimension,
```

python/equistore-torch/equistore/torch/documentation.py
```diff
@@ -298,6 +298,111 @@ def __ne__(self, other: "Labels") -> bool:
         different values)
         """
 
+    def append(self, name: str, values: torch.Tensor) -> "Labels":
+        """Append a new dimension to the end.
```
Luthaf, line 302: Same suggestions here as in equistore-core

python/equistore-torch/tests/labels.py
```diff
@@ -384,6 +384,41 @@ def test_intersection():
     assert torch.all(second_mapping == torch.LongTensor([-1, 0, -1]))
 
 
+def test_manipulation():
```
Luthaf, line 387:
```suggestion
def test_dimensions_manipulation():
```

python/equistore-torch/tests/labels.py
```diff
@@ -384,6 +384,41 @@ def test_intersection():
...
+    label = Labels("foo", torch.tensor([[42]]))
+
+    # # Labels.insert
```
Luthaf, line 390:
```suggestion
    # Labels.insert
```

Luthaf (commented):
Look good! The main discussion point is about the operations naming!

python/equistore-operations/equistore/operations/__init__.py
PicoCentauri: I also like the `_dimensions` a bit more. @jwa7  seems to be in favor of columns?

equistore-torch/src/labels.cpp
PicoCentauri: yes it is the other way around. But why is this also working...

python/equistore-operations/equistore/operations/__init__.py
Luthaf: these functions operate on TensorMap, the dimensions of which are always stored in Labels (each name/column in the Labels is a dimension of the high-dimensional sparse array we represent with a TensorMap). So for me the `_labels` part is redundant =)

equistore-torch/src/labels.cpp
Luthaf: not sure, maybe some torch magic/dispatch?

python/equistore-operations/equistore/operations/__init__.py
PicoCentauri: Fine for me, lets just wait if @jwa7 is also fine.

jwa7: @PicoCentauri @Luthaf Yep sounds good to me!

PicoCentauri:
I did the renaming. If you two @Luthaf and @jwa7 are happy we can merge this.

jwa7 (approved):
Looks good, thanks @PicoCentauri ! :)

PicoCentauri:
@Luthaf if you are also happy you can merge this.

Luthaf:
Looks good!

PR #301, opened 2023-07-17
Remove 'tensor' of a joined TensorMap if possible

PicoCentauri:
Adds the functionality to `join` to remove the `"tensor"` column name while joining TensorMaps.

Due to torch-script I can not use a try-catch block like

```python
try:
    return remove_labels_column(tensor_joined, name="tensor", axis=axis)
except EquistoreError:
    return tensor_joined
```

Instead, I have to check the uniqueness on the Python level before calling `remove_labels_column`. If someone has an idea to get around the his I am happy to change this.

### DIFF
--- a/python/equistore-operations/equistore/operations/join.py
+++ b/python/equistore-operations/equistore/operations/join.py
@@ -7,29 +7,184 @@ import numpy as np
 from equistore.core import Labels, TensorBlock, TensorMap
 
 from ._utils import _check_same_keys
+from .manipulate_dimension import remove_dimension
 
 
-def join(tensors: List[TensorMap], axis: str):
+def _disjoint_tensor_labels(tensors: List[TensorMap], axis: str) -> bool:
+    """Checks if all labels in a list of TensorMaps are disjoint.
+
+    We have to perform a check from all tensors to all others to ensure it
+    they are "fully" disjoint.
+    """
+    for i_tensor, first_tensor in enumerate(tensors[:-1]):
+        for second_tensor in tensors[i_tensor + 1 :]:
+            for key, first_block in first_tensor.items():
+                second_block = second_tensor.block(key)
+                if axis == "samples":
+                    first_labels = first_block.samples
+                    second_labels = second_block.samples
+                elif axis == "properties":
+                    first_labels = first_block.properties
+                    second_labels = second_block.properties
+
+                if len(first_labels.intersection(second_labels)):
+                    return False
+
+    return True
+
+
+def join(
+    tensors: List[TensorMap], axis: str, remove_tensor_name: bool = False
+) -> TensorMap:
     """Join a sequence of :py:class:`TensorMap` along an axis.
 
-    The ``axis`` parameter specifies the type join. For example, if
-    ``axis='properties'`` it will be the `tensor_maps` will be joined along the
+    The ``axis`` parameter specifies the type of joining. For example, if
+    ``axis='properties'`` the tensor maps in `tensors` will be joined along the
     `properties` dimension and for ``axis='samples'`` they will be the along the
-    samples dimension.
-
-    ``join`` will create an additional label `tensor` specifiying the original index in
-    the list of `tensor_maps`.  If `sample`/`property` names are not the same in all
-    `tensor_maps` they will be unified with a general name ``"property"``.
+    `samples` dimension.
 
     :param tensors:
         sequence of :py:class:`TensorMap` for join
     :param axis:
         A string indicating how the tensormaps are stacked. Allowed
         values are ``'properties'`` or ``'samples'``.
+    :param remove_tensor_name:
+        Remove the extra ``tensor`` dimension from labels if possible. See examples
+        above for the case where this is applicable.
 
     :return tensor_joined:
         The stacked :py:class:`TensorMap` with more properties or samples
         than the input TensorMap.
+
+    Examples
+    --------
+    Possible clashes of the meta data like ``samples``/``properties`` will be resolved
+    by one of the three following strategies:
+
+    1. If Labels names are the same, the values are unique and
+       ``remove_tensor_name=True`` we keep the names and join the values
+
+       >>> import numpy as np
+       >>> import equistore
+       >>> from equistore import Labels, TensorBlock, TensorMap
+
+       >>> values = np.array([[1.1, 2.1, 3.1]])
+       >>> samples = Labels("sample", np.array([[0]]))
+
+       Define two disjoint :py:class:`Labels`.
+
+       >>> properties_1 = Labels("n", np.array([[0], [2], [3]]))
+       >>> properties_2 = Labels("n", np.array([[1], [4], [5]]))
+
+       >>> block_1 = TensorBlock(
+       ...     values=values,
+       ...     samples=Labels.single(),
+       ...     components=[],
+       ...     properties=properties_1,
+       ... )
+       >>> block_2 = TensorBlock(
+       ...     values=values,
+       ...     samples=Labels.single(),
+       ...     components=[],
+       ...     properties=properties_2,
+       ... )
+
+       >>> tensor_1 = TensorMap(keys=Labels.single(), blocks=[block_1])
+       >>> tensor_2 = TensorMap(keys=Labels.single(), blocks=[block_2])
+
+       joining along the properties leads
+
+       >>> joined_tensor = equistore.join(
+       ...     [tensor_1, tensor_2], axis="properties", remove_tensor_name=True
+       ... )
+       >>> joined_tensor[0].properties
+       Labels(
+           n
+           0
+           2
+           3
+           1
+           4
+           5
+       )
+
+       If ``remove_tensor_name=False`` There will be an extra dimension ``tensor``
+       added
+
+       >>> joined_tensor = equistore.join(
+       ...     [tensor_1, tensor_2], axis="properties", remove_tensor_name=False
+       ... )
+       >>> joined_tensor[0].properties
+       Labels(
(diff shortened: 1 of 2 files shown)


python/equistore-operations/equistore/operations/join.py
```diff
@@ -126,7 +160,22 @@ def join(tensors: List[TensorMap], axis: str):
...
+            new_names = np.delete(block.samples, -1, axis=1)
+        elif axis == "properties":
+            n_values = len(block.properties)
+            # "tensor" name is inserted at the begininng of the Labels dimensions
+            new_names = np.delete(block.properties, 0, axis=1)
+
+        n_values_unique = len(np.unique(new_names, axis=0))
+
+        if n_values != n_values_unique:
+            return tensor_joined
```
jwa7, line 179: Perhaps `unique_metadata` can be used here?

```suggestion
    if len(equistore.unique_metadata(tensor_joined, axis=axis, name="tensor")) > 1:
        return tensor_joined
```

jwa7: Hmm apologies maybe not: see https://github.com/lab-cosmo/equistore/pull/314#discussion_r1273274348

python/equistore-operations/equistore/operations/join.py
```diff
@@ -17,9 +18,42 @@ def join(tensors: List[TensorMap], axis: str):
...
+        labels_1 = ["n"], [0, 2, 3]
+        labels_2 = ["n"], [1, 4, 5]
+
+       will lead to::
+
+        joined_labels = ["n"],
+                        [0, 2, 3, 1, 4, 5]
+
+    2. If Labels names are the same but the values are not unique, a new dimension
+       ``"tensor"`` is added to the names::
```
agoscinski, line 36:
```suggestion
       ``"tensor"`` is added at the beginning to the names which determines from which TensorMap the label entry originally comes from ::
```

maybe a comment on where the new name is added, always at the beginning?

python/equistore-operations/equistore/operations/join.py
```diff
@@ -17,9 +18,42 @@ def join(tensors: List[TensorMap], axis: str):
...
-    ``join`` will create an additional label `tensor` specifiying the original index in
-    the list of `tensor_maps`.  If `sample`/`property` names are not the same in all
-    `tensor_maps` they will be unified with a general name ``"property"``.
+    Possible clashes of the meta data like samples/properties names are resolved by one
+    of the three following strategies:
+
+    1. If Labels names are the same and the values are unique, we keep the names and
+       join the values::
+
+        labels_1 = ["n"], [0, 2, 3]
```
agoscinski, line 27: This could be actually super nice doctests. not enforcing this, but would be nice

python/equistore-operations/equistore/operations/join.py
```diff
@@ -17,9 +18,42 @@ def join(tensors: List[TensorMap], axis: str):
...
     samples dimension.
 
-    ``join`` will create an additional label `tensor` specifiying the original index in
-    the list of `tensor_maps`.  If `sample`/`property` names are not the same in all
-    `tensor_maps` they will be unified with a general name ``"property"``.
+    Possible clashes of the meta data like samples/properties names are resolved by one
+    of the three following strategies:
+
+    1. If Labels names are the same and the values are unique, we keep the names and
+       join the values::
```
agoscinski, line 25: why two double colons at the end? is this related to sphinx?

python/equistore-operations/equistore/operations/join.py
```diff
@@ -17,9 +18,42 @@ def join(tensors: List[TensorMap], axis: str):
...
+
+        labels_1 = ["n"], [0, 2, 3]
+        labels_2 = ["n"], [0, 2]
+
+       will lead to::
+
+            joined_labels = ["tensor", "n"],
+                            [[0, 0], [0, 2], [0, 3], [1, 0], [1, 2]]
+
+    3. If Labels names are different we change the names to ("tensor", "property"). This
```
agoscinski, line 46:
```suggestion
    3. If Labels names are different we change the names to ("tensor", "property"). The column "property" corresponds to the ith label entry TensorMap specified by tensor. This
```

python/equistore-operations/equistore/operations/join.py
```diff
@@ -126,7 +160,22 @@ def join(tensors: List[TensorMap], axis: str):
     else:
         tensor_joined = tensor.keys_to_properties("tensor")
 
-    # TODO: once we have functions to manipulate meta data we can try to
-    # remove the `tensor` label entry after joining.
-
-    return tensor_joined
+    # Remove "tensor" Labels dimension if a new Labels of all blocks would be unique
+    # Due to toch script we can use a try-except block but have to check in advance if
```
agoscinski, line 164:
```suggestion
    # Due to toch script we cant use a try-except block but have to check in advance if
```

agoscinski (commented):
If we input a TensorMap with a label column name "tensor" or "property" can we raise an error already in this function? Something like 
```
raise ValueError(f"join cannot be applied on TensorMaps with a column name 'tensor' in the TensorMap keys or in targeted axis {axis}. Please rename the corresponding column first using ....")
```

python/equistore-operations/equistore/operations/join.py
```diff
@@ -17,9 +18,42 @@ def join(tensors: List[TensorMap], axis: str):
...
+    3. If Labels names are different we change the names to ("tensor", "property"). This
+       case is only supposed to happen when joining in the property dimension, hence the
+       choice of names::
+
+            labels_1 = ["a"], [0, 2, 3]
+            labels_2 = ["b", "c"], [[0, 0], [1, 2]]
+
+       will lead to::
+
+            joined_labels = ["tensor", "property"],
```
agoscinski, line 55: What do you think about using a notation for "private" variables in python for metadata column names that can exclusively only used equistore for operations "_eqs_tensor" and "_eqs_property". To be honest, I don't have any other ideas of applications than this function, but it could be already here useful.

python/equistore-operations/equistore/operations/join.py
```diff
@@ -17,9 +18,42 @@ def join(tensors: List[TensorMap], axis: str):
     `properties` dimension and for ``axis='samples'`` they will be the along the
```
agoscinski, line 18:
```
`axis='properties'`` it will be the `tensor_maps` will be joined along the
```
->
```
`axis='properties'`` the tensor maps in `tensors` will be joined along the
```

python/equistore-operations/equistore/operations/join.py
PicoCentauri: Yes this renders the next paragraph as code.

python/equistore-operations/equistore/operations/join.py
PicoCentauri: Yes, I though the same. I think I can do it!

python/equistore-operations/equistore/operations/join.py
PicoCentauri: Hmm I think private names will make everything harder to follow.

python/equistore-operations/equistore/operations/join.py
PicoCentauri: Yes as we talked yesterday I will try to use Labels::intersection.

python/equistore-operations/equistore/operations/join.py
```diff
@@ -7,6 +7,7 @@
 from equistore.core import Labels, TensorBlock, TensorMap
 
 from ._utils import _check_same_keys
+from .manipulate_dimension import remove_dimension
 
 
 def join(tensors: List[TensorMap], axis: str):
```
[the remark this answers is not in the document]
PicoCentauri: Yes it is more `concatenate` but I would say we keep it like this and can rename the bunch when we have more operations.

PicoCentauri:
I added an new boolean parameter `remove_tensor_name` to control if the tensor dimension should be removed or not. We can also leave this out and completly and only put a comment that the user can remove the tensor using the `equistore.remove_dimension`. To be honest I am more and more in favor to do this. What do the reviewers think @jwa7, @agoscinski  and @Luthaf 

Also I added your the examples you want @agoscinski. Hope they are fine and understandable. I tried to keep the code to create the tensors as small as possible.

jwa7 (commented):
I like the having the bool option to remove, but think we should have a default to False. Thoughts?

python/equistore-operations/equistore/operations/join.py
agoscinski: They actually do this in pandas for their merge function
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.merge.html
> **indicator: bool or str, default False**
>
> If True, **adds a column to the output DataFrame called “_merge” with information on the source of each row**. The column can be given a different name by providing a string argument. The column will have a Categorical type with the value of “left_only” for observations whose merge key only appears in the left DataFrame, “right_only” for observations whose merge key only appears in the right DataFrame, and “both” if the observation’s merge key is found in both DataFrames.

agoscinski:
> I added an new boolean parameter remove_tensor_name to control if the tensor dimension should be removed or not. We can also leave this out and completly and only put a comment that the user can remove the tensor using the equistore.remove_dimension. To be honest I am more and more in favor to do this. What do the reviewers think @jwa7, @agoscinski and @Luthaf

Given the discussion in https://github.com/lab-cosmo/equistore/issues/315 we will break the API of join anyway, I dont think it matters much what we choose now. Let's just go with an option and set it to False as @jwa7 suggests it.

agoscinski (approved):

PR #302, opened 2023-07-17
Delete function to get references on rvalues in C++ API

Luthaf:
Otherwise we can get dangling references since the rvalue immediately goes out of scope. 

Previously, this code would silently result in an empty `sizes` vector:

```cpp
auto sizes = std::vector<size_t>();
for (auto size: block.values().shape()) {
    sizes.push_back(size);
}
```

Now, the code above fails to compile, and forces to put `block.values()` in a temporary.

----
:books: Documentation preview :books:: https://equistore--302.org.readthedocs.build/en/302/

### DIFF
--- a/equistore-core/include/equistore.hpp
+++ b/equistore-core/include/equistore.hpp
@@ -262,7 +262,7 @@ public:
     /// double value = array(2, 3, 1);
     /// ```
     template<typename ...Args>
-    T operator()(Args... args) const {
+    T operator()(Args... args) const & {
         auto index = std::array<size_t, sizeof... (Args)>{static_cast<size_t>(args)...};
         if (index.size() != shape_.size()) {
             throw Error(
@@ -281,7 +281,7 @@ public:
     /// array(2, 3, 1) = 5.2;
     /// ```
     template<typename ...Args>
-    T& operator()(Args... args) {
+    T& operator()(Args... args) & {
         if (is_const_) {
             throw Error("This NDArray is const, can not get non const access to it");
         }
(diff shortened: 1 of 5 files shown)


Luthaf:
(I found this issue while updating rascaline-torch)

equistore-core/include/equistore.hpp
```diff
@@ -1630,6 +1630,24 @@ class TensorBlock final {
         return array;
     }
 
+    /// Release the `eqs_block_t` pointer corresponding to this `TensorBlock`.
```
agoscinski, line 1633: this seems to be just moved, can we put it back to create less noise or is this intentionally?

equistore-core/include/equistore.hpp
```diff
@@ -1630,6 +1630,24 @@ class TensorBlock final {
...
+                "block since it is a view inside a TensorMap"
+            );
+        }
+        auto ptr = block_;
+        block_ = nullptr;
+        is_view_ = false;
+        return ptr;
+    }
+
+    friend class TensorMap;
```
agoscinski, line 1650: is this from debugging? I dont see any reason why this is needed for this PR

agoscinski (commented):
I think the problem that this PR solves is well described here https://artificial-mind.net/blog/2019/09/22/delete-your-rvalue-ref-assignments

I imagine something like
```
NDArray array{{1});
array(0) = 2;
```
would create a dangling references before this PR.

equistore-core/include/equistore.hpp
Luthaf: This is intentional: this function returns a pointer, and is also susceptible to the same issue. But in this case, this is no longer needed in the public interface (this was here to implement rascaline-torch before we had equistore-torch) and so is now moved to the private section of the class.

equistore-core/include/equistore.hpp
Luthaf: TensorMap uses the private function to move blocks around

equistore-core/include/equistore.hpp
agoscinski: Ah okay missed that it went to private

agoscinski (approved):
I see thank for the clarifications and the simple example. The link I posted seems to show a different problem even though it suggests the same solution.

PR #304, opened 2023-07-18
Torch: get a single block without selection in `TensorMap.block`

Luthaf:
Previously, `tensor.block()` would fail even if there is a single block in the TensorMap


agoscinski (approved):
Looks solid. PR enables usage in equistore TorchScript API
```python
block = tm.block()
```
when there is only one block. A test has been added.

PR #305, opened 2023-07-18
Add equistore-torch to equistore's extras_require

Luthaf:
This will allow to use `pip install equistore[torch]`

----
:books: Documentation preview :books:: https://equistore--305.org.readthedocs.build/en/305/


PicoCentauri (commented):
Very useful.  We can not use the `pyproject.toml` since we have to do some logic with the `uuid` and this doesn't work outside of the `setup.py`, right?

Luthaf:
> We can not use the pyproject.toml since we have to do some logic with the uuid and this doesn't work outside of the setup.py, right?

Exactly. Pip and friend defaults are not the best for monorepos

PR #310, opened 2023-07-22
Make dispatch TorchScript compatible

agoscinski:
Contains:
* torch.Tensor type is always checked first so TorchScript never interprets the np.ndarray is instance check
* _check_all_same_type with its type argument is not TorchScript compatible so it was split into two functions

TODO: Need to do tests, probably I am missing more things that are not TorchScript compatible

----
:books: Documentation preview :books:: https://equistore--310.org.readthedocs.build/en/310/


Luthaf:
I think this is ready for review, tests will happen as we move more and more operations to TorchScript

PicoCentauri (approved):

PR #313, opened 2023-07-25
Python CI adjustments

PicoCentauri:
Some minor additions to the CI and the pyproject file.

----
:books: Documentation preview :books:: https://equistore--313.org.readthedocs.build/en/313/

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -20,11 +20,11 @@ jobs:
           - os: ubuntu-20.04
             python-version: "3.7"
           - os: ubuntu-20.04
-            python-version: "3.10"
+            python-version: "3.11"
           - os: macos-11
-            python-version: "3.10"
+            python-version: "3.11"
           - os: windows-2019
-            python-version: "3.10"
+            python-version: "3.11"
     steps:
       - uses: actions/checkout@v3
 
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.10"
+          python-version: "3.11"
 
       - name: install dependencies
         run: python -m pip install cibuildwheel
@@ -89,7 +89,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.10"
+          python-version: "3.11"
       - name: install dependencies
         run: python -m pip install wheel build
       - name: package vendored code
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -17,6 +17,11 @@ classifiers = [
     "Operating System :: Microsoft :: Windows",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
```
Luthaf, line 24: Do we really need to list all Python versions here? It feels to me like busy work to keep this updated for all our pyproject.toml (there are 3 other such files just in this repo)

.github/workflows/python-tests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           - os: ubuntu-20.04
             python-version: "3.7"
           - os: ubuntu-20.04
-            python-version: "3.10"
+            python-version: "3.11"
           - os: macos-11
             python-version: "3.10"
```
Luthaf, line 25: I would also update the Python version on macos & windows

pyproject.toml
PicoCentauri: Hmm yes I am not sure either. These are the tags which will be shown at `Classifiers` on PyPI. We could also leave this out since the minimal version is fixed as well.

.github/workflows/python-tests.yml
PicoCentauri: Fine for me. But we leave Python 3.10 for all the other CI tasks? lint, docs and wheels

PicoCentauri: Then lets bump all of them to be consistent and avoid probable weird error messages.

However, torch doesn't still not support `3.11`, right?

Luthaf: > However, torch doesn't still not support 3.11, right?

What makes you think so? They have wheels for cp311 both here https://download.pytorch.org/whl/torch/ and on on PyPI: https://pypi.org/project/torch/#files

Luthaf: Oh, this: https://github.com/lab-cosmo/equistore/actions/runs/5655781307/job/15321547016#step:8:169

We run some of the tests against torch 1.11, which does not support Python 3.11. We could do a Python version in the matrix as well for these

Luthaf (approved):

Luthaf:
Hmm, the code looks good but the test failure also seems legit. Maybe a different numpy version is the root cause?

PicoCentauri:
It is an abs test failing

```
tests/abs.py .F                                                          [  1%]
```

I had the same issue rarely locally, but it never appeared in CI until now...

PicoCentauri:
I tried to make the abs test more _explicit_ (removing the pytest fixtures, because they caused problems also in the past). Maybe this will help...

PR #314, opened 2023-07-25
More helpful error msg when using `remove_dimension`

jwa7:
When attempting to remove a dimension of a TensorMap's metadata, an unhelpful EquistoreError is raised if the dimension has non-redundant values. This PR checks within the `remove_dimension` method for non-redundancies in the named labels dimension and raises a more helpful error message.


python/equistore-operations/equistore/operations/manipulate_dimension.py
```diff
@@ -210,8 +216,12 @@ def remove_dimension(tensor: TensorMap, axis: str, name: str) -> TensorMap:
         properties = block.properties
 
         if axis == "samples":
+            if len(np.unique(samples.column(name))) > 1:
```
Luthaf, line 219: I don't think this check works: we could have a column with `-1, -1, -2, -2` but when removing it the values are still unique

jwa7: Hmm yeah good point. It seems that @PicoCentauri also is doing a similar check in `join` https://github.com/lab-cosmo/equistore/pull/301. Perhaps it would be good to have a utility function `def column_is_redundant(labels, name) -> bool:` that does the proper check and can be used both here and in `join`?

jwa7: Thoughts @Luthaf @PicoCentauri ?

PicoCentauri: Function could be useful, but I don't know if it should be public.

PicoCentauri:
What is the actual error message that is raised? Maybe it is more that we should improve the error message on the rust side...

python/equistore-operations/equistore/operations/manipulate_dimension.py
Luthaf: What's wrong with trying to remove it and then letting the exception bubble up? Potentially catching the root exception and throwing a new one with more context

Luthaf: Arf, I forgot about that ... I don't see a much better way than making the rust-side error clearer then =/

jwa7: Ok yeah I think that's a good alternative :)

PicoCentauri: Do you want to do it @jwa7 ?

If yes maybe #306 can help to find the rust files you want to look at.

Luthaf: The error comes from here: https://github.com/lab-cosmo/equistore/blob/139958001633368b1ffd725aca916084df724d29/equistore-core/src/labels.rs#L172-L174

We are validating entry by entry the new Labels in a LabelsBuilder, while computing and cache the position of each entry. If is it helpful, we could move the validation in another function.

jwa7: After a discussion with @PicoCentauri it seems that it is not so easy to modify the error message rust-side in a helpful way. We are going back to the idea of a separate function that checks for redundancy, maybe as a private function. I think this would just be a near-copy of the code in Philip's `join` PR, but generalised of course. THoughts @Luthaf ?

PicoCentauri:
I think the plan after some discussion is to add the example we do for `Labels.remove()`  also in `remove_dimensions`, right?

https://github.com/lab-cosmo/equistore/blob/040134f7ca86cfd82d37fe3d9798b400a52591ea/python/equistore-core/equistore/core/labels.py#L518-L534

Luthaf:
yes!

PicoCentauri:
I added the example

python/equistore-operations/equistore/operations/manipulate_dimension.py
```diff
@@ -197,7 +197,26 @@ def remove_dimension(tensor: TensorMap, axis: str, name: str) -> TensorMap:
     TensorMap with 1 blocks
     keys: key
            0
-    """
+
+    Removing a dimension can only be performed if the resulting :py:class:`Labels` will
+    be unque.
```
Luthaf, line 202:
```suggestion
    Removing a dimension can only be performed if the resulting :py:class:`Labels` will
    contain unique entries.
```

Luthaf (approved):

jwa7:
Thanks @PicoCentauri !

PR #316, opened 2023-07-25
Add labels.permute and permute_dimension

PicoCentauri:
Following #299 this add an operation for permuting dimensions/columns of labels and labels in tensors. This will be useful as for sorting the rows as well.  

----
:books: Documentation preview :books:: https://equistore--316.org.readthedocs.build/en/316/

### DIFF
--- a/equistore-torch/src/labels.cpp
+++ b/equistore-torch/src/labels.cpp
@@ -293,6 +293,39 @@ TorchLabels LabelsHolder::insert(int64_t index, std::string name, torch::Tensor 
 }
 
 
+TorchLabels LabelsHolder::permute(std::vector<int64_t> dimensions_indexes) {
+    auto names = this->names();
+
+    if (dimensions_indexes.size() != names.size()) {
+        C10_THROW_ERROR(
+            ValueError,
+            "the length of `dimensions_indexes` (" + std::to_string(dimensions_indexes.size()) + ") "
+            "does not match the number of dimensions in the Labels (" + std::to_string(names.size()) + ")"
+        );
+    }
+
+    auto new_names = std::vector<std::string>();
+
+    for (auto index : dimensions_indexes) {
+        if (index < 0) {
+            index += names.size();
+        }
+        if (index >= names.size()) {
+            C10_THROW_ERROR(
+                IndexError,
+                "out of range index "  + std::to_string(index) + " for labels dimensions (" +
+                std::to_string(names.size()) + ")"
+            );
+        }
+        new_names.push_back(names[index]);
+    }
+ 
+    auto new_values = this->values().index({torch::indexing::Slice(), torch::tensor(dimensions_indexes)});
+
+    return torch::make_intrusive<LabelsHolder>(std::move(new_names), std::move(new_values));
+}
+
+
 TorchLabels LabelsHolder::remove(std::string name) {
     auto names = this->names();
 
--- a/python/equistore-core/equistore/core/labels.py
+++ b/python/equistore-core/equistore/core/labels.py
@@ -492,6 +492,39 @@ class Labels:
 
         return Labels(names=new_names, values=new_values)
 
+    def permute(self, dimensions_indexes: List[int]) -> "Labels":
+        """Permute dimensions according to ``dimensions_indexes`` in the
+        :py:class:`Labels`.
+
+        :param dimensions_indexes: desired ordering of the dimensions
+        :raises ValueError: if length of ``dimensions_indexes`` does not match the
+            Labels length
+        :raises ValueError: if duplicate values are present in ``dimensions_indexes``
+
+        >>> import numpy as np
+        >>> from equistore import Labels
+        >>> label = Labels(["foo", "bar", "baz"], np.array([[42, 10, 3]]))
+        >>> label
+        Labels(
+            foo  bar  baz
+            42   10    3
+        )
+        >>> label.permute([2, 0, 1])
+        Labels(
+            baz  foo  bar
+             3   42   10
+        )
+        """
+        if len(dimensions_indexes) != len(self.names):
+            raise ValueError(
+                f"the length of `dimensions_indexes` ({len(dimensions_indexes)}) does "
+                f"not match the number of dimensions in the Labels ({len(self.names)})"
+            )
+
+        names = [self.names[d] for d in dimensions_indexes]
+
+        return Labels(names=names, values=self.values[:, dimensions_indexes])
+
     def remove(self, name: str) -> "Labels":
         """Remove ``name`` from the dimensions of the :py:class:`Labels`.
 
--- a/python/equistore-operations/equistore/operations/__init__.py
+++ b/python/equistore-operations/equistore/operations/__init__.py
@@ -46,9 +46,10 @@ from .join import join
 from .lstsq import lstsq
 from .manipulate_dimension import (
     append_dimension,
+    insert_dimension,
+    permute_dimensions,
     remove_dimension,
     rename_dimension,
-    insert_dimension,
 )
 from .multiply import multiply
 from .one_hot import one_hot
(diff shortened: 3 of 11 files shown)


equistore-torch/src/labels.cpp
```diff
@@ -293,6 +293,40 @@ TorchLabels LabelsHolder::insert(int64_t index, std::string name, torch::Tensor
...
+        );
+    }
+
+    auto new_names = std::vector<std::string>();
+
+    for (auto index : dims) {
+        if (index < 0) {
+            index += names.size();
+        }
+        new_names.push_back(names.at(index));
```
Luthaf, line 314: even if `at` will throw an exception, the corresponding error message will be quite bad. You could also use `C10_THROW_ERROR(IndexError, "...")` here to give a nicer error

equistore-torch/src/labels.cpp
```diff
@@ -293,6 +293,40 @@ TorchLabels LabelsHolder::insert(int64_t index, std::string name, torch::Tensor
...
+    }
+
+    if (std::set<std::string>(new_names.begin(), new_names.end()).size() != names.size()) {
+        C10_THROW_ERROR(
+            ValueError,
+            "duplicate dims are not allowed"
```
Luthaf, line 320:
```suggestion
            "duplicate dimensions are not allowed"
```

Also, should we do this check on the integer values instead, after normalizing `-1` to `size -1`?

python/equistore-core/equistore/core/labels.py
```diff
@@ -492,6 +492,40 @@ def insert(self, index: int, name: str, values: np.ndarray) -> "Labels":
 
         return Labels(names=new_names, values=new_values)
 
+    def permute(self, dims: List[int]) -> "Labels":
```
Luthaf, line 495: naming suggestions
```suggestion
    def permute(self, dimensions: List[int]) -> "Labels":
```

```suggestion
    def permute(self, dimensions_indexes: List[int]) -> "Labels":
```

```suggestion
    def permute(self, columns: List[int]) -> "Labels":
```

python/equistore-core/equistore/core/labels.py
```diff
@@ -492,6 +492,40 @@ def insert(self, index: int, name: str, values: np.ndarray) -> "Labels":
...
+        if len(dims) != len(self.names):
+            raise ValueError(
+                f"number of dimensions Labels ({len(self.names)}) does not match the "
+                f"length of the desired ordering of dimensions ({len(dims)})."
+            )
+
+        names = [self.names[d] for d in dims]
+
+        if len(set(names)) != len(names):
+            raise ValueError("duplicate dims are not allowed")
```
Luthaf, line 525:
```suggestion
            raise ValueError("duplicate dimensions are not allowed")
```

Luthaf: Also, do we need to check this anyway? Since the rust code will also do this check

python/equistore-operations/equistore/operations/manipulate_dimension.py
```diff
@@ -162,6 +172,76 @@ def insert_dimension(
     return TensorMap(keys=keys, blocks=blocks)
 
 
+def permute_dimension(tensor: TensorMap, axis: str, dims: List[int]) -> TensorMap:
+    """Permute dimensions of a :py:class:`equistore.Labels`.
+
+    For ``axis=="samples"`` gradients are not permuted.
```
Luthaf, line 178:
```suggestion
    For ``axis=="samples"`` gradients samples dimensions are not permuted.
```

python/equistore-operations/tests/manipulate_labels_column.py
```diff
@@ -80,7 +80,7 @@ def test_append_properties():
 
 
 def test_append_unknown_axis():
-    with pytest.raises(ValueError, match="foo is not a valid axis."):
+    with pytest.raises(ValueError, match="'foo' is not a valid axis."):
```
Luthaf, line 83: could you move the file to `manipulate_dimensions`?

python/equistore-core/equistore/core/labels.py
PicoCentauri: Ohh yes you are right!

python/equistore-core/equistore/core/labels.py
PicoCentauri: Lets go with the second one. Even though it is the longest. For me `dimensions` are more the actual Labels or the `names`.

python/equistore-operations/tests/manipulate_labels_column.py
PicoCentauri: Yes sure. I don't know why I didn't...

equistore-torch/src/labels.cpp
```diff
@@ -293,6 +293,40 @@ TorchLabels LabelsHolder::insert(int64_t index, std::string name, torch::Tensor
...
 
+TorchLabels LabelsHolder::permute(std::vector<int64_t> dimensions_indexes) {
+    auto names = this->names();
+
+    if (dimensions_indexes.size() != names.size()) {
+        C10_THROW_ERROR(
+            ValueError,
+            "number of dimensions Labels (" + std::to_string(names.size()) + ") does " +
+            "not match the length of the desired ordering of dimensions (" +
+            std::to_string(dimensions_indexes.size()) + ")."
```
Luthaf, line 304:
```suggestion
            "the length of `dimensions_indexes` (" + std::to_string(dimensions_indexes.size()) + ")"
            "does not match the number of dimensions in the Labels (" + std::to_string(names.size()) + ")"
```

equistore-torch/src/labels.cpp
```diff
@@ -293,6 +293,40 @@ TorchLabels LabelsHolder::insert(int64_t index, std::string name, torch::Tensor
...
+
+    for (auto index : dimensions_indexes) {
+        if (index < 0) {
+            index += names.size();
+        }
+        if (index >= names.size()) {
+            C10_THROW_ERROR(
+                IndexError,
+                "list index " + std::to_string(index) + " is out of range " +
+                std::to_string(names.size())
```
Luthaf, line 318:
```suggestion
                "out of range index "  + std::to_string(index) + " for labels dimensions (" +
                std::to_string(names.size()) + ")"
```

python/equistore-core/equistore/core/labels.py
```diff
@@ -492,6 +492,40 @@ def insert(self, index: int, name: str, values: np.ndarray) -> "Labels":
...
+        Labels(
+            baz  foo  bar
+             3   42   10
+        )
+        """
+        if len(dimensions_indexes) != len(self.names):
+            raise ValueError(
+                f"number of dimensions Labels ({len(self.names)}) does not match the "
+                f"length of the desired ordering of dimensions "
+                f"({len(dimensions_indexes)})."
```
Luthaf, line 522:
```suggestion
                f"the length of `dimensions_indexes` ({len(dimensions_indexes)})"
                f"does not match the number of dimensions in the Labels ({len(self.names)})"
```

Luthaf (approved):
The code looks good, I have some small error messages suggestions

python/equistore-operations/equistore/operations/__init__.py
```diff
@@ -46,9 +46,10 @@
 from .lstsq import lstsq
 from .manipulate_dimension import (
     append_dimension,
+    insert_dimension,
+    permute_dimension,
```
PicoCentauri, line 50: Just a minor thing. Should we name this `permute_dimensions` with an extra `s`. The function is permuting several dimensions and not only one. Against this renaming would be that it currently fits in the name space of the other functions `inser_dimension`, `remove_dimension` etc.

Luthaf: that sounds good

Luthaf (approved):

PR #317, opened 2023-08-09
Raise error when checking keys in operations

PicoCentauri:
In #280 we changed the naming of a lot of checking functions. However, we did not change these in the downstream methods where we use `_check_same_gradients` and `_check_same_keys`. Currently no errors are raised by all of the effected functions.

I hope I did not forgot some functions. I will also add some simple tests for this to prevent this in the future.

JWA edit: tagging relevant issue #318

# TODO
- [x] tests

### DIFF
--- a/python/equistore-operations/equistore/operations/_utils.py
+++ b/python/equistore-operations/equistore/operations/_utils.py
@@ -1,4 +1,4 @@
-from typing import List, Union
+from typing import List, Sequence
 
 import numpy as np
 
@@ -12,28 +12,36 @@ class NotEqualError(Exception):
 
 
 def _check_same_keys(a: TensorMap, b: TensorMap, fname: str) -> bool:
+    """
+    Returns true if the keys of 2 TensorMaps are the same, without specification of the
+    order, and false otherwise.
+    """
     return not bool(_check_same_keys_impl(a, b, fname))
 
 
 def _check_same_keys_raise(a: TensorMap, b: TensorMap, fname: str) -> None:
+    """
+    If the keys of 2 TensorMaps are not the same, raises a NotEqualError, othwerwise
+    returns None.
+    """
     message = _check_same_keys_impl(a, b, fname)
     if message != "":
         raise NotEqualError(message)
 
 
 def _check_same_keys_impl(a: TensorMap, b: TensorMap, fname: str) -> str:
-    """Check if metadata between two TensorMaps is consistent for an operation.
+    """
+    Checks if the keys of 2 TensorMaps are the same, without specification of the order.
+    Returns an empty str if they are the same, otherwise returns a str message of a
+    meaningful error.
 
-    The functions verifies that
-
-    1. The key names are the same.
-    2. The number of blocks in the same
-    3. The block key indices are the same.
+    The functions verifies that the key names and length are the same, and that the key
+    values are the same without specification of exact order.
 
     :param a: first :py:class:`TensorMap` for check
     :param b: second :py:class:`TensorMap` for check
     :param fname: name of the function where the check is performed. The input will be
-                  used to generate a meaningful error message.
+        used to generate a meaningful error message.
     """
 
     keys_a = a.keys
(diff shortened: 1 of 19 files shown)


python/equistore-operations/equistore/operations/_utils.py
```diff
@@ -198,8 +198,6 @@ def _check_same_gradients_impl(
     for parameter, grad_a in a.gradients():
         grad_b = b.gradient(parameter)
 
-        _check_same_gradients(grad_a, grad_b, props=props, fname=fname)
-
```
PicoCentauri, line 202: @agoscinski I do not see any reasons why this was here. It should have created an infinite recursion loop, since 
`_check_same_gradients` will call `_check_same_gradients_impl` which is the current function...

python/equistore-operations/equistore/operations/lstsq.py
```diff
@@ -148,7 +148,7 @@ def _lstsq_block(X: TensorBlock, Y: TensorBlock, rcond, driver) -> TensorBlock:
     Y_n_properties = Y.values.shape[-1]
     Y_values = Y.values.reshape(-1, Y_n_properties)
 
-    _check_same_gradients(X, Y, props=None, fname="lstsq")
+    _check_same_gradients_raise(X, Y, props=None, fname="lstsq")
```
jwa7, line 151: The doc tests don't like it when I set `props=["samples", "components", "properties"]` but pass with `props=None`. @PicoCentauri please may you just verify for me that this is definitely what we want in the context of least squares?

python/equistore-operations/equistore/operations/_utils.py
agoscinski: I think because it was already there before PR #280 and we just did not notice it when doing the refactor

See one commit before:
https://github.com/lab-cosmo/equistore/blob/8eb034f69e1c968fdb562cadda41301a5c7d5190/python/equistore-operations/equistore/operations/_utils.py

python/equistore-operations/equistore/operations/subtract.py
```diff
@@ -33,7 +32,6 @@ def subtract(A: TensorMap, B: Union[float, TensorMap]) -> TensorMap:
     :return: New :py:class:`TensorMap` with the same metadata as ``A``.
     """
     if isinstance(B, TensorMap):
-        _check_same_keys(A, B, "subtract")
```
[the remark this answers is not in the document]
agoscinski: Probably it is a bit confusing for the user if the error pops up at add, but I am okay.

python/equistore-operations/tests/same_raise.py
```diff
@@ -0,0 +1,87 @@
...
+
+def tensor():
+    tensor = equistore.load(
+        path.join(DATA_ROOT, "qm7-power-spectrum.npz"),
+        use_numpy=True,
+    )
+
+    # Ensure that the tensor has at least one gradient. This avoids dropping
+    # gradient tests silently by removing gradients from the reference data
+    assert "positions" in tensor.block(0).gradients_list()
```
agoscinski, line 23: I don't understand, tensor is directly loaded from the file how can it have not a gradient if it is in the file? Is some test writing to the file?

python/equistore-operations/equistore/operations/lstsq.py
PicoCentauri: The reason was that checking for `"properties"` does not make sense for `lstsq`. The values of X are a matrix while for Y it is only a vector.

Luthaf: > I also think we could set props=["samples", "components", "properties"] as the default to the function instead of None. What do you think?

if we have a default it should be a tuple, not a list (i.e. `props=("samples", "components", "properties")`), because mutable default values are problematic in Python. Otherwise, this sounds like a good idea to me!

PicoCentauri: Yes, yes indeed. I will change this.

python/equistore-operations/tests/same_raise.py
agoscinski: I feel like this assert makes in makes more more sense in the context of the test but I can also see that you don't want that this happens again in a new test. I am okay with this.

For me this is more understandable:
```suggestion
    assert len(tensor.block(0).gradients_list()) > 0, ("Tensor must have at least one gradient. When no gradients are present certain tests will pass without testing anything.")

```

python/equistore-operations/equistore/operations/add.py
```diff
@@ -33,19 +37,19 @@ def add(A: TensorMap, B: Union[float, TensorMap]) -> TensorMap:
...
-        _check_same_keys(A, B, "add")
+        _check_same_keys_raise(A, B, "add")
         for key, block_A in A.items():
             block_B = B[key]
-            _check_blocks(
+            _check_blocks_raise(
                 block_A,
                 block_B,
-                props=["samples", "components", "properties"],
+                check=["samples", "components", "properties"],
```
jwa7, line 46: Here you pass a list. Shall we change the typing on all the functions in `_utils.py` to be instead `Sequence[str]` instead of `Tuple[str]`?

python/equistore-operations/equistore/operations/_utils.py
```diff
@@ -118,74 +167,96 @@ def _check_blocks_impl(
...
         raise NotEqualError(message)
 
 
 def _check_same_gradients_impl(
-    a: TensorBlock, b: TensorBlock, props: List[str], fname: str
-) -> Union[str, None]:
+    a: TensorBlock,
+    b: TensorBlock,
+    fname: str,
+    check: Tuple[str] = ("samples", "components", "properties"),
```
jwa7, line 239:
```suggestion
    check: Sequence[str] = ("samples", "components", "properties"),
```

python/equistore-operations/equistore/operations/_utils.py
```diff
@@ -1,4 +1,4 @@
-from typing import List, Union
+from typing import List, Tuple
```
jwa7, line 1:
```suggestion
from typing import List, Sequence
```

jwa7 (commented):
Looks good thank you! Just a suggestion on typing :)

PicoCentauri:
I rebased and squashed the commits. If there are no further comments we can merge this from my side.

jwa7 (approved):

python/equistore-operations/tests/same_raise.py
PicoCentauri: Yes, this is a more helpful message. I added this to the test.

agoscinski:
looks good to me

agoscinski (approved):

PR #319, opened 2023-08-14
Add `different_keys` option to join

### LINKED ISSUE #315: Complement the .join() function with an outer_join(), inner_join(), full/symmetric_join()
Currently, joining requires tensor maps with identical keys (I hope this is still the case?).

It would be nice to have a join function that joins along and returns tensor maps only with the intersecting keys,
or adds the left and right-handed differences of the two keys and blocks to the new TensorMap.

This is relevant to writing a torch data loader for structures with different compositions.
A general collate_fn in a Dataloader can then use the join function, even if the TensorMaps from the torch Dataset have different keys due to other species pairs or compositions.

I've attached this figure of SQL Table joining operations. That highlights what I have in mind.

image

PicoCentauri:
Fixes #315 

As discussed in the issue this PR ads the `different_keys` to the join function. Also the join function should now be fully operational with torch and numpy.

I do some further comments in the code.


python/equistore-operations/equistore/operations/reduce_over_samples.py
```diff
@@ -149,7 +149,6 @@ def _reduce_over_samples_block(
 
     if reduction == "mean" or reduction == "std" or reduction == "var":
         bincount = _dispatch.bincount(index)
-        bincount = _dispatch.array_like_data(values_result, bincount)
```
PicoCentauri, line 152: These are leftovers from the old Labels class where the values have to be numpy arrays.

python/equistore-operations/equistore/operations/join.py
```diff
@@ -285,3 +301,117 @@ def join(
...
+    """
+    # Constrauct a Labels object with intersected keys
+    all_keys = tensors[0].keys
+    for tensor in tensors[1:]:
+        all_keys = all_keys.intersection(tensor.keys)
+
+    # Create new blocks and discard bocks not present in all_keys
+    new_tensors = []
+    for tensor in tensors:
+        new_blocks = [block_to(tensor.block(key)) for key in all_keys]
```
PicoCentauri, line 320: I don't know if `block_to` is doing the correct thing here. I introduce this because @bananenpampe had problems when trying to use join with a rascaline TensorMap.

python/equistore-operations/equistore/operations/_dispatch.py
```diff
@@ -90,8 +90,8 @@ def bincount(input, weights=None, minlength=0):
 
 
 def array_like_data(array, data):
-    """Return `data` with the same data-type and array-type of array"""
-    if _check_all_same_type([array], type(data)):
+    """Create an object with the same type as `array` containing the given `data`."""
+    if type(data) in (np.ndarray, TorchTensor):
```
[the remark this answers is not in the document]
DavideTisi: I do not know I would keep `_check_all_same_type` at least for now for the other functions.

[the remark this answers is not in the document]
PicoCentauri: Yes for the other functions of course. Only in this one.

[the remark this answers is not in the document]
Luthaf: This code looks wrong to me: if array is torch.Tensor and data is numpy, we want to convert to torch,not return data directly

python/equistore-operations/equistore/operations/join.py
```diff
@@ -48,10 +51,13 @@ def join(
     :param axis:
         A string indicating how the tensormaps are stacked. Allowed
         values are ``'properties'`` or ``'samples'``.
+    :param different_keys: Method to handle different keys between the tensors. For
+        ``"error"`` keys in all tensors have to be the same. For ``"intersection"``
+        only blocks present in all tensors will be taken be present in the joined
+        tensor. For ``"union"`` new empty blocks will be created if a key is missing.
```
Luthaf, line 57:
```suggestion
        tensor. For ``"union"`` missing keys will be treated like if they where associated with an empty block.
```

python/equistore-operations/equistore/operations/join.py
```diff
@@ -285,3 +301,117 @@ def join(
...
         return tensor_joined
+
+
+def _intersection(tensors: List[TensorMap]):
+    """Create a new tensors list where keys are based on the intersection from all
+    tensors.
+
+    Blocks corresponding to keys that are not present in all tensor will be discarded.
+    """
+    # Constrauct a Labels object with intersected keys
```
Luthaf, line 312:
```suggestion
    # Construct a Labels object with intersected keys
```

python/equistore-operations/equistore/operations/join.py
Luthaf: I don't think we want to do conversions here. If the user wants to join torch & numpy data, they need to call `equistore.to` themself.

python/equistore-operations/equistore/operations/join.py
```diff
@@ -285,3 +301,117 @@ def join(
         return remove_dimension(tensor_joined, name="tensor", axis=axis)
     else:
         return tensor_joined
+
+
+def _intersection(tensors: List[TensorMap]):
```
Luthaf, line 306:
```suggestion
def _tensors_intersection(tensors: List[TensorMap]) -> List[TensorMap]:
```

python/equistore-operations/equistore/operations/join.py
```diff
@@ -218,13 +233,9 @@ def join(
...
     #
-    # A nested list with sublist of different shapes can not be handled by np.unique.
-    unique_names = np.unique(functools.reduce(operator.concat, names_list))
-
-    # Label names are unique: We can do an equal check only checking the lengths.
-    names_are_same = np.all(
-        len(unique_names) == np.array([len(names) for names in names_list])
-    )
+    # A nested list with sublist of different shapes can not be handled by `set`.
+    unique_names = set(functools.reduce(operator.concat, names_list))
```
Luthaf, line 237: I feel like this will not work with TorchScript. Could we rewrite this without using functools?

python/equistore-operations/equistore/operations/join.py
```diff
@@ -285,3 +301,117 @@ def join(
...
+    new_tensors = []
+    for tensor in tensors:
+        new_blocks = [block_to(tensor.block(key)) for key in all_keys]
+
+        new_tensors.append(TensorMap(keys=all_keys, blocks=new_blocks))
+
+    return new_tensors
+
+
+def _union(tensors: List[TensorMap], axis: str):
```
Luthaf, line 327:
```suggestion
def _tensor_union(tensors: List[TensorMap], axis: str) -> List[TensorMap]:
```

python/equistore-operations/equistore/operations/join.py
```diff
@@ -285,3 +301,117 @@ def join(
...
+    """Create a new tensors list where keys are based on the union from all tensors.
+
+    Missing keys will be filled by empty blocks having containing zero samples.
```
Luthaf, line 330:
```suggestion
    Missing keys will be filled by empty blocks having containing no labels in the ``axis`` dimension.
```

python/equistore-operations/equistore/operations/join.py
```diff
@@ -285,3 +301,117 @@ def join(
...
+            # reference_block
+            if axis == "samples":
+                values = _dispatch.empty_like(
+                    array=reference_block.values,
+                    shape=(0,) + reference_block.values.shape[1:],
+                )
+                samples = Labels.empty(reference_block.samples.names)
+                properties = reference_block.properties
+            else:
+                values = _dispatch.empty_like(
```
Luthaf, line 370:
```suggestion
                assert axis == "properties"
                values = _dispatch.empty_like(
```

python/equistore-operations/equistore/operations/join.py
PicoCentauri: Of course it doesn't work

```python
import functools
import operator

import torch

@torch.jit.script
def func() -> str:
    names_list = [["1", "2"], ["1", "5"]]
    return set(functools.reduce(operator.concat, names_list))

func()
```

throws an error. But, this is just a list of lists. We most likely can do it with a comprehension.

python/equistore-operations/equistore/operations/_dispatch.py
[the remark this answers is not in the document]
PicoCentauri: Makes sense then we just do a simple type check between `data` and `array`.

PicoCentauri:
Thanks for the helpful review @Luthaf!

python/equistore-operations/equistore/operations/_dispatch.py
```diff
@@ -98,9 +98,12 @@ def bincount(input, weights=None, minlength=0):
 
 
 def array_like_data(array, data):
-    """Return `data` with the same data-type and array-type of array"""
-    if isinstance(array, TorchTensor):
-        _check_all_torch_tensor([data])
+    """Create an object with the same type as `array` containing the given `data`."""
+    if type(array) is type(data):
+        return data
```
Luthaf, line 103: This is still missing `dtype` conversion. I'm not sure why this change is needed, could you explain?

python/equistore-operations/equistore/operations/join.py
```diff
@@ -285,3 +303,118 @@ def join(
...
+    new_tensors = []
+    for tensor in tensors:
+        _, map, _ = all_keys.intersection_and_mapping(tensor.keys)
+
+        missing_keys = Labels(
+            names=tensor.keys.names, values=all_keys.values[map == -1]
+        )
+
+        new_keys = tensor.keys.union(missing_keys)
+        new_blocks = [block_to(block) for block in tensor]
```
Luthaf, line 349: why are we doing `block_to` here? Especially without any argument, that will not do much? If we need a copy of the block, we can call `block.clone()`

python/equistore-operations/equistore/operations/join.py
```diff
@@ -285,3 +303,118 @@ def join(
...
+    return new_tensors
+
+
+def _tensors_union(tensors: List[TensorMap], axis: str) -> List[TensorMap]:
+    """Create a new tensors list where keys are based on the union from all tensors.
+
+    Missing keys will be filled by empty blocks having containing no labels in the
+    ``axis`` dimension.
+    """
+    # Constrauct a Labels object with all keys
```
Luthaf, line 334:
```suggestion
    # Construct a Labels object with all keys
```

python/equistore-operations/equistore/operations/_dispatch.py
PicoCentauri: This is weird code indeed. Let me change this back to something useful. The problem is that it does not work if data is a `list`.

python/equistore-operations/equistore/operations/_dispatch.py
```diff
@@ -97,13 +99,12 @@ def bincount(input, weights=None, minlength=0):
...
 
 
-def array_like_data(array, data):
-    """Return `data` with the same data-type and array-type of array"""
+def list_to_array(array, data: List):
+    """Create an object from data wit the same type as `array`."""
     if isinstance(array, TorchTensor):
-        _check_all_torch_tensor([data])
-        return torch.Tensor(data, device=array.device)
+        tensor = torch.Tensor(data, device=array.device)
```
Luthaf, line 105:
```suggestion
        tensor = torch.tensor(data, device=array.device, dtype=array.dtype)
```

this does not work?

PicoCentauri: I tried this at some point and it failed with a weird error message. I tried it again 5 minutes ago and it worked... I change it. Thanks for the suggestion!

Luthaf:
https://github.com/lab-cosmo/equistore/pull/325 should fix the CI.

python/equistore-operations/equistore/operations/_dispatch.py
Luthaf: `torch.Tensor` and `torch.tensor` are different things, which might explain why one works and not the other. The documentation seems to point people toward the function `torch.tensor`, but if the code works with the class constructor here, that's fine for me.

Luthaf (approved):
Looks good! Let's merge #325 and then rebase this one to check the CI fix.

PR #320, opened 2023-08-15
Make operations availble from equistore-torch

Luthaf:
Part of #133, this only introduces the infrastructure to import the code twice within two different contexts, and porting a single operation to be fully compatible with TorchScript. (the corresponding commit can be used as inspiration for the other operations, which I'll leave to future PR)

----
:books: Documentation preview :books:: https://equistore--320.org.readthedocs.build/en/320/

### DIFF
--- a/python/equistore-operations/equistore/operations/remove_gradients.py
+++ b/python/equistore-operations/equistore/operations/remove_gradients.py
@@ -24,8 +24,8 @@ def remove_gradients(
     if remove is None:
         remove = tensor.block(0).gradients_list()
 
-    blocks = []
-    for block in tensor:
+    blocks: List[TensorBlock] = []
+    for block in tensor.blocks():
         new_block = TensorBlock(
             values=block.values,
             samples=block.samples,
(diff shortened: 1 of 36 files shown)


python/equistore-operations/equistore/operations/remove_gradients.py
```diff
@@ -24,8 +24,8 @@ def remove_gradients(
     if remove is None:
         remove = tensor.block(0).gradients_list()
 
-    blocks = []
-    for block in tensor:
+    blocks: List[TensorBlock] = []
```
jwa7, line 27: What is this magic?? :D

jwa7 (approved):
Looks fine to me :)

python/equistore-operations/equistore/operations/remove_gradients.py
PicoCentauri: typehinting :-)

PicoCentauri (approved):
Looks good. Next step would be continuing removing the explicit numpy dependencies from the operations.

PR #322, opened 2023-08-16
Make equistore.operations._utils TorchScript compatible

agoscinski:
- typing.Sequence is not TorchScript compatible
- change iterator to `range(len(keys)))`
- removed np.all operations for comparison (I don't know if this has been changed in the API, but a comparison between Labels returns a boolean in the current version)

~TODO torch script tests~

----
:books: Documentation preview :books:: https://equistore--322.org.readthedocs.build/en/322/

### DIFF
--- a/python/equistore-operations/equistore/operations/_utils.py
+++ b/python/equistore-operations/equistore/operations/_utils.py
@@ -19,7 +19,7 @@ def _check_same_keys(a: TensorMap, b: TensorMap, fname: str) -> bool:
 
 def _check_same_keys_raise(a: TensorMap, b: TensorMap, fname: str) -> None:
     """
-    If the keys of 2 TensorMaps are not the same, raises a NotEqualError, othwerwise
+    If the keys of 2 TensorMaps are not the same, raises a NotEqualError, otherwise
     returns None.
     """
     message = _check_same_keys_impl(a, b, fname)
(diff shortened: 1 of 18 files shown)


python/equistore-operations/equistore/operations/_utils.py
```diff
@@ -179,7 +183,7 @@ def _check_same_gradients(
     a: TensorBlock,
     b: TensorBlock,
     fname: str,
-    check: Sequence[str] = ("samples", "components", "properties"),
+    check: List[str] = ("samples", "components", "properties"),
```
PicoCentauri, line 186: I would change this back to a `Tuple`. Also we also have to change this in `equal_metadata.py`

Luthaf: I think TorchScript checked the number of tuple elements in the type; i.e. `Tuple[str, str, str]` does not match `Tuple[str, str]`. And it does not supports the syntax for arbitrary sized tuple `Tuple[str, ...]`

Luthaf: So we might actually have to change this back to using a special value `None`/`"all"`/... to check everything. Maybe `check: Union[List[str], str] = "all",`, which can be set to `"all"`, `"samples"`, `"components"`, `"properties"` or a list containing any of the last three.

agoscinski: You are right, I did not test the most recent change. Agree with `None/"all"` solution. No strong opinions if we do "all" or None.

frostedoyster:
Currently getting an absurd error from `_utils.py` in #324. Essentially, the compiler says that it wants the default value for the `check` parameter as a tuple of strings (which it is, but it shows an error...)

frostedoyster:
Ok, yes the issue is that, as @Luthaf says, the size of the tuple needs to be specified

PicoCentauri (approved):
Thanks @agoscinski  and @Luthaf  for the work LGTM!

PR #325, opened 2023-08-21
Fix documentation build by pinning sphinx

Luthaf:
Sphinx recently updated all path from string to pathlib, but some plugins we use are not yet updated.

We are hitting https://github.com/breathe-doc/breathe/issues/943 and https://github.com/pradyunsg/furo/discussions/693 for other versions of sphinx.

----
:books: Documentation preview :books:: https://equistore--325.org.readthedocs.build/en/325/

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
-sphinx >=4.4
+sphinx ==7.1.2
 sphinx-toggleprompt # hide the prompt (>>>) in python doctests
 pygments >=2.11     # syntax highligthing
 breathe >=4.33      # C and C++ => sphinx through doxygen


docs/requirements.txt
```diff
@@ -1,4 +1,4 @@
-sphinx >=4.4
+sphinx ==7.1.2
```
PicoCentauri, line 1: to be consistent with rascaline

```suggestion
sphinx < 7.2    # breathe is not compatible with sphinx 7.2 yet
```

PicoCentauri (changes requested):

docs/requirements.txt
Luthaf: I tried this first, but it gives me error with furo (it seems to resolve to sphinx 6.2 which also fails)

Luthaf: not sure why rascaline is not failing ...

PicoCentauri: okay very nice! lets go for it.

PicoCentauri (approved):

PR #328, opened 2023-08-21
Make math functions torch-script-able

frostedoyster:
These are `add()`, `subtract()`, `pow()`, `multiply()`, `divide()`.

----
:books: Documentation preview :books:: https://equistore--328.org.readthedocs.build/en/328/


frostedoyster:
At the moment, the scripted version of `TensorMap` doesn't seem to be recognizable as a "type".

agoscinski (commented):
> At the moment, the scripted version of TensorMap doesn't seem to be recognizable as a "type".

Yes torch.TensorMap is not a type but a ScriptClass object during runtime. One has to do this dispatching which can be done with `check_isinstance`. During scripting the python inbuilt `isinstance` function works. However we  cannot use `check_isinstance` in this case because we cannot pass a type with TorchScript. So I added a dispatch for scripting into add.

Also I did need to change the order of the if-then-else, because TorchScript could not infer from the instance check that the type must be `TensorMap`, it still assumed the Union type and then I returned an error in the `_check_same_keys_raise` that only accepts a `TensorMap` as second argument.

python/equistore-operations/equistore/operations/pow.py
```diff
@@ -1,42 +1,74 @@
...
+    _check_blocks_raise,
+    _check_same_gradients_raise,
+    _check_same_keys_raise,
+)
 
 
-def pow(A: TensorMap, B: float) -> TensorMap:
-    r"""Return a new :class:`TensorMap` with the same metadata of ``A``
-    and the values being the element-wise ``B``-power of ``A.values``.
+def divide(A: TensorMap, B: Union[float, int, TensorMap]) -> TensorMap:
```
DavideTisi, line 12: why ``divide`` is in ``pow.py``? maybe copy and paste?

python/equistore-torch/tests/operations/pow.py
```diff
@@ -0,0 +1,17 @@
...
+def test_operation_as_python():
+    tensor = load_data("qm7-power-spectrum.npz")
+    power_tensor = equistore.torch.pow(tensor, 2.0)
+    assert equistore.torch.equal_metadata(power_tensor, tensor)
+
+
+def test_operation_as_torch_script():
+    tensor = load_data("qm7-power-spectrum.npz")
+    power_tensor = torch.jit.script(equistore.torch.pow)(tensor, 2.0)
+    assert equistore.torch.equal_metadata(power_tensor, tensor)
```
DavideTisi, line 17: should also check the values

DavideTisi (changes requested):

python/equistore-operations/equistore/operations/pow.py
frostedoyster: Yes 😄

python/equistore-torch/tests/operations/pow.py
frostedoyster: That's a fair point. We should probably decide a policy for these tests. Are they only supposed to check some very basic properties to ensure that the ouput isn't completely wrong or should they be proper tests? @Luthaf 
Worth deciding before writing more of these

Luthaf: I think checking the output in one simple case would be nice and allow to catch some errors. We don't have to be as thorough as the main tests for corner cases and special values.

One extra thing that could be helpful in these tests would be to check the type of the output (see `remove_gradients`).

python/equistore-torch/tests/operations/add.py
```diff
@@ -0,0 +1,19 @@
+import torch
+
+import equistore.torch
+
+from .data import load_data
+
+
+def test_operation_as_python():
```
Luthaf, line 8: instead of duplicating all the tests, could you use the same pattern as https://github.com/lab-cosmo/equistore/pull/327/commits/dd87c5bde341314e178f314a88bd3e00673559f9 or https://github.com/lab-cosmo/equistore/commit/b4969ed51d55a8a845acb4bad22ed1c86734ddc9#diff-2c35cd5d88b7f753ab0edbf63d88798ec6bab12cb0d6da020bbe19d427df434f?

Luthaf: This way it is also easier to add more tests for the actual behavior of the code

Luthaf (approved):

PR #330, opened 2023-08-22
Make sure the Rust binding use the shared library on windows

Luthaf:
The code was failing to find the shared library, and then defaulting to the static version. This caused issues when using the code from rascaline, since the corresponding Python module where using different versions of the code.

----
:books: Documentation preview :books:: https://equistore--330.org.readthedocs.build/en/330/

### DIFF
--- a/equistore/build.rs
+++ b/equistore/build.rs
@@ -46,15 +46,28 @@ fn main() {
         equistore_core.push(splitted[..splitted.len() - 1].join("."));
     }
 
-    let mut install_dir = cmake::Config::new(&equistore_core)
+    let install_dir = cmake::Config::new(&equistore_core)
         .define("CARGO_EXE", env!("CARGO"))
         .define("RUST_BUILD_TARGET", std::env::var("TARGET").unwrap())
+        .define("BUILD_SHARED_LIBS", if cfg!(feature="static") { "OFF" } else { "ON" })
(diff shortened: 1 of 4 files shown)


PicoCentauri:
Is there a way to test this even further. To test were also passing before. If we destroy this again we will not notice...

equistore/build.rs
```diff
@@ -46,15 +46,28 @@ fn main() {
         equistore_core.push(splitted[..splitted.len() - 1].join("."));
     }
 
-    let mut install_dir = cmake::Config::new(&equistore_core)
+    let install_dir = cmake::Config::new(&equistore_core)
         .define("CARGO_EXE", env!("CARGO"))
         .define("RUST_BUILD_TARGET", std::env::var("TARGET").unwrap())
+        .define("BUILD_SHARED_LIBS", if cfg!(feature="static") { "OFF" } else { "ON" })
+        .define("EQUISTORE_INSTALL_BOTH_STATIC_SHARED", "OFF")
```
Luthaf, line 53: @PicoCentauri this should be our guard against this failure: we don't install the "other" library anymore, so if we break this again it will fail to link.

PicoCentauri: Ahh yes I see. Okay lets go for it then.

PicoCentauri (approved):

PR #331, opened 2023-08-22
Make join() torch-script-able

PicoCentauri:
Still work in progress due to the wonderful compiler.

----
:books: Documentation preview :books:: https://equistore--331.org.readthedocs.build/en/331/

### DIFF
--- a/python/equistore-operations/equistore/operations/_dispatch.py
+++ b/python/equistore-operations/equistore/operations/_dispatch.py
@@ -1,4 +1,4 @@
-from typing import List, Optional
+from typing import List, Optional, Union
 
 import numpy as np
 
@@ -112,10 +112,10 @@ def copy(array):
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
-def list_to_array(array, data: List):
+def list_to_array(array, data: Union[List[int], List[List[int]]]):
     """Create an object from data with the same type as ``array``."""
     if isinstance(array, TorchTensor):
-        return torch.Tensor(data, device=array.device, dtype=array.dtype)
+        return torch.tensor(data).to(array.dtype).to(array.device)
     elif isinstance(array, np.ndarray):
         return np.array(data, dtype=array.dtype)
     else:
@@ -486,6 +486,14 @@ def to_index_array(array):
         return array
     else:
         raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def unique(array, axis: Optional[int] = None):
+    """Find the unique elements of an array."""
+    if isinstance(array, TorchTensor):
+        return torch.unique(array, dim=axis)
+    elif isinstance(array, np.ndarray):
+        return np.unique(array, axis=axis)
 
 
 def unique_with_inverse(array, axis: Optional[int] = None):
--- a/python/equistore-operations/equistore/operations/join.py
+++ b/python/equistore-operations/equistore/operations/join.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Union
 
 from . import _dispatch
 from ._classes import Labels, TensorBlock, TensorMap
@@ -22,6 +22,11 @@ def _disjoint_tensor_labels(tensors: List[TensorMap], axis: str) -> bool:
                 elif axis == "properties":
                     first_labels = first_block.properties
                     second_labels = second_block.properties
+                else:
+                    raise ValueError(
+                        "Only `'properties'` or `'samples'` are "
+                        "valid values for the `axis` parameter."
+                    )
 
                 if len(first_labels.intersection(second_labels)):
                     return False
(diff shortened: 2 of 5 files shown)


python/equistore-torch/equistore/torch/operations.py
```diff
@@ -89,7 +89,3 @@ def check_isinstance(obj, ty):
 # with the newly created module
 sys.modules[spec.name] = module
 spec.loader.exec_module(module)
-
-# Step 3: override the objects/functions/classes exposed by this module with the data
-# from the module we just created.
-globals().update(module.__dict__)
```
frostedoyster, line 95: Why is this not necessary anymore?

python/equistore-operations/equistore/operations/manipulate_dimension.py
```diff
@@ -312,8 +312,8 @@ def remove_dimension(tensor: TensorMap, axis: str, name: str) -> TensorMap:
     if axis == "keys":
         keys = keys.remove(name=name)
 
-    blocks = []
-    for block in tensor:
+    blocks: List[TensorBlock] = []
+    for _, block in tensor.items():
```
frostedoyster, line 316:
```suggestion
    for block in tensor.blocks():
```

(just for style and consistency with other scriptable functions)

python/equistore-operations/equistore/operations/join.py
```diff
@@ -244,14 +258,27 @@ def join(
...
 
-        for block in tensor:
+    keys = Labels(
+        names=["tensor"] + keys.names,
+        values=new_keys_values.reshape(-1, n_keys_dimensions),
+    )
+
+    blocks: List[TensorBlock] = []
+    for tensor in tensors:
+        for _, block in tensor.items():
```
frostedoyster, line 281:
```suggestion
        for block in tensor.blocks():
```

python/equistore-operations/equistore/operations/_dispatch.py
```diff
@@ -101,12 +101,12 @@ def bincount(input, weights=None, minlength=0):
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
-def list_to_array(array, data: List):
+def list_to_array(array, data: List[List[int]]):
     """Create an object from data with the same type as ``array``."""
     if isinstance(array, TorchTensor):
-        return torch.Tensor(data, device=array.device, dtype=array.dtype)
+        return torch.tensor(data)
```
frostedoyster, line 107:
```suggestion
        return torch.tensor(data).to(array.dtype).to(array.device)
```

If we want to stay true to the docstring and the previous implementation

python/equistore-operations/equistore/operations/_dispatch.py
```diff
@@ -101,12 +101,12 @@ def bincount(input, weights=None, minlength=0):
...
     elif isinstance(array, np.ndarray):
-        return np.array(data, dtype=array.dtype)
+        return np.array(data)
```
frostedoyster, line 109:
```suggestion
        return np.array(data, dtype=array.dtype)
```

frostedoyster:
I've tried to resolve the conflicts between this and `master`.
Two potential issues:
- linter won't be happy
- I've tried `def list_to_array(array, data: Union[List[int], List[List[int]]]):`, because `master` had a `list_to_array` function that takes `List[int]`, while the homonymous function in this branch takes `List[List[int]]]`. Odds are that TorchScript won't like this so we'll have to split that into two functions

python/equistore-torch/equistore/torch/operations.py
PicoCentauri: Because we expose all functions to the current namespace with `pec.loader.exec_module(module)`. I think this is leftover when @Luthaf started the work on this.

frostedoyster: Perfect, thanks for the explanation. I see that it doesn't break anything so all good

python/equistore-operations/equistore/operations/manipulate_dimension.py
PicoCentauri: done

python/equistore-operations/equistore/operations/_dispatch.py
```diff
@@ -112,12 +112,12 @@ def copy(array):
...
 
-def list_to_array(array, data: List):
+def list_to_array(array, data: Union[List[int], List[List[int]]]):
     """Create an object from data with the same type as ``array``."""
     if isinstance(array, TorchTensor):
-        return torch.Tensor(data, device=array.device, dtype=array.dtype)
+        return torch.tensor(data).to(array.dtype).to(array.device)
     elif isinstance(array, np.ndarray):
-        return np.array(data, dtype=array.dtype)
+        return np.array(data, array.dtype)
```
frostedoyster, line 120:
```suggestion
        return np.array(data, dtype=array.dtype)
```

python/equistore-operations/equistore/operations/_utils.py
```diff
@@ -42,7 +42,7 @@ def _check_same_keys_impl(a: TensorMap, b: TensorMap, fname: str) -> str:
         used to generate a meaningful error message.
     """
 
-    keys_a = a.keys
+    keys_a: Labels = a.keys
```
frostedoyster, line 45:
```suggestion
    keys_a = a.keys
```

frostedoyster (approved):

PR #332, opened 2023-08-23
Make `reduce_over_samples` torch-script-able

frostedoyster:
I had to eliminate the `with np.errstate()` block because it was incompatible with TorchScript.
Perhaps we can document instead in what cases the numpy warning will be given

----
:books: Documentation preview :books:: https://equistore--332.org.readthedocs.build/en/332/

### DIFF
--- a/python/equistore-operations/equistore/operations/_dispatch.py
+++ b/python/equistore-operations/equistore/operations/_dispatch.py
@@ -72,7 +72,7 @@ def allclose(
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
-def bincount(input, weights=None, minlength=0):
+def bincount(input, weights: Optional[TorchTensor] = None, minlength : int = 0):
     """Count number of occurrences of each value in array of non-negative ints.
     Equivalent of ``numpy.bitcount(input, weights, minlength)``
 
@@ -97,6 +97,17 @@ def bincount(input, weights=None, minlength=0):
         if weights is not None:
             _check_all_np_ndarray([weights])
         return np.bincount(input, weights=weights, minlength=minlength)
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
+    
+
+def copy(array):
+    """Returns a copy of ``array``. 
+    The new data is not shared with the original array"""
+    if isinstance(array, TorchTensor):
+        return array.clone()
+    elif isinstance(array, np.ndarray):
+        return array.copy()
     else:
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
@@ -219,7 +230,7 @@ def lstsq(X, Y, rcond, driver=None):
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
-def nan_to_num(X, nan=0.0, posinf=None, neginf=None):
+def nan_to_num(X, nan: float = 0.0, posinf: Optional[float] = None, neginf: Optional[float] = None):
     """Equivalent to np.nan_to_num(X, nan, posinf, neginf)"""
     if isinstance(X, TorchTensor):
         return torch.nan_to_num(X, nan=nan, posinf=posinf, neginf=neginf)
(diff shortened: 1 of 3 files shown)


python/equistore-operations/equistore/operations/_dispatch.py
```diff
@@ -330,7 +339,7 @@ def ones_like(array, shape=None, requires_grad=False):
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
-def empty_like(array, shape=None, requires_grad=False):
+def empty_like(array, shape: Optional[List[int]] = None, requires_grad: bool = False):
```
Luthaf, line 342: I don't think this should be part of this PR

python/equistore-operations/equistore/operations/reduce_over_samples.py
```diff
@@ -258,22 +265,17 @@ def _reduce_over_samples_block(
...
                     gradient_values_result = 2 * (
                         values_grad_result - gradient_values_result
                     )
                 else:  # std
                     for i, s in enumerate(new_gradient_samples):
-                        # only numpy raise a warning for division by zero
-                        # so the statement catch that
-                        # for torch there is nothing to catch
-                        # both numpy and torch give inf for the division by zero
-                        with np.errstate(divide="ignore", invalid="ignore"):
```
Luthaf, line 272: could we keep it by defining a function that does nothing for torch? Something like

```py
if torch.jit.is_scripting():
    def np_errstate(divide: str, invalid: str) -> None:
        return None
else:
    np_errstate = np.errstate
```

python/equistore-operations/equistore/operations/_dispatch.py
frostedoyster: `zeros_like` and `empty_like` of `_dispatch` are used inside the reduce operations. I'll make sure to rebase

python/equistore-operations/equistore/operations/reduce_over_samples.py
frostedoyster: The issue is that we don't want to import torch in that file, right?

frostedoyster: Plus, honestly, if the std is undefined, it's probably right that the user receives a warning instead of just a std of zero

Luthaf: it is already exposed I think: `from ._classes import torch_jit_is_scripting`

Luthaf: yes: https://github.com/lab-cosmo/equistore/blob/f09ec47369f71e4d71d02e372bd809da9ec5293f/python/equistore-operations/equistore/operations/equal_metadata.py#L6

frostedoyster: True. I've made a first attempt but the compiler is saying that, within a `with x():` expression, `x()` must return an object (see CI output)

frostedoyster: Found a solution

python/equistore-operations/equistore/operations/_dispatch.py
frostedoyster: Rebased

python/equistore-operations/equistore/operations/reduce_over_samples.py
```diff
@@ -76,31 +92,36 @@ def _reduce_over_samples_block(
...
-        remaining_samples = [
-            s_name for s_name in block_samples.names if s_name not in sample_names
-        ]
+        remaining_samples_final: List[str] = []
+        for s_name in block_samples.names:
+            if s_name in sample_names:
+                continue
+            remaining_samples_final.append(s_name)
+    else:
+        remaining_samples_final = remaining_samples
```
Luthaf, line 101:
```suggestion
        remaining_samples_names = remaining_samples
```

if we need a separate variable here, this might make the content a bit clearer

Luthaf (approved):
Looks good overall, I have one small suggestion

python/equistore-operations/equistore/operations/reduce_over_samples.py
frostedoyster: Done

PR #337, opened 2023-08-24
Make `to` torch-script-able

frostedoyster:
----
:books: Documentation preview :books:: https://equistore--337.org.readthedocs.build/en/337/


python/metatensor-operations/metatensor/operations/to.py
```diff
@@ -116,19 +120,81 @@ def block_to(
...
-    return _block_to(block, backend, dtype, device, requires_grad)
+    # Walk the tree of gradients without recursion
+    # (recursion is not supported by torchscript)
+    current_location: List[str] = []
+    current_block = block
+    transformed_blocks: List[TensorBlock] = []
+    last_visited = ""
+
+    while True:
+        print(len(transformed_blocks))
```
PicoCentauri, line 131:
```suggestion
```

leftover debug statement?

python/metatensor-operations/metatensor/operations/to.py
```diff
@@ -116,19 +120,81 @@ def block_to(
...
+        gradient_names = current_block.gradients_list()
+        n_gradients = len(gradient_names)
+        if last_visited == "":  # first time we see this block
```
PicoCentauri, line 134: You should write a test for this code. It is quite complex and seems it will break easily.

python/metatensor-operations/metatensor/operations/to.py
```diff
@@ -116,19 +120,81 @@ def block_to(
...
+                )  # reach current location
+            else:  # more gradients to convert
+                # walk forward:
+                current_block = current_block.gradient(
+                    gradient_names[index_last_visited + 1]
+                )
+                current_location.append(gradient_names[index_last_visited + 1])
+                last_visited = ""
+
+    # at this point, transformed_blocks should only contain the final transformed block:
```
PicoCentauri, line 184:
```suggestion
    # at this point, transformed_blocks contain only the final transformed block:
```

python/metatensor-operations/metatensor/operations/to.py
```diff
@@ -116,19 +120,81 @@ def block_to(
...
+    return transformed_blocks[0]
```
PicoCentauri, line 185: what about the other elements of `transformed_blocks`?

python/metatensor-operations/metatensor/operations/to.py
```diff
@@ -116,19 +120,81 @@ def block_to(
...
 
-    return _block_to(block, backend, dtype, device, requires_grad)
+    # Walk the tree of gradients without recursion
+    # (recursion is not supported by torchscript)
+    current_location: List[str] = []
+    current_block = block
+    transformed_blocks: List[TensorBlock] = []
+    last_visited = ""
+
+    while True:
```
PicoCentauri, line 130: What we do here is a depth-first search. Maybe it is worth if there a pseudocode which is a bit simpler. There is for example one here:

https://en.wikipedia.org/wiki/Depth-first_search

Also it maybe make sense to extract the code in the end to made this available for other classes.

python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -434,43 +444,58 @@ def rand_like(array, shape: Optional[List[int]] = None, requires_grad: bool = Fa
...
+        if backend == "torch":
+            # We need this to keep gradients of the tensor
+            new_array = array.to(dtype=dtype).to(device=device)
             if requires_grad is not None:
-                new_array.requires_grad = requires_grad
+                new_array.requires_grad_(requires_grad)
             return new_array
 
+        elif backend == "numpy":
+            if torch.jit.is_scripting():
```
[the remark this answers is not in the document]
frostedoyster: Completely true. We'll have to remove the check and live with whatever the error is when calling `.numpy()` in a torchscript model.

python/metatensor-operations/metatensor/operations/to.py
frostedoyster: I've read and understood the algorithm. My first impression is that it could be used to improve the code significantly, but adapting it to our use-case and the functions we have available in equistore would take me one or two days and unfortunately I don't have that much time. I will add more comments to explain my current algorithm, which should probably just be taken as a temporary fix to make `to()` scriptable. If, in the future, we want to support gradients of gradients for more operations, we should definitely revisit this topic. Of course, if anybody else in the group (perhaps with a CS background) wants to work on it, that would be an ideal solution

python/metatensor-operations/metatensor/operations/to.py
frostedoyster: They will all have been popped from the list and added to the first one as gradients

PicoCentauri (approved):
Very good tests. This will help us in the future!

PR #340, opened 2023-08-24
Make `one_hot()` torch-script-able

frostedoyster:
Since torchscript doesn't seem to support `torch.get_default_dtype()`, I decided to output the encoding as integers

----
:books: Documentation preview :books:: https://equistore--340.org.readthedocs.build/en/340/

### DIFF
--- a/python/equistore-operations/equistore/operations/_dispatch.py
+++ b/python/equistore-operations/equistore/operations/_dispatch.py
@@ -110,6 +110,20 @@ def copy(array):
         return array.copy()
     else:
         raise TypeError(UNKNOWN_ARRAY_TYPE)
+    
+
+def eye_like(array, size: int):
+    """
+    Create an identity matrix with the given ``size``, and the same
+    dtype and device as ``array``.
+    """
+
+    if isinstance(array, TorchTensor):
+        return torch.eye(size).to(array.dtype).to(array.device)
+    elif isinstance(array, np.ndarray):
+        return np.eye(size, dtype=array.dtype)
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
 def list_to_array(array, data: List):
--- a/python/equistore-operations/equistore/operations/_dispatch.py
+++ b/python/equistore-operations/equistore/operations/_dispatch.py
@@ -110,7 +110,7 @@ def copy(array):
         return array.copy()
     else:
         raise TypeError(UNKNOWN_ARRAY_TYPE)
-    
+
 
 def eye_like(array, size: int):
     """
--- a/python/equistore-operations/equistore/operations/one_hot.py
+++ b/python/equistore-operations/equistore/operations/one_hot.py
@@ -71,5 +71,7 @@ def one_hot(labels: Labels, dimension: Labels):
             )
         indices[i] = position
 
-    one_hot_array = _dispatch.eye_like(dimension.values, len(dimension))[indices]
+    one_hot_array = _dispatch.eye_like(dimension.values, len(dimension))[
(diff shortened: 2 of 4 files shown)


python/equistore-operations/equistore/operations/_dispatch.py
```diff
@@ -112,6 +112,20 @@ def copy(array):
...
 
 
+def eye_like(array, size: int):
+    """
+    Create an identity matrix with the given ``size``, and the same
+    dtype and device as ``array``.
+    """
+
+    if isinstance(array, TorchTensor):
+        return torch.eye(size).to(array.dtype).to(array.device)
```
Luthaf, line 122:
```suggestion
        return torch.eye(size).to(dtype=array.dtype, device=array.device)
```

python/equistore-operations/equistore/operations/one_hot.py
```diff
@@ -60,15 +59,19 @@ def one_hot(labels: Labels, dimension: Labels) -> np.ndarray:
...
     name = dimension.names[0]
 
-    indices = np.zeros(len(labels), dtype=np.int64)
-    for i, entry in enumerate(labels[name]):
-        position = dimension.position([entry])
+    indices = _dispatch.zeros_like(dimension.values, [len(labels)])
+    labels_name = labels.column(name)
+    for i in range(labels_name.shape[0]):
+        entry = labels_name[None, i]
+        position = dimension.position(entry)
```
Luthaf, line 66: instead of looping through everything in Python, we should also be able to rewrite this in term of `Labels.union`/`Labels.intersection`

python/equistore-torch/tests/operations/one_hot.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+    one_hot_encoding = one_hot(original_labels, possible_labels)
+
+    # check output type
+    assert isinstance(one_hot_encoding, torch.Tensor)
+
+    # check values
+    print(one_hot_encoding)
+    print(correct_encoding)
+    assert torch.allclose(one_hot_encoding, correct_encoding)
```
Luthaf, line 33:
```suggestion
    assert torch.all(one_hot_encoding == correct_encoding)
```

Luthaf (approved):

python/equistore-operations/equistore/operations/_dispatch.py
frostedoyster: Unfortunately this single call doesn't map to the `aten` functions and it's necessary to split it into a call that changes the dtype and one that changes the device

python/equistore-operations/equistore/operations/one_hot.py
frostedoyster: I tried, but the issue is that we would have to create a `Labels` by extracting a single dimension of `labels`, and that leads to repeated entries.

frostedoyster: I'll now try with array/tensor operations

frostedoyster: Update: I can't see an obvious way that doesn't require a lot of dispatch code and makes the error messages a lot more cryptic. I would stick with the current implementation until the operation becomes a bottleneck in our models

python/equistore-operations/equistore/operations/_dispatch.py
Luthaf: ok!

Luthaf (approved):

Luthaf:
Yes, we had issues with readthedocs, we contacted them recently and they should have updated the runner used to build the docs. Hopefully future PR should behave better!

PR #342, opened 2023-08-24
Generalize and speed up math gradients

### LINKED ISSUE #335: Optimize and generalize gradients for math operations
Among the math operations, `pow()` is the only one that handles gradients efficiently and accounting for potential additional dimensions.
It would be nice to have the same behavior in `add()`, `subtract()`, `multiply()`, `divide()`.

frostedoyster:
Fixes #335.
Comment: the gradients of all the binary functions (+, -, /, *) will not work if the sparsity patterns of the gradients are different along their `sample` dimensions. This was the case also before this PR and I don't think it's trivial to fix it.

----
:books: Documentation preview :books:: https://equistore--342.org.readthedocs.build/en/342/


python/equistore-operations/equistore/operations/divide.py
```diff
@@ -113,29 +113,43 @@ def _divide_block_block(block_1: TensorBlock, block_2: TensorBlock) -> TensorBlo
...
-                gradient_2.samples.column("sample") == i_sample
-            )[0]
-
-            value_grad = (
-                -block_1.values[i_sample]
-                * gradient_2.values[i_sample_grad_2]
-                / block_2.values[i_sample] ** 2
+        gradient_values_1 = gradient_1.values
+        gradient_values_2 = gradient_2.values
+        if gradient_values_1.shape != gradient_values_2.shape:
```
Luthaf, line 118: from the code below, it looks like you are assuming more than the same shape, you are also assuming the same values in the same order
```suggestion
        if gradient_1.samples != gradient_2.samples:
```

Luthaf:
> Comment: the gradients of all the binary functions (+, -, /, *) will not work if the sparsity patterns of the gradients are different along their sample dimensions. This was the case also before this PR and I don't think it's trivial to fix it.

It should not be too hard to fix: everything missing can be considered 0, so the output gradient sample is the union of the input gradient samples, and one can use `Labels.union_and_mapping` to get the union and where to write the values corresponding to both inputs. But we can leave this to a future PR.

----------------

This PR needs to have some tests, maybe something with custom gradients having different number of components?

python/equistore-operations/equistore/operations/divide.py
frostedoyster: This has to be assumed, no? Otherwise, in all these binary operations, what are the output samples supposed to be... those coming from gradients_1 or gradients_2? See the construction of the output TensorBlocks (that I didn't touch)

Luthaf: I'm fine with making this assumption, but the code should check for the assumption it is making, and it was checking something different

python/metatensor-operations/metatensor/operations/add.py
```diff
@@ -106,6 +106,15 @@ def _add_block_block(block_1: TensorBlock, block_2: TensorBlock) -> TensorBlock:
...
 
+        if gradient_1.values.shape != gradient_2.values.shape:
+            raise ValueError(
+                "The two gradient blocks must have the same shape. "
+                "Different sparsity patterns along the samples dimension "
+                "are not supported."
+            )
+        if gradient_1.samples != gradient_2.samples:
+            raise ValueError("The two gradient blocks must have the same samples")
+
```
Luthaf, line 117: Both of these conditions are already checked in the main function with `_check_same_gradients_raise`, so instead of duplicating checks we can have an assert here (sorry, I should have caught that earlier)

```suggestion
        assert gradient_1.values.shape == gradient_2.values.shape
```

Luthaf: And similarly for the other functions!

frostedoyster: Ahh my bad, I should have noticed as well

Luthaf (approved):

PR #348, opened 2023-09-01
Improve <labels>_names methods

Luthaf:
- components_names now return a List[str] instead of List[list[str]]
- properties and samples names are now pluralized

the first change is for better torch script compatibility (the compiler does not know what to do with `components_names == [["something"]]`)

----
:books: Documentation preview :books:: https://metatensor--348.org.readthedocs.build/en/348/


frostedoyster (approved):
Everything looks good!

PR #351, opened 2023-09-04
Make linear algebra solvers torch-script-able

frostedoyster:
----
:books: Documentation preview :books:: https://metatensor--351.org.readthedocs.build/en/351/

### DIFF
--- a/python/metatensor-operations/metatensor/operations/lstsq.py
+++ b/python/metatensor-operations/metatensor/operations/lstsq.py
@@ -13,7 +13,7 @@ from ._utils import (
 def lstsq(
     X: TensorMap,
     Y: TensorMap,
-    rcond: Optional[float] = None,
+    rcond: Optional[float],
     driver: Optional[str] = None,
 ) -> TensorMap:
     r"""
--- a/python/metatensor-operations/metatensor/operations/lstsq.py
+++ b/python/metatensor-operations/metatensor/operations/lstsq.py
@@ -1,4 +1,5 @@
 import warnings
+from typing import List, Optional
 
 from . import _dispatch
 from ._classes import TensorBlock, TensorMap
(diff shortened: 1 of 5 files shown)


python/metatensor-operations/metatensor/operations/lstsq.py
```diff
@@ -9,7 +10,12 @@
 )
 
 
-def lstsq(X: TensorMap, Y: TensorMap, rcond, driver=None) -> TensorMap:
+def lstsq(
+    X: TensorMap,
+    Y: TensorMap,
+    rcond: Optional[float] = None,
```
Luthaf, line 16: I think this was intentionally not set to None by default, because numpy and torch have different behavior here

```suggestion
    rcond: Optional[float],
```

python/metatensor-operations/metatensor/operations/solve.py
```diff
@@ -93,7 +93,7 @@ def _solve_block(X: TensorBlock, Y: TensorBlock) -> TensorBlock:
     """
     # TODO handle properties and samples not in the same order?
 
-    if not np.all(X.samples == Y.samples):
+    if not _dispatch.all(X.samples.values == Y.samples.values):
```
Luthaf, line 96:
```suggestion
    if not X.samples == Y.samples:
```

this was an old leftover, we can now compare Labels directly

python/metatensor-operations/metatensor/operations/solve.py
```diff
@@ -106,7 +106,7 @@ def _solve_block(X: TensorBlock, Y: TensorBlock) -> TensorBlock:
             )
 
         for X_component, Y_component in zip(X.components, Y.components):
-            if not np.all(X_component == Y_component):
+            if not _dispatch.all(X_component.values == Y_component.values):
```
Luthaf, line 109: you can replace the loop with 

```py
if  X.components != Y.components:
```

python/metatensor-operations/metatensor/operations/solve.py
```diff
@@ -93,25 +93,18 @@ def _solve_block(X: TensorBlock, Y: TensorBlock) -> TensorBlock:
...
         )
 
-    if len(X.components) > 0:
-        if len(X.components) != len(Y.components):
+    for X_component, Y_component in zip(X.components, Y.components):
+        if X_component != Y_component:
             raise ValueError(
                 "X and Y blocks in `solve` should have the same components \
                 in the same order"
             )
```
Luthaf, line 106:
```suggestion
    if X.components != Y.components:
        raise ValueError(
            "X and Y blocks in `solve` should have the same components \
            in the same order"
        )
```

PR #353, opened 2023-09-04
Make `slice()` torch-scriptable

frostedoyster:
Failing at the moment due to a weird error given by `check_isinstance`.


python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -113,6 +113,19 @@ def bincount(input, weights: Optional[TorchTensor] = None, minlength: int = 0):
...
 
+def bool_array_like(bool_list: List[bool], like):
+    """ 
+    Converts the input list of bool to a numpy array or torch tensor
+    based on the type of `like`.
+    """
+    if isinstance(like, TorchTensor):
+        return torch.tensor(bool_list, device=like.device)
+    elif isinstance(like, np.ndarray):
+        return np.array(bool_list)
```
Luthaf, line 124: should this set a `dtype=np.bool/torch.bool` as well? Just to catch mistakes early if needed

python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -136,6 +149,19 @@ def eye_like(array, size: int):
...
+    """
+    if isinstance(like, TorchTensor):
+        return torch.full(shape, fill_value, dtype=like.dtype, device=like.device)
+    elif isinstance(like, np.ndarray):
+        return np.full(shape, fill_value)
+    else:
+        raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
 def list_to_array(array, data: Union[List[int], List[List[int]]]):
```
Luthaf, line 167: should we rename this one to `int_array_like`?

python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -146,6 +172,23 @@ def list_to_array(array, data: Union[List[int], List[List[int]]]):
         return np.array(data)
     else:
         raise TypeError(UNKNOWN_ARRAY_TYPE)
+    
+
+def mask(array, axis: int, mask):
```
Luthaf, line 177: what's the issue with `array[mask]`?

python/metatensor-operations/tests/split.py
```diff
@@ -520,7 +516,7 @@ def test_split_errors(self):
             ),
         self.assertEqual(
             str(cm.exception),
-            "each element in grouped_labels must be 'Labels', not <class 'str'>",
```
Luthaf, line 523: Instead of removing the error message context, the type check should only be done if not in torch script mode (since in torch script mode the types are checked statically)

python/metatensor-operations/metatensor/operations/_dispatch.py
frostedoyster: Good idea

python/metatensor-operations/metatensor/operations/_dispatch.py
frostedoyster: But it can also be a float, no?

frostedoyster: But it can also be a float, no?

python/metatensor-operations/metatensor/operations/_dispatch.py
frostedoyster: The issue is when doing properties. In that case, you would really like to do `array[..., mask]`, but indexing after an ellipsis is not supported in torchscript

python/metatensor-operations/metatensor/operations/_dispatch.py
Luthaf: data is defined a `int`, and I don't think we ever call it with float

python/metatensor-operations/metatensor/operations/_dispatch.py
Luthaf: ok, fair enough. Might be worth adding a small comment explaining that to the next person who will touch this code!

python/metatensor-operations/metatensor/operations/_dispatch.py
frostedoyster: Ok

python/metatensor-operations/metatensor/operations/_dispatch.py
frostedoyster: Will add the comment

Luthaf:
Rebased & force pushed

Luthaf (approved):

PR #359, opened 2023-09-07
Keep a reference to the Labels in Labels.values

### LINKED ISSUE #293: `Labels().values` as numpy arrays should wrap ExternalCPUArrays
Just as TensorBlock values are, when the `Labels(...).values` are a numpy array, they should wrap an external CPU array. This would prevent the following behaviour attributed to random memory access when the labels object is not stored in a variable and therefore has no reference:

```python
import numpy as np
import equistore
from equistore import Labels

vals = [0, 1, 2, 3, 4, 5]
Labels(names=["_"], values=np.array(vals).reshape(-1, 1)).values

>>> array([[          0],
       [          0],
       [-2085138416],
       [          2],
       [          2]], dtype=int32)
```

where instead the correct values are only returned if instead the Labels object is assigned to a variable first:

```python
labels = Labels(names=["_"], values=np.array(vals).reshape(-1, 1))
labels.values
>>>array([[0],
       [1],
       [2],
       [3],
       [4]], dtype=int32)
```

Luthaf:
Fix #293

----
:books: Documentation preview :books:: https://metatensor--359.org.readthedocs.build/en/359/

### DIFF
--- a/docs/src/reference/python/labels.rst
+++ b/docs/src/reference/python/labels.rst
@@ -4,3 +4,12 @@ Labels
 .. autoclass:: metatensor.Labels
     :members:
     :undoc-members:
+
+
+.. autoclass:: metatensor.core.labels.LabelsEntry
+    :members:
+    :undoc-members:
+
+
+.. autoclass:: metatensor.core.labels.LabelsValues()
+    :show-inheritance:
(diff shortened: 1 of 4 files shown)


jwa7 (approved):
Looks good, thanks @Luthaf !

PR #361, opened 2023-09-08
Move labels to device in `to()`

frostedoyster:
The code is a bit complicated:
- `view().to_owned()` used to make copies of labels (`Labels.to()` works in-place for now, perhaps it should return a different object like `Tensor.to()`)
- `Labels.to()` only accepts `torch.device` (not str), so I'm having to use a `torch.device` from a block

----
:books: Documentation preview :books:: https://metatensor--361.org.readthedocs.build/en/361/

### DIFF
--- a/metatensor-torch/include/metatensor/torch/labels.hpp
+++ b/metatensor-torch/include/metatensor/torch/labels.hpp
@@ -25,6 +25,10 @@ namespace details {
     /// strings or a tuple of strings into a `std::vector<std::string>`.
     /// `argument_name` is used in the error message.
     std::vector<std::string> normalize_names(torch::IValue names, const std::string& argument_name);
+
+    /// Transform a torch::IValue into a torch::Device. The input torch::IValue
+    /// must be either a string or a torch.device, otherwise a runtime error
+    /// will be raised.
     torch::Device normalize_device(torch::IValue device);
 }
 
@@ -99,7 +103,7 @@ public:
     TorchLabels rename(std::string old_name, std::string new_name) const;
 
     /// Move the values for these Labels to the given `device`
-    TorchLabels to(torch::Device device) const;
+    TorchLabels to(torch::IValue device) const;
 
     /// Get the values associated with a single dimension (i.e. a single column
     /// of `values()`) in these labels.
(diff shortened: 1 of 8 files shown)


frostedoyster:
Perhaps we should also check that everything is on the same device (and has the same dtype) when creating a metatensor.torch.TensorBlock/Map

Luthaf:
We can change all of these: make `Labels.to` return a new object, accept a string as well as a torch device, and add checks that everything is on the same device/has the same dtype when constructing TensorBlock/TensorMap. 

Do you want to give it a go? Otherwise I can do something quick

metatensor-torch/include/metatensor/torch/labels.hpp
```diff
@@ -25,6 +25,7 @@ namespace details {
     /// strings or a tuple of strings into a `std::vector<std::string>`.
     /// `argument_name` is used in the error message.
     std::vector<std::string> normalize_names(torch::IValue names, const std::string& argument_name);
+    torch::Device normalize_device(torch::IValue device);
```
Luthaf, line 28: Please add a small docstring for this function, in particular explaining what's epxected in the `torch::IValue`

metatensor-torch/include/metatensor/torch/labels.hpp
```diff
@@ -98,7 +99,7 @@ class METATENSOR_TORCH_EXPORT LabelsHolder: public torch::CustomClassHolder {
     TorchLabels rename(std::string old_name, std::string new_name) const;
 
     /// Move the values for these Labels to the given `device`
-    void to(torch::Device device);
+    TorchLabels to(torch::Device device) const;
```
Luthaf, line 102:
```suggestion
    TorchLabels to(torch::IValue device) const;
```

And then call `normalize_device` internally. This way we don't need an additional lambda function in `register.cpp`

Luthaf:
> add checks that everything is on the same device/has the same dtype when constructing TensorBlock/TensorMap.

this is not in the PR, do you want to open an issue for it?

Luthaf (approved):

PR #363, opened 2023-09-08
Port split tests to pytest

### LINKED ISSUE #176: Porting `equistore.operations` to `pytest`

Luthaf:
This is the last one!

fix #176

----
:books: Documentation preview :books:: https://metatensor--363.org.readthedocs.build/en/363/

### DIFF
--- a/python/metatensor-operations/tests/split.py
+++ b/python/metatensor-operations/tests/split.py
@@ -1,623 +1,293 @@
 import os
-import unittest
-from typing import List, Union
 
 import numpy as np
+import pytest
 
 import metatensor
-from metatensor import Labels, TensorBlock, TensorMap
+from metatensor import Labels
 
 
 DATA_ROOT = os.path.join(os.path.dirname(__file__), "data")
-TEST_FILE_1 = "qm7-spherical-expansion.npz"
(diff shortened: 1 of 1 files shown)


jwa7 (approved):
Looks good, thanks for doing this! :)

PR #372, opened 2023-09-14
Make the `metatensor` python distribution an empty meta-package

### LINKED ISSUE #369: Improve error message when metatensor-core without metatensor is installed
We had the issue with @bananenpampe when metatensor-core is installed (for example when installing rascaline) and one tries to import TensorMap, the error message is confusing with its `unknown location`. For me it pin points more to a broken installation than a missing package.
```python
Cell In[4], line 1
----> 1 from metatensor import TensorMap

ImportError: cannot import name 'TensorMap' from 'metatensor' (unknown location)
```
The `unknown location` could be fixed by just adding an `__init__.py` into the `python/metatensor-core/metatensor`. But I guess there is a reason why we do not have an `__init__.py` with a `from .core import *` in the metatensor-core package, but only in the metatensor package? Maybe you have ideas to improve the error message somehow nevertheless.

Luthaf:
The corresponding code is now in `metatensor-core`, meaning users can alway `import Labels from metatensor` regardless of installed distributions.

I also moved all code in `metatensor.core` back in `metatensor`, since there is no longer a need for separate module.

This might be slightly confusing, since the `metatensor` distribution installs nothing, and `metatensor-core` installs the `metatensor` package, but I still think it is better than what we currently have.

Fix #369

----
:books: Documentation preview :books:: https://metatensor--372.org.readthedocs.build/en/372/


python/metatensor-core/metatensor/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+# version number and dependency management:
+#
+# - the `metatensor-core` distribution contains the Python bindings to the
+#   metatensor-core C API in the `metatensor` python package. External package
+#   integrating with metatensor such as rascaline only depend on metatensor-core.
+# - the `metatensor-operations` distribution contains the operations to manipulate
+#   metatensor data. Is is using Python's namespace packages
+#   (https://peps.python.org/pep-0420/) to install itself in `metatensor/operations`.
+# - the `metatensor-torch` distribution contains the TorchScript bindings to the C API.
+#   it installs itself in `metatensor/torch`.
```
agoscinski, line 11: maybe move this paragraph to the `./core/__init__.py` ?

agoscinski (commented):
The package installation seems a bit broken:

I installed package from the root folder `pip install .` and the packages are empty. I can import `metatensor`  and `metatensor.core`, but there are no classes I can import. Then I reinstalled metatensor-core from the respective python directory.  Then `metatensor` works as expected, but `metatensor.core` is still empty

Luthaf:
> The package installation seems a bit broken:

Did you run `./scripts/clean-python.sh` first? Otherwise the install can pick up files from their previous location.

> but metatensor.core is still empty

`metatensor.core` no longer exists, everything has been moved back inside `metatensor`

agoscinski (approved):
> Did you run ./scripts/clean-python.sh first? Otherwise the install can pick up files from their previous location.

Thanks, you are right. Works now.

What is still confusing that `pip uninstall metatensor` is not uninstalling anything. I saw in other packages that uninstalling the main package, also uninstalls all extra packages. I am not sure if this was already considered, but could not been done because of other problems. Maybe its worth a follow up issue? In any case this is not related to this PR.

Luthaf:
> What is still confusing that pip uninstall metatensor is not uninstalling anything

Very true. I don't know if this is possible though, I'll look it up.

PR #373, opened 2023-09-15
make LabelsEntry available to import from core

agoscinski:
Minor issue, low, but seems like an easy fix. To import the LabelsEntry from core one has to do
```python
from metatensor.core.labels import LabelsEntry
```
while from torch it can be imported like
```python
from metatensor.torch import LabelsEntry
```
To me it seems intuitive if we could also import LabelsEntry from core like this
```python
from metatensor.core import LabelsEntry
```

----
:books: Documentation preview :books:: https://metatensor--373.org.readthedocs.build/en/373/

### DIFF
--- a/python/metatensor-core/metatensor/__init__.py
+++ b/python/metatensor-core/metatensor/__init__.py
@@ -18,7 +18,7 @@
 from .version import __version__  # noqa
 from . import utils  # noqa
 from .block import TensorBlock  # noqa
-from .labels import Labels  # noqa
+from .labels import Labels, LabelsEntry  # noqa
 from .status import MetatensorError  # noqa
 from .tensor import TensorMap  # noqa
 


python/metatensor-core/metatensor/core/__init__.py
```diff
@@ -1,13 +1,14 @@
...
-from .labels import Labels
+from .labels import Labels, LabelsEntry
 from .status import MetatensorError
 from .tensor import TensorMap
 
 from .io import load, save
 
 __all__ = [
     "Labels",
+    "LabelsEntry",
```
Luthaf, line 11: I would not put it in `__all__` (this mainly control what gets imported when the user does `from metatensor import *`)

agoscinski: ok
```suggestion
```

Luthaf (approved):

PR #379, opened 2023-09-19
Preserve device in Labels::union and intersection

Luthaf:
Built on top of #375 and #378 which should be merged first

~TODO: figure a way to test this? Since we don't have cuda on CI~


Luthaf:
> figure a way to test this? Since we don't have cuda on CI

It seems that we can use the `"meta"` device (device which only tracks size) for these tests!

frostedoyster (approved):
I didn't know the "meta" device, it sounds funny in this contest. Perhaps to be used more in tests that deal with devices

PR #380, opened 2023-09-20
Fix performance issues coming from backtrace collection

Luthaf:
Both during iteration over labels and `tensor.block(key)`. This brings this code

```py
for key in tensor.keys:
    block = tensor.block(key)
```

from 40 ms to 2.5 ms (with a tensor containing 30 block) when using metatensor-torch. This is still a bit slower than metatensor-core at 600 us, but I can't find obvious performance problem in a profiler.

----
:books: Documentation preview :books:: https://metatensor--380.org.readthedocs.build/en/380/

### DIFF
--- a/metatensor-torch/src/labels.cpp
+++ b/metatensor-torch/src/labels.cpp
@@ -702,6 +702,30 @@ std::string LabelsHolder::repr() const {
 
 /******************************************************************************/
 
+LabelsEntryHolder::LabelsEntryHolder(TorchLabels labels, int64_t index):
+    labels_(std::move(labels))
+{
+    auto size = labels_->values().size(0);
+    if (index < -size || index >= size) {
+        // We prefer not using `C10_THROW_ERROR` here since this is an expected
+        // error (it will and must happen once per iteration over the labels),
+        // and `C10_THROW_ERROR` is quite expensive since it construct a full
+        // backtrace of the C++ code.
+        //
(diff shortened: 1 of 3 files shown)


metatensor-torch/src/labels.cpp
```diff
@@ -702,6 +702,30 @@ std::string LabelsHolder::repr() const {
 
 /******************************************************************************/
 
+LabelsEntryHolder::LabelsEntryHolder(TorchLabels labels, int64_t index):
+    labels_(std::move(labels))
+{
+    auto size = labels_->values().size(0);
+    if (index < -size || index >= size) {
```
PicoCentauri, line 709: why does this only raises an error on windows?

PicoCentauri (commented):
mhuch easier code is always nice! just a minor question.

metatensor-torch/src/labels.cpp
Luthaf: The other way around (`ifndef`, not `ifdef`). The constructor of IndexError taking two strings is not available on windows (it would have to be explicitly exported by torch), so in this case we let the error happen with a full backtrace in the indexing code.

PicoCentauri: Ahh okay, I see now the comment also makes more sense!

PicoCentauri (approved):

PR #384, opened 2023-09-26
More package metadata

Luthaf:
The root README needs to be markdown instead of rst since Cargo only supports markdown.

----
:books: Documentation preview :books:: https://metatensor--384.org.readthedocs.build/en/384/


PicoCentauri:
> The root README needs to be markdown instead of rst since Cargo only supports markdown.

WHAAAAT!!!

README.md
```diff
@@ -0,0 +1,29 @@
+# Metatensor
```
PicoCentauri, line 1: Aren't importing this somewhere in our docs? If yes, this will be broken now. If no we are fine.

agoscinski:
> Metatensor is a self-describing sparse tensor data format supporting gradients for atomistic machine learning and beyond.

For me the gradient support is an essential part that contrasts this package from others. But maybe the description now too long. Feel free to use your version also, if you feel like this not suitable. Don't want to overthink this one line description.

README.md
Luthaf: Fairly sure we don't import the README, only Contributing

Luthaf:
I could do something like "Metatensor is a self-describing sparse tensor data format for atomistic machine learning and beyond; storing values and gradients of these values together." in the README, and keep the shorter one for the different packages descriptions.

> > The root README needs to be markdown instead of rst since Cargo only supports markdown.
> 
> WHAAAAT!!!

Yep, rst is almost only used in the Python ecosystem. Everyone else is using markdown.

README.md
PicoCentauri: okay then I am fine.

PicoCentauri:
> I could do something like "Metatensor is a self-describing sparse tensor data format for atomistic machine learning and beyond; storing values and gradients of these values together." in the README, and keep the shorter one for the different packages descriptions.

Yes fine for me.

>>> The root README needs to be markdown instead of rst since Cargo only supports markdown.

>> WHAAAAT!!!

>Yep, rst is almost only used in the Python ecosystem. Everyone else is using markdown.

I know and it is fine for to change it.

PicoCentauri (approved):

PR #387, opened 2023-09-27
Add a test building metatensor-torch from sdist

Luthaf:
Since building metatensor-torch wheel requires a version of metatensor-torch available through pip, we use `dir2pi` from https://pypi.org/project/pip2pi/ to generate a pip-compatible server serving freshly built wheels. The tests also moved to a bash script since we need to start an HTTP server in the background, and cleanup once we exit.

----
:books: Documentation preview :books:: https://metatensor--387.org.readthedocs.build/en/387/

### DIFF
--- a/scripts/build-python.sh
+++ b/scripts/build-python.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+
+set -eux
+
+ROOT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")"/.. && pwd)
+cd "$ROOT_DIR"
+
+TMP_DIR="$1"
+rm -rf "$TMP_DIR"/dist
+
+./scripts/package-core.sh
+./scripts/package-torch.sh
+
+# check building sdist from a checkout, and wheel from the sdist
+python -m build python/metatensor-core --outdir "$TMP_DIR"/dist
+python -m build python/metatensor-operations --outdir "$TMP_DIR"/dist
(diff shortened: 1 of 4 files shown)


scripts/build-python.sh
```diff
@@ -0,0 +1,38 @@
...
+dir2pi --no-symlink "$TMP_DIR"/dist
+
+PYPI_SERVER_PID=""
+function cleanup() {
+    kill $PYPI_SERVER_PID
+}
+# Make sure to stop the Python server on script exit/cancellation
+trap cleanup INT TERM EXIT
+
+python -m http.server --directory "$TMP_DIR"/dist 8888 &
```
agoscinski, line 30: Maybe since this port is often busy with some jupyter notebook we do a check before?
```bash
PORT=8888
if [[ ! -z "$(netstat -nl | grep :$PORT)" ]]
then
    echo "An application is listening to port $PORT. Please free up the port first."  >&2
    exit 1
fi
```
https://serverfault.com/a/309111

agoscinski (commented):
Looks good to me. I am just worried that port 8888 is busy and the user get a not understandable error message. So I would propose to check if port is available. Maybe in addition we also choose some other default port. I see myself already always changing this file because I have a jupyter notebook running on this port.

scripts/build-python.sh
Luthaf: Very good point. I used `nc` since `netstat` is not installed everywhere, and I also changed the default port to something less likely to already be in use.

PR #394, opened 2023-10-05
Fix rust crate pre-release

Luthaf:
----
:books: Documentation preview :books:: https://metatensor--394.org.readthedocs.build/en/394/


Luthaf:
CI fails because we don't yet support torch 2.1, #395 should fix it!

PR #396, opened 2023-10-06
Suggest updating pip in installation instructions

Luthaf:
A lot of people are seeing issues because they have an old pip version, this should help!

----
:books: Documentation preview :books:: https://metatensor--396.org.readthedocs.build/en/396/

### DIFF
--- a/docs/src/get-started/installation.rst
+++ b/docs/src/get-started/installation.rst
@@ -24,6 +24,9 @@ From source:
 
 .. code-block:: bash
 
+    # Make sure you are using the latest version of pip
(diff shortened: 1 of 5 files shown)


PicoCentauri:
Does this also updates setuptools? I had once problems on a machine where setuptools was too old and it was not detected by our pyproject.toml

Luthaf:
Pip should install a recent setuptools when building the wheel. I've recently updated the bounds on setuptools version in pyproject.toml, so everything should be fine on this front

PicoCentauri:
Okay!

PR #399, opened 2023-10-10
Use {sample,component,property}_names

Luthaf:
----
:books: Documentation preview :books:: https://metatensor--399.org.readthedocs.build/en/399/


Luthaf:
This will be a breaking change, so I want to do it before the first release.

PicoCentauri (approved):
Reading the source code is more fluent in my opinion. I like it!

PR #404, opened 2023-10-20
Two fixes for sample reduction with torch

Luthaf:
In the metatensor-core version, the `index` must be converted to a torch tensor before passing it to `index_add`. Since it comes from the Labels, it is a numpy array by default.

In the metatensor-torch version, `unique_with_inverse` does not work when reducing to a single sample

----
:books: Documentation preview :books:: https://metatensor--404.org.readthedocs.build/en/404/

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -17,7 +17,7 @@ package = external
 package_env = build-metatensor-core
 lint_folders = "{toxinidir}/python" "{toxinidir}/setup.py"
 build_single_wheel = --no-deps --no-build-isolation --check-build-dependencies
-test_options = --cov={env_site_packages_dir}/metatensor --cov-append --import-mode=append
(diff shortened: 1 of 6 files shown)


tox.ini
```diff
@@ -72,14 +72,14 @@ commands =
...
 deps =
     torch
-    {[testenv:operations-notorch-tests]deps}
+    {[testenv:operations-numpy-tests]deps}
 changedir =
-    {[testenv:operations-notorch-tests]changedir}
+    {[testenv:operations-numpy-tests]changedir}
 commands =
-    {[testenv:operations-notorch-tests]commands}
+    {[testenv:operations-numpy-tests]commands}
```
jwa7, line 82: Should these not be `operations-torch-tests` ?

jwa7: (lines  76 - 82)

jwa7 (commented):
Just a question about the tox file:

PR #405, opened 2023-10-20
Atomistic models based on metatensor-torch

Luthaf:
This PR contains all the code required to define, export, load and validate arbitrary atomistic models based on metatensor-torch. The user has to provide a TorchScript-compatible `torch.nn.Module`, with the following signature:

```py
def forward(self, system: System, run_options: ModelRunOptions) -> Dict[str, TensorBlock]:
    ...
```

The `System` contains positions, cell and the neighbors lists requested by any submodules, which are requested by defining a `requested_neighbors_lists` function (this allow e.g. rascaline to request a NL without the end module knowing about it):

```py
def requested_neighbors_lists(self) -> List[NeighborsListOptions]:
    ...
```

The `ModelRunOptions` is what the engine wants; in particular it contains a list of `ModelOutput` (the model `forward` function can return multiple outputs, and the MD engine should request the ones it wants) and the set of `selected_atoms` on which to run the calculation.

When exporting a model, the user should use `ModelCapabilities` to declare what the model is able to do, as well as the units it uses as input & output. `MetatensorAtomisticModule` then does unit conversion between what the engine provides and the model wants for input; and what the engine wants and the model provides on output.

-----

Still TBD:

- [x] Python side documentation
- [x] Commented example on how to use this API
- [x] Decide on the naming for the new metadata classes (`ModelCapabilities`, `ModelOutput`, `ModelRunOptions`)
- [x] Define a standard for metatensor metadata of some outputs: what should be the samples/components/properties names and values for the `energy` output, for the `dipole` output, …
- [x] Can a model provide both per-atom and per-structure of the same quantity? How would this look like?
- [x] Document the standard above
- [x] Record the torch extension and torch version used when saving the model
- [ ] ~Add a way to profile model execution time~: will be done in a later PR
- [x] Provide a function to connect neighbors lists distances with positions & cell in the computational graph

----
:books: Documentation preview :books: https://metatensor--405.org.readthedocs.build/en/405/


python/metatensor-torch/metatensor/torch/atomistic/units.py
```diff
@@ -0,0 +1,40 @@
...
+            "nanometer": 10.0,
+            "nm": 10.0,
+        },
+    ),
+    "energy": Quantity(
+        baseline="ev",
+        conversions={
+            "ev": 1.0,
+            "mev": 1e-3,
+            "hartree": 27.211396641308,
```
PicoCentauri, line 37: kj/mol maybe

python/metatensor-torch/metatensor/torch/atomistic/units.py
```diff
@@ -0,0 +1,40 @@
...
+        self.conversions = conversions
+
+        assert self.baseline in self.conversions
+        assert self.conversions[self.baseline] == 1.0
+
+
+KNOWN_QUANTITIES = {
+    "length": Quantity(
+        baseline="angstrom",
+        conversions={
```
PicoCentauri, line 25: This looks a bit over engineering for me and I don't understand why is there a baseline argument. We just define our baseline argument and everything else is converted to this. A but like `scipy.constants`.

python/metatensor-torch/metatensor/torch/atomistic/model.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+class MetatensorAtomisticModule(torch.nn.Module):
+    """TODO"""
+
+    # Some annotation to make the TorchScript compiler happy
+    _requested_neighbors_lists: List[NeighborsListOptions]
+    _known_quantities: Dict[str, Quantity]
+
+    def __init__(self, module: torch.nn.Module, capabilities: ModelCapabilities):
```
PicoCentauri, line 17: `capabilities` is something like the target? i.e. "forces", "dipole moments", "partial charges"?

PicoCentauri (commented):
I have to wrap my head around this. The syntax and the logic, but could be very good.

I have some first specific comments in the units.

python/metatensor-torch/metatensor/torch/atomistic/units.py
Luthaf: yes, we'll need a lot more units & quantities here!

python/metatensor-torch/metatensor/torch/atomistic/units.py
Luthaf: The baseline is just here to clarify the conversion factors. We could maybe define something nicer based on scipy.constants, but we need this code to be compatible with torchscript.

We don't (and can't) impose units on either the model or the MD engine, so we can not convert everything to internal units. Instead, we need to know about all the possible units at run time to convert between the engine and model units.

python/metatensor-torch/metatensor/torch/atomistic/model.py
Luthaf: `capabilities.outputs` contains the different targets the model is able to produce. `capabilities` contains other information about the model (unit used as input, species it can handle).

Maybe we should rename this to `ModelDefinition` or something, and also store in there the model authors, papers to cite, date of training, etc.

agoscinski: I also stumbled on this term, I thought about `ModelConfig`, but the term does not capture really that in includes the model outputs.

> Maybe we should rename this to ModelDefinition or something, and also store in there the model authors, papers to cite, date of training, etc.
 
I like `ModelDefinition`, then this would include a metainformation variable of type a `Dict[str, str]` ?

Luthaf: I would prefer to give more structure to the data, and have multiple fields of type `str` instead of a dict. We could have a `other` field if people want to store more data in there, but I would start without this for now.

agoscinski: Maybe we could consider metadata that is included in models from PyTorch Hub or hugging face
https://pytorch.org/docs/stable/hub.html#torch.hub.load
https://huggingface.co/docs/transformers/v4.34.1/en/model_doc/auto#transformers.AutoModel.from_pretrained

I identify these parameters
- `provider` hub platform (e.g. "Hugging Face")
- `repo_url` (e.g. "https://camembert-model.fr")
- `model_name`  (e.g. "camenbert")
- `model_checkpoint` (e.g. "camembert-large")

agoscinski: Might be also useful to consider, the metadata that is contained in ONNX models https://onnx.ai/onnx/intro/concepts.html#metadata I identified these as possibly useful

- __producer_version__: The version of the generating tool.
like version of the model or commit ID
- __model_version__: The version of the model itself, encoded in an integer.
I think this as basically what `model_checkpoint` above is expressing, just that they express it as a string instead of a number. If you look at models from the computer vision community https://pytorch.org/vision/stable/models.html it also makes more sense. For example the first model the `model_name`  would be `AlexNet`  and the version would be `AlexNet_Weights.IMAGENET1K_V1`. I think that would be good enough to track models. Sometimes they use the `_VX` at the end to describe that the training procedure changed a bit, but sometimes they change the name at the beginning to do that. It is a mess.
- __model_license__: The well-known name or URL of the license under which the model is made available.
Not sure if this makes sense, as the license should be stored in the repo where the model comes from I would say
- __doc_string__: Human-readable documentation for this model

Luthaf: I'm not sold on the hugging face style metadata, it feels more related to a full repository of models instead of a single one. The ONNX metatadata makes a lot more sense to me.

frostedoyster:
Do I understand correctly that one "System" is one structure?
How much work would it be to change the signature to the following?
```py
def forward(self, systems: List[System], run_options: ModelRunOptions) -> Dict[str, TensorBlock]:
```
This would allow a single model to be trained and exported, since the forward needs to take in multiple structures during training. At the moment, I can only see this working if, after having trained the model, you manually convert it to a different class that has the `system: System` signature as the forward function, unless I'm missing something.
For now, we could have the interface fail if `len(systems) != 1`, and in the future this would facilitate PIMD and other techniques where the MD engine can ask for multiple evaluations at the same time

Luthaf:
We had a couple of discussion on taking a single system or multiple, but I can't remember the argument for going with a single one. I agree having more than one system is quite useful here, I'll give it a go and check if everything works the same.

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -0,0 +1,322 @@
...
+            assert len(energy) == 1
+            assert energy.sample_names == ["structure", "atom"]
+            assert torch.all(energy.block().samples["structure"] == 0)
+            assert torch.all(
+                energy.block().samples["atom"] == torch.arange(positions.shape[0])
+            )
+            energies = energy.block().values
+            assert energies.shape == (len(atoms), 1)
+
+            energy = sum_over_samples(energy, sample_names=["atom"])
```
Luthaf, line 195: @frostedoyster regarding how to handle having the same output both per-atom and per-structure, this is how it's done in the ASE calculator. Feels to me that this is simple enough to not require more infrastructure than what we have. If the end user wants both, compute per-atom & then sum over samples to get the per-structure.

Luthaf:
This is now ready for review! I know this is a very large PR, but each commit should be reviewable independently.

python/metatensor-torch/metatensor/torch/atomistic/model.py
Luthaf: We currently collect some minimal metadata, which should be expanded with stuff like this. If you agree @agoscinski I would leave the metadata definition to a later PR?

Luthaf:
Yes, we'll have to do a release for rascaline as well. Unfortunately, this will not mean we'll have pre-built versions, since the code depends on PyTorch and we can not have tags for PyTorch versions in the wheel name.

PicoCentauri:
Okay, but at least no more hardcoded github links :-P

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -0,0 +1,331 @@
...
+        outputs = self._model(
+            [system],
+            run_options,
+            check_consistency=self.parameters["check_consistency"],
+        )
+        energy = outputs["energy"]
+
+        if run_options.outputs["energy"].per_atom:
+            assert len(energy) == 1
+            assert energy.sample_names == ["structure", "atom"]
```
frostedoyster, line 196: In our models, it would be easier to have the samples be `["structure", "center"]`. This would be cleaner on the model side, so we don't have to re-create a TensorMap. It also fits multi-center properties perhaps, where the samples might be `["structure", "center_1", "center_2"]`...

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
frostedoyster: Yes, I feel like most models do the summation internally. In that case, they can interface with us (wrap) and they won't support per-atom energies. So it seems to me that, with the current design, the summation is supposed to be done by the model, while any gradient calls would happen outside. I still think it would be cleaner (in metatensor-models) to have the model output the forces as well though, using the "explicit gradients" that we have here

metatensor-torch/src/atomistic.cpp
```diff
@@ -0,0 +1,986 @@
+#include <cctype>
+#include <cstring>
+
+#include <array>
+#include <sstream>
+#include <algorithm>
+
+#include <torch/torch.h>
+#include <nlohmann/json.hpp>
```
frostedoyster, line 9: I'm worried about the compilation time with this new dependency. I assume that this header includes a lot of smaller headers; can we get the same functionality using one of the smaller headers?

metatensor-torch/include/metatensor/torch/atomistic.hpp
```diff
@@ -0,0 +1,388 @@
...
+}
+
+
+/// Custom autograd node going (positions, cell) => neighbors list distances
+///
+/// This does not actually computes the neighbors list, but registers a new node
+/// in the overall computational graph, allowing to re-use neighbors list
+/// computed outside of torch. The public interface for this is the
+/// `register_autograd_neighbors()` function.
+class METATENSOR_TORCH_EXPORT NeighborsAutograd: public torch::autograd::Function<NeighborsAutograd> {
```
frostedoyster, line 102: What does this do? Isn't the model supposed to handle this part using torch operations? What use case were you thinking of?

metatensor-torch/include/metatensor/torch/atomistic.hpp
```diff
@@ -0,0 +1,388 @@
...
+    std::vector<NeighborsListOptions> known_neighbors_lists() const;
+
+    /// Add custom data to this system, stored as `TensorBlock`.
+    ///
+    /// This is intended for experimentation with models that need more data as
+    /// input, and moved into a field of `SystemHolder` later.
+    ///
+    /// @param name name of the data
+    /// @param values values of the data
+    void add_data(std::string name, TorchTensorBlock values);
```
frostedoyster, line 218: Do we have particular types of data in mind here? I would do `TensorMap`s to have more flexibility (e.g., two quantities that include components of different symmetries (L) each)

metatensor-torch/src/atomistic.cpp
```diff
@@ -0,0 +1,986 @@
...
+
+    auto it = data_.find(name);
+    if (it == data_.end()) {
+        C10_THROW_ERROR(ValueError,
+            "no data for '" + name + "' found in this system"
+        );
+    }
+
+    TORCH_WARN_ONCE(
+        "custom data '", name, "' is experimental, please contact the developers ",
```
frostedoyster, line 543: I would say developers of metatensor here, maybe random people will think this is a torch error and they will contact the torch developers.
Unrelated but https://github.com/google-research/bert/issues/1396

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
Luthaf: I'm not sure I'm following. If the model does not support per-atom output, it can declare it, and then we will raise an error if the engine request per-atom output.

Same with explicit gradients: if the model supports it, it can declare it and then the engine can use it. Otherwise the engine falls back to backward propagation, which should always be supported.

metatensor-torch/src/atomistic.cpp
Luthaf: It did not felt to me a lot slower to compile here. In practise, I find that torch headers contain a lot more code and are the main part slowing down compilation here. I'll run some compile-time tests here.

We could use something smaller like https://github.com/DaveGamble/cJSON, but it's a pure C library, so the code is going to get a lot more convoluted.

metatensor-torch/include/metatensor/torch/atomistic.hpp
Luthaf: >  What use case were you thinking of?

> allowing to re-use neighbors list computed outside of torch.

I.e. use a NL computed in LAMMPS, but pretend that it was computed with pytorch for autograd purposes. This avoid having to re-do the `positions[j] - positions[i] + shift @ cell` calculation, since we already have the corresponding values.

metatensor-torch/src/atomistic.cpp
Luthaf: fair point!

metatensor-torch/src/atomistic.cpp
frostedoyster: Ok... I just assumed torch wasn't a problem because it's compiled, but I guess it's very large

metatensor-torch/include/metatensor/torch/atomistic.hpp
frostedoyster: Ok, good!

python/metatensor-torch/metatensor/torch/atomistic/documentation.py
```diff
@@ -111,8 +111,13 @@ def known_data(self) -> List[str]:
 class NeighborsListOptions:
     """Options for the calculation of a neighbors list"""
 
-    def __init__(self, model_cutoff: float, full_list: bool):
-        ...
+    def __init__(self, model_cutoff: float, full_list: bool, requestor: str = ""):
```
PicoCentauri, line 123: Below you write that there is a list of requestors and a new one can be added with `add_requestor`. Why do we need this initial argument `requestor` wouldn't it be a bit more consistent to have an API like

```python
ns = NeighborsListOptions(model_cutoff=5, full_list=False)
ns. add_requestor("my_module")
```

metatensor-torch/src/atomistic.cpp
```diff
@@ -111,71 +113,81 @@ NeighborsListOptions NeighborsListOptionsHolder::from_json(const std::string& js
...
+    if (!is_floating_point(positions_.scalar_type())) {
         C10_THROW_ERROR(ValueError,
-            "invalid components for `cell`: the first component should be 'cell_abc'=[0, 1, 2]"
+            "`positions` must be a tensor of floating point data, got " +
+            scalar_type_name(positions_.scalar_type()) + " instead"
         );
     }
 
-    if (*components[1] != metatensor::Labels({"xyz"}, {{0}, {1}, {2}})) {
+    if (cell_.sizes().size() != 2) {
```
PicoCentauri, line 345: maybe explicitly check for  a 3x3 matrix here?

python/metatensor-torch/metatensor/torch/atomistic/documentation.py
```diff
@@ -0,0 +1,293 @@
+from typing import Dict, List, Optional
+
+import torch
+
+from ..documentation import Labels, TensorBlock
+
+
+class System:
```
PicoCentauri, line 8: Do we also want to provide converters as we do it in rascaline: i.e from a ase object to an systems object?

PicoCentauri (changes requested):
Very nice. I think this will bring us very close to running our models.

The code is clean and function and classes definitions are well documented and tested.

What I am missing is an overall example most likely with the ase calculator, how everything works together. Maybe starting from ase list of atoms and then running this inside an `MetatensorCalculator`. You did this basically for a test but we should copy this also in an example.

python/metatensor-torch/metatensor/torch/atomistic/units.py
```diff
@@ -0,0 +1,71 @@
...
+            "angstrom": 1.0,
+            "bohr": 0.5291772105638411,
+            "nanometer": 10.0,
+            "nm": 10.0,
```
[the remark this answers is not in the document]
Luthaf: these are constants, not sure against what you'd want to test them 😄

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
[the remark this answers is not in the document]
Luthaf: I think the i-PI calculator needs more than one file (one in i-PI and one here). But this is definitively needed in a future PR!

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -0,0 +1,341 @@
...
+    InputError,
+    PropertyNotImplementedError,
+    all_properties as ALL_ASE_PROPERTIES,
+)
+
+
+# import here to get an error early if the user is missing metatensor-operations
+from .. import sum_over_samples_block  # isort: skip
+
+FilePath = Union[str, bytes, pathlib.PurePath]
```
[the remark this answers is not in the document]
Luthaf: that was my understanding yes. PurePath is a base of both PosixPath and WindowsPath

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -0,0 +1,341 @@
...
+    """
+
+    def __init__(
+        self,
+        model: Union[
+            FilePath,
+            torch.jit.RecursiveScriptModule,
+            MetatensorAtomisticModule,
+        ],
+        check_consistency=False,
```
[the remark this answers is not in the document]
Luthaf: I'd say this is a bit different: `check_consistency` is used to check that metatensor-atomistic-models standard is respected (no missing neighbors list, right species, right output shape & metadata) and give a nice error message in case they are not.

The operations checks are there for more "unsafe" checks: skipping them might produce wrong results. Here you'll get a weird error deep down in the model instead. 

We might look for a way to unify both (with some form of check severity maybe?), but since the operations checks are not used for now, I would leave it for a later PR.

tox.ini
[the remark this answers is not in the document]
Luthaf: I'd say no: the code we are providing is only useful to people who are already using ase. So either you have ase and you can use the calculator, or you don't and then you don't care. But the user is directly using ase here.

metatensor-torch/src/atomistic.cpp
Luthaf: just below

python/metatensor-torch/metatensor/torch/atomistic/model.py
```diff
@@ -0,0 +1,386 @@
...
+    ...                 properties=Labels(["energy"], torch.IntTensor([[0]])),
+    ...             )
+    ...
+    ...         return outputs
+    ...
+
+    Wrapping and exporting this model would then look like this:
+
+    >>> import os
+    >>> import tempfile
```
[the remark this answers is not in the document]
PicoCentauri: Okay for me.

python/metatensor-torch/tests/atomistic/ase_calculator.py
```diff
@@ -0,0 +1,189 @@
...
+)
+from metatensor.torch.atomistic.ase_calculator import MetatensorCalculator
+
+
+ase = pytest.importorskip("ase")
+import ase.build  # noqa  isort: skip
+import ase.calculators.lj  # noqa isort: skip
+
+
+class LennardJones(torch.nn.Module):
```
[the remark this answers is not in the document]
PicoCentauri: I think for an example it doesn't have to be super fast but more readable, and if it working well enough for lets say 20 particles it is fine at least for me. But, I let you decide.

Vectorization could help for sure.

python/metatensor-torch/metatensor/torch/atomistic/documentation.py
PicoCentauri: Okay, for a compromise we could add a comment stating like:

```rst
:param requestor: Who requested this neighbors list. You can add additional requestors
     later using :meth:`add_requestor`.
```

metatensor-torch/src/atomistic.cpp
PicoCentauri: ahh yes 🙈

python/metatensor-torch/metatensor/torch/atomistic/documentation.py
PicoCentauri: Okay, 100% okay for me. Once we replaced the system in rascaline we could add a comment here that some converters exist in the rascaline or the `metatensor-models` package.

python/metatensor-torch/metatensor/torch/atomistic/units.py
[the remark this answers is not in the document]
Luthaf: I used ASE constants, since they already have atomic-related units.

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -0,0 +1,341 @@
...
+    def todict(self):
+        # used by ASE to save the calculator
+        raise NotImplementedError("todict is not yet implemented")
+
+    @classmethod
+    def fromdict(cls, dict):
+        # used by ASE to load a saved calculator
+        raise NotImplementedError("fromdict is not yet implemented")
+
+    def run_model(
```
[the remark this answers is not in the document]
Luthaf: Not currently, but ASE's `Calculator` has a lot of functions already, so I would rather make sure we don't run into conflict down the line.

This is mainly intended as a way to run a model on ASE data easily. We could also rename to `run_metatensor_model` or something more specific.

python/metatensor-torch/tests/atomistic/ase_calculator.py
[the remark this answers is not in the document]
Luthaf: My issue is with people copy-pasting the example as-is and ignoring any kind of warning we put on it. I'll see what I can do for an example!

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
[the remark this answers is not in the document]
PicoCentauri: Okay, can we maybe add this to the docstring that we do not forget and blindly rename this at some point?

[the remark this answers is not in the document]
Luthaf: The docstring is now 

```
Run the model on the given ``atoms``, computing properties according to the
``outputs`` and ``selected_atoms`` options.

The output of the model is returned directly, and as such the blocks' ``values``
will be :py:class:`torch.Tensor`.

This is intended as an easy way to run metatensor models on
:py:class:`ase.Atoms` when the model can predict properties not supported by the
usual ASE's calculator interface.
```

Anything else I should put in here?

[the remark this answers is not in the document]
PicoCentauri: Nope that is good.

PicoCentauri (approved):
I am happy.

Luthaf:
I found another small bug, I'll wait for CI to pass & then this should be good to go!

PR #406, opened 2023-10-23
Improve usage of metatensor in a cmake subdirectory

Luthaf:
----
:books: Documentation preview :books:: https://metatensor--406.org.readthedocs.build/en/406/


metatensor-core/CMakeLists.txt
```diff
@@ -69,18 +69,16 @@ set(EXTRA_RUST_FLAGS "" CACHE STRING "Flags used to build rust code")
...
-    if("${CMAKE_BUILD_TYPE}" STREQUAL "" AND "${CMAKE_CONFIGURATION_TYPES}" STREQUAL "")
-        message(STATUS "Setting build type to 'release' as none was specified.")
-        set(CMAKE_BUILD_TYPE "release"
-            CACHE STRING
-            "Choose the type of build, options are: debug or release"
-        FORCE)
-        set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS release debug)
-    endif()
+if("${CMAKE_BUILD_TYPE}" STREQUAL "" AND "${CMAKE_CONFIGURATION_TYPES}" STREQUAL "")
+    message(STATUS "Setting build type to 'release' as none was specified.")
```
PicoCentauri, line 73: Should release really the default?

metatensor-torch/tests/cmake-project/CMakeLists.txt
```diff
@@ -2,14 +2,27 @@ cmake_minimum_required(VERSION 3.16)
...
+    add_subdirectory(../../../ metatensor)
+    find_package(Torch)
+
+    # load metatensor and metatensor-torch from the build path
+    set(CMAKE_BUILD_RPATH "$<TARGET_FILE_DIR:torch>;$<TARGET_FILE_DIR:metatensor::shared>;$<TARGET_FILE_DIR:metatensor_torch>")
+else()
+    # If building a dev version, we also need to update the REQUIRED_METATENSOR_VERSION
+    # in the same way we update the metatensor-torch version
+    include(../../cmake/dev-versions.cmake)
+    set(REQUIRED_METATENSOR_TORCH_VERSION "0.1.0")
```
PicoCentauri, line 19: Couldn't we put a variable here. To avoid changing the version in many places when we bump the version?

PicoCentauri (commented):
Building in subdirs make sense. I have some minor ideas of improvements.

metatensor-core/CMakeLists.txt
Luthaf: What else would you want to see? I think we need to build with optimizations by default so random users get the full performance of the code. We could also include some basic debug information, but we don't really make use of it anywhere.

metatensor-torch/tests/cmake-project/CMakeLists.txt
Luthaf: I don't think so. This is the same as pip dependencies declaration around here: https://github.com/lab-cosmo/metatensor/blob/b5480cbd8f5c2fc4f59ea812e8c7cabd64caad5f/python/metatensor-operations/setup.py#L129, which we also need to update for every release.

PicoCentauri: okay

metatensor-core/CMakeLists.txt
PicoCentauri: Yeah most likely optimizing everything if nothing is given is a good idea.

PR #411, opened 2023-10-31
Fix x_to_y labels device

### LINKED ISSUE #409: Missing device check in TensorMap.keys_to_samples() [and probably some other operations]
Some operations (such as `tensormap.keys_to_samples()`) may re-initialize the internal `Labels` objects, which have the default "cpu" device. It results in a device conflict later on, if `metatensor.torch.operations.sum_over_samples()` or some other type of index-based operation is used. 

---

**Example**: 

```
tensor_2 = tensor_1.keys_to_samples('keys')
tensor_3 = metatensor.torch.operations.sum_over_samples(tensor_2, ['samples'])
```

will result in the following error, while using GPU as a primary device

```
metatensor.torch.operations.sum_over_samples(
  File "/home/mazitov/apps/anaconda3/envs/torch-alchemical/lib/python3.10/site-packages/metatensor/operations/reduce_over_samples.py", line 514, in sum_over_samples
    return _reduce_over_samples(
  File "/home/mazitov/apps/anaconda3/envs/torch-alchemical/lib/python3.10/site-packages/metatensor/operations/reduce_over_samples.py", line 369, in _reduce_over_samples
    _reduce_over_samples_block(
  File "/home/mazitov/apps/anaconda3/envs/torch-alchemical/lib/python3.10/site-packages/metatensor/operations/reduce_over_samples.py", line 170, in _reduce_over_samples_block
    _dispatch.index_add(
  File "/home/mazitov/apps/anaconda3/envs/torch-alchemical/lib/python3.10/site-packages/metatensor/operations/_dispatch.py", line 338, in index_add
    output_array.index_add_(0, index, input_array)
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument index in method wrapper_CUDA_index_add_)
```

frostedoyster:
Fixes #409. I've been able to write a test that fails but I don't know what the best way to fix the problem is

----
:books: Documentation preview :books:: https://metatensor--411.org.readthedocs.build/en/411/

### DIFF
--- a/metatensor-torch/src/tensor.cpp
+++ b/metatensor-torch/src/tensor.cpp
@@ -337,7 +337,8 @@ TorchTensorMap TensorMapHolder::keys_to_properties(torch::IValue keys_to_move, b
     } else if (keys_to_move.isCustomClass()) {
         auto selection = keys_to_move.toCustomClass<LabelsHolder>();
         auto tensor = tensor_.keys_to_properties(selection->as_metatensor(), sort_samples);
-        return torch::make_intrusive<TensorMapHolder>(std::move(tensor))->to(device);
+        auto result = torch::make_intrusive<TensorMapHolder>(std::move(tensor));
+        return result->to(device);
     } else {
         C10_THROW_ERROR(TypeError,
             "TensorMap::keys_to_properties first argument must be a `str`, list of `str` or `Labels`"
--- a/metatensor-torch/src/tensor.cpp
+++ b/metatensor-torch/src/tensor.cpp
@@ -350,7 +350,8 @@ TorchTensorMap TensorMapHolder::keys_to_samples(torch::IValue keys_to_move, bool
     if (keys_to_move.isString() || keys_to_move.isList() || keys_to_move.isTuple()) {
         auto selection = extract_list_str(keys_to_move, "TensorMap::keys_to_samples first argument");
         auto tensor = tensor_.keys_to_samples(selection, sort_samples);
-        return torch::make_intrusive<TensorMapHolder>(std::move(tensor))->to(device);
+        auto result = torch::make_intrusive<TensorMapHolder>(std::move(tensor));
+        return result->to(device);
     } else if (keys_to_move.isCustomClass()) {
         auto selection = keys_to_move.toCustomClass<LabelsHolder>();
         auto tensor = tensor_.keys_to_samples(selection->as_metatensor(), sort_samples);
--- a/metatensor-torch/src/tensor.cpp
+++ b/metatensor-torch/src/tensor.cpp
@@ -356,7 +356,8 @@ TorchTensorMap TensorMapHolder::keys_to_samples(torch::IValue keys_to_move, bool
     } else if (keys_to_move.isCustomClass()) {
         auto selection = keys_to_move.toCustomClass<LabelsHolder>();
         auto tensor = tensor_.keys_to_samples(selection->as_metatensor(), sort_samples);
-        return torch::make_intrusive<TensorMapHolder>(std::move(tensor))->to(device);
+        auto result = torch::make_intrusive<TensorMapHolder>(std::move(tensor));
+        return result->to(device);
     } else {
         C10_THROW_ERROR(TypeError,
             "TensorMap::keys_to_samples first argument must be a `str`, list of `str` or `Labels`"
--- a/metatensor-torch/src/tensor.cpp
+++ b/metatensor-torch/src/tensor.cpp
@@ -58,6 +58,17 @@ TorchTensorMap TensorMapHolder::copy() const {
     return torch::make_intrusive<TensorMapHolder>(this->tensor_.clone());
 }
 
+TorchTensorMap TensorMapHolder::to(torch::Device device) {
+    const auto keys = this->keys()->to(device);
+    auto blocks = std::vector<torch::intrusive_ptr<TensorBlockHolder>>();
+    for (size_t i=0; i<this->tensor_.keys().count(); i++) {
+        blocks.push_back(
+            torch::make_intrusive<TensorBlockHolder>(this->tensor_.block_by_id(i), torch::IValue())->to(device)
+        );
+    }
+    return torch::make_intrusive<TensorMapHolder>(keys, blocks);
+}
+
 TorchLabels TensorMapHolder::keys() const {
     return torch::make_intrusive<LabelsHolder>(this->tensor_.keys());
 }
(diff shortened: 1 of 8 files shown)


frostedoyster:
https://github.com/lab-cosmo/metatensor/blob/ef15c7326ef5cc188d1d8f99370316d19379f239/metatensor-torch/src/tensor.cpp#L336-L350
Here, we would like to change the device of some of the labels. However, the new `Labels::to` returns a new labels object, so that we would have to rebuild all blocks and finally the `TensorMap` to be returned. Should we instead restore a `Labels::to_inplace` or something along those lines and use it here?

Luthaf:
I would prefer to keep the immutability of TensorMap & friends, and go for a rebuild of the blocks and TensorMap after moving to a new device. This could also be done though a new set of functions `TensorBlock::to` and `TensorMap::to` (also moving the data!)

Longer term we could discuss having ways to do this move in-place, but I would prefer to take some time to think about all the implications of it!

frostedoyster:
@Luthaf is this more or less what you meant?

metatensor-torch/src/tensor.cpp
```diff
@@ -318,14 +329,15 @@ static std::vector<std::string> extract_list_str(const torch::IValue& keys_to_mo
 }
 
 TorchTensorMap TensorMapHolder::keys_to_properties(torch::IValue keys_to_move, bool sort_samples) const {
+    auto device = this->keys()->values().device();
     if (keys_to_move.isString() || keys_to_move.isList() || keys_to_move.isTuple()) {
         auto selection = extract_list_str(keys_to_move, "TensorMap::keys_to_properties first argument");
         auto tensor = tensor_.keys_to_properties(selection, sort_samples);
-        return torch::make_intrusive<TensorMapHolder>(std::move(tensor));
+        return torch::make_intrusive<TensorMapHolder>(std::move(tensor))->to(device);
```
Luthaf, line 336:
```suggestion
        auto result = torch::make_intrusive<TensorMapHolder>(std::move(tensor));
        return result->to(device);
```

For more clarity

metatensor-torch/src/tensor.cpp
```diff
@@ -318,14 +329,15 @@ static std::vector<std::string> extract_list_str(const torch::IValue& keys_to_mo
...
     } else if (keys_to_move.isCustomClass()) {
         auto selection = keys_to_move.toCustomClass<LabelsHolder>();
         auto tensor = tensor_.keys_to_properties(selection->as_metatensor(), sort_samples);
-        return torch::make_intrusive<TensorMapHolder>(std::move(tensor));
+        return torch::make_intrusive<TensorMapHolder>(std::move(tensor))->to(device);
```
Luthaf, line 340:
```suggestion
        auto result = torch::make_intrusive<TensorMapHolder>(std::move(tensor));
        return result->to(device);
```

metatensor-torch/src/tensor.cpp
```diff
@@ -334,14 +346,15 @@ TorchTensorMap TensorMapHolder::keys_to_properties(torch::IValue keys_to_move, b
 }
 
 TorchTensorMap TensorMapHolder::keys_to_samples(torch::IValue keys_to_move, bool sort_samples) const {
+    auto device = this->keys()->values().device();
     if (keys_to_move.isString() || keys_to_move.isList() || keys_to_move.isTuple()) {
         auto selection = extract_list_str(keys_to_move, "TensorMap::keys_to_samples first argument");
         auto tensor = tensor_.keys_to_samples(selection, sort_samples);
-        return torch::make_intrusive<TensorMapHolder>(std::move(tensor));
+        return torch::make_intrusive<TensorMapHolder>(std::move(tensor))->to(device);
```
Luthaf, line 353:
```suggestion
        auto result = torch::make_intrusive<TensorMapHolder>(std::move(tensor));
        return result->to(device);
```

metatensor-torch/src/tensor.cpp
```diff
@@ -334,14 +346,15 @@ TorchTensorMap TensorMapHolder::keys_to_properties(torch::IValue keys_to_move, b
...
     } else if (keys_to_move.isCustomClass()) {
         auto selection = keys_to_move.toCustomClass<LabelsHolder>();
         auto tensor = tensor_.keys_to_samples(selection->as_metatensor(), sort_samples);
-        return torch::make_intrusive<TensorMapHolder>(std::move(tensor));
+        return torch::make_intrusive<TensorMapHolder>(std::move(tensor))->to(device);
```
Luthaf, line 357:
```suggestion
        auto result = torch::make_intrusive<TensorMapHolder>(std::move(tensor));
        return result->to(device);
```

metatensor-torch/src/tensor.cpp
```diff
@@ -350,9 +363,10 @@ TorchTensorMap TensorMapHolder::keys_to_samples(torch::IValue keys_to_move, bool
 }
 
 TorchTensorMap TensorMapHolder::components_to_properties(torch::IValue dimensions) const {
+    auto device = this->keys()->values().device();
     auto selection = extract_list_str(dimensions, "TensorMap::components_to_properties argument");
     auto tensor = this->tensor_.components_to_properties(selection);
-    return torch::make_intrusive<TensorMapHolder>(std::move(tensor));
+    return torch::make_intrusive<TensorMapHolder>(std::move(tensor))->to(device);
```
Luthaf, line 369:
```suggestion
    auto result = torch::make_intrusive<TensorMapHolder>(std::move(tensor));
    return result->to(device);
```

Luthaf (commented):
Yes, this is exactly what I had in mind! 

One point I'm unsure about is if these should be exposed to the end users. On one hand it could be interesting, and play part of the role of `metatensor.to` operation; on the other hand we need the same API in metatensor-core.

For metatensor-core, calling `TensorMap.to("cuda")` would do nothing (maybe a warning?) if the arrays are numpy; and should move the values to CUDA is the arrays are torch. But they currently can not move the samples values. We could accept this limitation and document it; or not expose these new functions to Python.

Another solution (that would be a separate PR) could also be to do the same thing in -core as we do in -torch, and duplicate the labels values with the same arrays backend as the values, which in the torch case would allow to move them to GPU.

What do you think?

frostedoyster:
To me it feels like we have a lot of flexibility with the current `to()` operation in Python, because that can switch between backends, dtypes and devices. The limitations regarding the labels is already there, so I don't see an issue with keeping it or documenting it better.
I really do believe that we should have block.to() and tensor.to(), just like we have labels.to(), but it's a lot of work. And also a lot of demolition of all the functionality we worked very hard on in the `metatensor.to()` operation, which is however just a big hack in practice. So I don't know really

Luthaf:
Ok, let's open an issue for block.to() and tensor.to(), and leave it for later

Luthaf (approved):
This it good to go for me!

Issue #370, opened 2023-09-13
Enforce the same `device/dtype` when constructing `metatensor.torch.TensorBlock/TensorMap`s


frostedoyster:
Not doing so would most likely result in operations failing due to device or type mismatches

Luthaf:
I think the checks should be done in the metatensor-torch C++ code, since the same issue can happen when creating data from C++.

Luthaf:
> I agree, devices should be consistent at creation time. I can't see use cases to have different pieces on different devices.

@ceriottm in https://github.com/lab-cosmo/metatensor/issues/413#issuecomment-1789830279

> Perhaps to distribute calculations. Some blocks could live in cuda:0, others in cuda:1, and so on

> @frostedoyster in https://github.com/lab-cosmo/metatensor/issues/413#issuecomment-1790135849


---

I could see a point for different blocks to be on different variations on the same kind of device (i.e. `cuda:0`/`cuda:1`), but not on different kind of devices. The keys would then have to be on the same kind of device as well.

ceriottm:
Should we enforce consistency between data and metadata  of a block, and let complete freedom between different blocks in the same map?

Luthaf:
Thinking a bit more about this, I don't think we can have different devices for different `block.values`, because `keys_to_xxx` requires all the data to be on the same device.

For distributed calculations, I feel like it would be better to distribute over samples instead of blocks, and then use separate TensorMap on separate devices. Can you see a point where splitting distributed calculations over samples would not work?

Luthaf:
Fixed by #414

PR #414, opened 2023-11-03
Check `metatensor.torch.TensorBlock/Map` at construction

frostedoyster:
Implements #370

----
:books: Documentation preview :books:: https://metatensor--414.org.readthedocs.build/en/414/

### DIFF
--- a/metatensor-torch/src/tensor.cpp
+++ b/metatensor-torch/src/tensor.cpp
@@ -59,7 +59,7 @@ TensorMapHolder::TensorMapHolder(TorchLabels keys, const std::vector<TorchTensor
     }
 
     auto device = keys->values().device();
-    auto dtype = blocks[0]->values().dtype();
+    auto scalar_type = blocks[0]->values().scalar_type();
 
     for (const auto& block : blocks) {
         if (block->values().device() != device) {
@@ -68,10 +68,11 @@ TensorMapHolder::TensorMapHolder(TorchLabels keys, const std::vector<TorchTensor
                 "got " + block->values().device().str() + " and " + device.str()
             );
         }
-        if (block->values().dtype() != dtype) {
+        if (block->values().scalar_type() != scalar_type) {
             C10_THROW_ERROR(TypeError,
                 "cannot create TensorMap: all blocks must have the same dtype, "
-                "got " + std::string(block->values().dtype().name()) + " and " + std::string(dtype.name())
+                "got " + std::string(c10::toString(block->values().scalar_type())) +
+                " and " + std::string(c10::toString(scalar_type))
             );
         }
     }
--- a/metatensor-torch/src/tensor.cpp
+++ b/metatensor-torch/src/tensor.cpp
@@ -4,6 +4,7 @@
 #include "metatensor/torch/tensor.hpp"
 #include "metatensor/torch/array.hpp"
 #include "metatensor/torch/block.hpp"
+#include "metatensor/torch/misc.hpp"
 
 using namespace metatensor_torch;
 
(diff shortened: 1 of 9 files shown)


frostedoyster:
I don't fully understand the structure of `TorchLabels`, so I haven't been able to extract their device

metatensor-torch/src/tensor.cpp
```diff
@@ -15,7 +15,46 @@ bool custom_class_is(torch::IValue ivalue) {
...
+
+    auto block_0 = this->tensor_.block_by_id(0);
+    const auto array_0 = block_0.mts_array();
+    const auto* ptr_0 = static_cast<metatensor::DataArrayBase*>(array_0.ptr);
+    const auto* torch_tensor_wrapper_0 = dynamic_cast<const TorchDataArray*>(ptr_0);
+    if (torch_tensor_wrapper_0 == nullptr) {
+        C10_THROW_ERROR(ValueError,
+            "this TensorBlock does not contain a C++ torch Tensor"
+        );
+    }
```
Luthaf, line 30: I don't love that we need to access these internal details here. It might be worth to make this constructor private or even remove it entirely to prevent misusing it. I'll have a look at how much pain removing this would create.

metatensor-torch/src/tensor.cpp
```diff
@@ -52,7 +91,29 @@ static std::vector<metatensor::TensorBlock> blocks_from_torch(const std::vector<
...
+    for (const auto& block : blocks) {
+        if (block->values().device() != device) {
+            C10_THROW_ERROR(TypeError,
+                "cannot create TensorMap: keys and blocks must be on the same device"
+            );
+        }
+        if (block->values().dtype() != dtype) {
+            C10_THROW_ERROR(TypeError,
+                "cannot create TensorMap: all blocks must have the same dtype"
+            );
```
Luthaf, line 113: same comments as above: this should be a TypeError, and the error message should contain some information about the encoutered device/dtype.

Luthaf (approved):
This looks good for me, let me know if it works for you as well and I can merge it!

frostedoyster:
What happens here?
https://github.com/lab-cosmo/metatensor/blob/59cb41e906475967884becb0d8693f94a204b832/metatensor-torch/src/tensor.cpp#L72-L74
Are the keys always returned on the CPU? If that's the case, isn't it counterintuitive to ask the user to supply them on the device of the blocks but always return them on the CPU?

Luthaf:
Good point! I'll check how to fix this.

Luthaf:
This should be fine, the GPU tensor is stored inside the metatensor-core Labels. Does this fail? We should add a test for it!

frostedoyster:
Are you sure?
https://github.com/lab-cosmo/metatensor/blob/59cb41e906475967884becb0d8693f94a204b832/metatensor-torch/src/labels.cpp#L126-L130C7
It seems like the `metatensor::Labels` are always on CPU. The GPU values look like they're stored inside a different field in the torch Labels class. Then when we use the `TensorMapHolder(metatensor::TensorMap)` constructor, the new TensorMap won't have any keys on the GPU, right? I'm a bit confused

Luthaf:
The code you linked to creates a metatensor-core Labels, which is always on CPU. The lines below register the original tensor (not the CPU one) as custom user data; and that's the one we fetch when accessing `Labels.values` in metatensor-torch

https://github.com/lab-cosmo/metatensor/blob/59cb41e906475967884becb0d8693f94a204b832/metatensor-torch/src/labels.cpp#L133-L136

When moving a Labels to GPU, we actually end up with two copies of the data: one on CPU to be used by metatensor-core in the `x_to_y` functions, and one on GPU that the users can access and use in their code.

frostedoyster:
Thank you very much, I understand now. I'll try to write some tests quickly

python/metatensor-torch/tests/block.py
```diff
@@ -234,3 +234,32 @@ def forward(self, x: TensorBlockWrap) -> TensorBlockWrap:
 
     module = TestModule()
     module = torch.jit.script(module)
+
+
+def test_different_device():
+    with pytest.raises(ValueError):
```
Luthaf, line 240: please check the error message here

metatensor-torch/tests/block.cpp
```diff
@@ -77,4 +77,15 @@ TEST_CASE("Blocks") {
...
+
+    SECTION("different devices") {
+        REQUIRE_THROWS(
+            TensorBlockHolder(
+                torch::full({3, 2}, 11.0),
+                LabelsHolder::create({"s"}, {{0}, {2}, {1}})->to(torch::kMeta),
+                std::vector<TorchLabels>{},
+                LabelsHolder::create({"p"}, {{0}, {1}})
+            )
+        );
```
Luthaf, line 89: Please use `CHECK_THROWS_WITH` here. `CHECK_` instead of `REQUIRES_` just means that the test will continue running if this specific check fails. The `_WITH` part checks the error message associated with the exception. (You might need to use `StartsWith(....)` here, see others tests for usage example)

```suggestion
        CHECK_THROWS_WITH(
            TensorBlockHolder(
                torch::full({3, 2}, 11.0),
                LabelsHolder::create({"s"}, {{0}, {2}, {1}})->to(torch::kMeta),
                std::vector<TorchLabels>{},
                LabelsHolder::create({"p"}, {{0}, {1}})
            ),
            "expected error message..."
        );
```

PR #416, opened 2023-11-09
Get more information when n-commits-since-last-tag.py fails

Luthaf:
Some people where having issues installing the code for the tutorial during the MLIP-23 workshop, this should help debug issues.

# Reviewer checklist

----
:books: Documentation preview :books:: https://metatensor--416.org.readthedocs.build/en/416/

### DIFF
--- a/scripts/n-commits-since-last-tag.py
+++ b/scripts/n-commits-since-last-tag.py
@@ -4,6 +4,7 @@ This script calls git to get the number of commits since the last tag matching a
 pattern (pattern given on the command line).
 """
 import os
+import re
 import subprocess
 import sys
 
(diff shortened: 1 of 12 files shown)


scripts/n-commits-since-last-tag.py
```diff
@@ -17,26 +18,42 @@ def warn_and_exit(message, exit_code=0):
...
     if len(sys.argv) != 2:
         warn_and_exit(f"usage: {sys.argv[0]} <tag-prefix>", exit_code=1)
 
-    result = subprocess.run(
-        ["git", "rev-parse", "--show-toplevel"],
-        capture_output=True,
-        check=False,
-        encoding="utf8",
-    )
+    result = run_subprocess(["git", "--version"], check=False)
```
PicoCentauri, line 46: `check=True` does not work? Then the error should be displayed as well right?

scripts/n-commits-since-last-tag.py
```diff
@@ -17,26 +18,42 @@ def warn_and_exit(message, exit_code=0):
...
+
+    if result.returncode != 0:
+        warn_and_exit("could not run `git`, is it installed on your system?")
+
+    # We need git >=2.0 for `git tag --sort=-creatordate` below
+    _, _, git_version, *_ = result.stdout.split()
```
PicoCentauri, line 52: Wow, fancy dropping of return values. 😅

PicoCentauri (commented):
Seems like some good debugging messages. I have question to one subprocess call.

scripts/n-commits-since-last-tag.py
Luthaf: We don't want `check=True` here, because we don't want to get an exception if the user does not have git installed. Instead we manually check the return code below, send a warning and exit.

scripts/n-commits-since-last-tag.py
Luthaf: Yeah, the issue is that on most Linux, git version is `git version 2.40`, but on macOS we get `git version 2.39.3 (Apple Git-145)`. This tries to handle both!

scripts/n-commits-since-last-tag.py
PicoCentauri: I just tried it and it is still throwing an error, i. when I do 

` run_subprocess(["gits", "--version"], check=False)`

I get 

```python
Cell In[11], line 1
----> 1 run_subprocess(["gits", "--version"], check=False)

Cell In[2], line 2, in run_subprocess(args, check)
      1 def run_subprocess(args, check=True):
----> 2     output = subprocess.run(
      3         args,
      4         capture_output=True,
      5         encoding="utf8",
      6         check=False,
      7     )
      9     if check and output.returncode != 0:
     10         raise Exception(
     11             f"failed to run '{' '.join(args)}' (exit code {output.returncode})\n"
     12             f"stdout: {output.stdout}\n"
     13             f"stderr: {output.stderr}\n"
     14         )
     
     ...
     
     FileNotFoundError: [Errno 2] No such file or directory: 'gits'
```

But maybe I am doing something wrong.

scripts/n-commits-since-last-tag.py
PicoCentauri: Yes, I know these struggles. I played around with these things and then I decided to use versioneer for everything.

I hope this is not something which will be broken within every release of a new macOS version. 😅

scripts/n-commits-since-last-tag.py
Luthaf: I did not check for "the user does not have git", I'll see what I can do!

Luthaf: Should be fixed now!

PR #423, opened 2023-11-27
Use default tox error for non existing environments

PicoCentauri:
Non existing tox environments will raise an error by default starting from [tox `4.9.0`](https://tox.wiki/en/latest/changelog.html#v4-9-0-2023-08-16).

```bash
philiploche@Philips-MacBook-Air:~/repos/lab-cosmo/metatensor$ tox -e fooo
ROOT: HandledError| provided environments not found in configuration file:
fooo
```

I think we can bump our min version and remove our custom code.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-min_version = 4.0
+min_version = 4.9.0
 # these are the default environments, i.e. the list of tests running when you
 # execute `tox` in the command-line without anything else
 envlist =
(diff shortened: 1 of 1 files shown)


PicoCentauri:
However, `tox > 4.8` requires `Python >= 3.8`. We still want to support Python 3.7.

Luthaf:
Looks like Python 3.7 is already EOL =/. What was the reason for still supporting 3.7?

PicoCentauri:
I think we do not have a specific reason. Our philosophy is to support as many versions as possible to avoid probable issues especially on HPC systems that do not have recent Python versions. Even though I never fully accepted this as a reason ;-).

Luthaf:
Ok, so most HPC centers have a module for a more recent Python, and Python3 in ubuntu-20.04 is 3.8; so I think we can drop Python 3.7 now.

Luthaf:
Longer term, I would try to follow Python EOL, with exceptions if some common Linux distro are still using an old one by default.

PR #427, opened 2023-12-02
Implementation of ModuleMap in metatensor-learn

agoscinski:
Moves the ModuleTensorMap from equisolve https://github.com/lab-cosmo/equisolve/pull/62 here with a renaming to ModuleMap and some additional doc tests.

### Questions 
- renaming of `Linear` to `LinearMap`?
- I don't like that I copy the `utils.py` in `python/metatensor-learn/tests/utils.py`and `python/metatensor-torch/tests/learn/utils.py` is there an easy solution for this?
- I wanted to add some example that explains the usage of the class for different use cases in atomistic learning, since it has some parameters and constructors that are not trivially translatable to a use case. I am not exactly sure in which docs folder I would put this. Is this `docs` also the place for subpackages or would I create a `python/metatensor-learn/docs` folder? It would be a python file with documentation

### Reviewer checklist

----
:books: Documentation preview :books:: https://metatensor--427.org.readthedocs.build/en/427/


agoscinski:
> docs is fine. Would this be an example based on sphinx-gallery?

Yes, because the ModuleMap's application to different scenario's are not apparent from reading the API, I would say.
#### Use case: Equivariant learning (lambda's are in the keys)
different type of module per key (lam=0 nonlinear, rest linear)
#### Use case: sparse features (neighbor species are in the tensor map keys).
same type of module but different module per key (weights are separated) 
#### Use case: Per central-species model vs combined central-species model (central-species are in the keys).
separate (per central-species) or same module (combined central-species) for each central species

I thought about using random tensor's as input. I am not sure if `metatensor-learn`is the right spot, since this is not agnostic to atomistic learning.

Luthaf:
>  I am not sure if metatensor-learn is the right spot, since this is not agnostic to atomistic learning.

I would put this in the `atomistic models` section of the documentation then

agoscinski:
> I would put this in the atomistic models section of the documentation then

Okay, should we add for now just a `.rst` file with all the python code as code block? For executed code we would need a bit more infrastructure. I am not even sure if we want to have examples in this repo, but it makes sense for this project.

For atom-agnostic documentation we can use `metatensor/docs/learn`?

agoscinski:
Ah okay found now the PR Philip was mentioning. Would then just wait till #431 is merged

Luthaf:
> For atom-agnostic documentation we can use metatensor/docs/learn?

For which kind of documentation? For API reference, `reference/learn` is fine; for tutorials the files can live in `python/examples/learn` but I'm not sure where to put them in the rendered documentation. We can pull them in `atomistic/tutorials` for atomistic stuff, and maybe a new `tutorials` section, with a `metatensor-learn` subsection for the rest?

agoscinski:
Pinning a TODO from conversation: Use of `Labels.position` function to circumvent the hashing problem of LabelsEntry

Luthaf:
Now that #426 has been merged, can you rebase this PR on top of master?

agoscinski:
> Pinning a TODO from conversation: Use of Labels.position function to circumvent the hashing problem of LabelsEntry

Embedded this TODO. Using now Labels.position function to determine the index to access the module from a ModuleList. It required a change in the `__init__` constructor which I think make it more consistent with the `from_module` constructor.

I do not copy the modules on construction of ModuleMap, that is more consistent with torch.nn.ModuleList/Dict etc.. Only exception is ModuleMap.from_module with `many_to_one=False` because you need to create separate module that requires copying. I am thinking to change the default behavior to copy in every constructor to be consistent within the class rather than the halfbaked consistency with torch.nn.ModuleList/Dict. I don't see a problem in usage when everything is copied on construction.

The member variable `self._modules` seems to cause problems with `torch.nn.Module` inheritance, therefore it is called `self._modules`.

agoscinski:
I am not super happy with the usage `TORCH_KWARGS`. It would be better in my mind if we use an environment variable that is retrieved in the python code, then it can be also be set in the tox, but I would leave this for some other PR.
Also that we have multiple utility that create TensorMaps in different ways, but making a generic one needs more consideration and would be also something more for `metatensor-operations` , so this I also leave for another PR.
An tutorial example of the usage of ModuleMap I would also leave for another time

agoscinski:
Not sure what is wrong with codecov, I will try it just tomorrow again to see if its consistent
```
[2024-01-15T09:31:04.537Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 502 - 
<html><head>
```

python/metatensor-learn/metatensor/learn/nn/module_map.py
```diff
@@ -0,0 +1,470 @@
...
+        It got completely lost because we cannot know in general what the output is.
+        You can add in the initialization of the ModuleMap a TensorMap that contains
+        the intended output Labels.
+    """
+
+    def __init__(
+        self,
+        in_keys: Labels,
+        modules: Sequence[Module],
+        out_tensor: Optional[TensorMap] = None,
```
abmazitov, line 151: It seems that the only reason to have the `out_tensor` in the ModuleMap is to provide a name for properties dimension of the output tensor. In my experience, this feature is really redundant, because a simple `feature` or `_` or any other placeholder works perfectly, since you never explicitly address this name anyway. 

Moreover, usually the presence of the output tensor or array means that the values will be stored in this specific tensor, and this is not the case, since we actually create the new `TensorMap`.

python/metatensor-learn/metatensor/learn/nn/module_map.py
```diff
@@ -0,0 +1,470 @@
...
+    ):
+        super().__init__()
+        if len(in_keys) != len(modules):
+            raise ValueError(
+                "in_keys and modules must match in length, but found "
+                f"{len(in_keys) != len(modules)} [len(in_keys) != len(modules)]"
+            )
+        self._in_keys: Labels = in_keys
+        self._modules_list = ModuleList(modules)
```
abmazitov, line 160: This is a rookie question, but why can't we just used the `ModuleDict` with this fancy interface to store the modules?

python/metatensor-learn/metatensor/learn/nn/module_map.py
```diff
@@ -0,0 +1,470 @@
...
+            out_blocks.append(out_block)
+
+        return TensorMap(tensor.keys, out_blocks)
+
+    def forward_block(self, key: LabelsEntry, block: TensorBlock) -> TensorBlock:
+        position = self._in_keys.position(key)
+        if position is None:
+            raise KeyError(f"key {key} not found in modules.")
+        module_idx: int = position
+        module: ModuleMapInterface = self._modules_list[module_idx]
```
abmazitov, line 292: Since we have a `get_module` method, maybe we can use it here?

python/metatensor-learn/metatensor/learn/nn/module_map.py
```diff
@@ -0,0 +1,470 @@
...
+            properties = self._out_tensor.block(key).properties
+        return TensorBlock(
+            values=out_values,
+            properties=properties,
+            components=block.components,
+            samples=block.samples,
+        )
+
+    @property
+    def modules(self):
```
abmazitov, line 306: AFAICS, this property method is never used in a code. Should we still keep it here?

It seems, that there may be two purposes for such a method: 
1. If we want to create a shortcut for `self._modules_list`
2. If we want to display the modules used inside a `ModuleMap`

So the first feature is not used anyway. As for the second - we can just see the modules with the basic printouts (`__str__()` and `__repr__()`)

python/metatensor-learn/metatensor/learn/nn/module_map.py
```diff
@@ -0,0 +1,470 @@
...
+    @property
+    def out_tensor(self) -> Optional[TensorMap]:
+        """
+        The tensor map that is used to determine properties labels of the output of
+        forward function.
+        """
+        return self._out_tensor
+
+
+class Linear(ModuleMap):
```
abmazitov, line 346: Okay, here I'm not sure about the general idea of how it works. It seems like if we want to use a `Linear` module, we should not only create an input `TensorMap`, but also an output `TensorMap` to provide desired dimensions size. IMO, creating tensormaps is really something painful, particularly if we want to stack a few `Linear` modules together.

So it's probably better to have the closest possible to the `torch.nn.Linear` implementation with explicit dimensions size, because this is something people would expect

python/metatensor-learn/metatensor/learn/nn/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .module_map import ModuleMap, Linear  # noqa
```
abmazitov, line 1: Generally, I would recommend to store different module implementations (as `Linear`) in separate files

python/metatensor-learn/metatensor/learn/nn/module_map.py
Luthaf: How about `out_features=None`, and have `None` map to `Labels.range("_", len(output))`?

Luthaf: I think giving the option for the end user to specify properties labels if they want to is good (defaulting to `Labels.range("_", len(features))`).

But agreed on the implication that data would be stored there. How about `out_metadata`? or `out_properties: Tuple[Labels, List[Labels]]` (as a poor man's `Dict[LabelEntry, Labels]`)?

python/metatensor-learn/metatensor/learn/nn/module_map.py
Luthaf: What would be the key type? a string is not ideal since it requires creating a new string for each lookup. Here the list index can be found with a proper hash lookup (though `Labels.position(...)`), making this behave a lot like an actual dict for runtime-performance

Luthaf:
> Not sure what is wrong with codecov, I will try it just tomorrow again to see if its consistent

codecov upload is failing very regularly, you can restart the job & it should pass. Cf https://github.com/codecov/codecov-action/issues/926 as well.

python/metatensor-learn/metatensor/learn/nn/module_map.py
agoscinski: > This is a rookie question, but why can't we just used the ModuleDict with this fancy ModuleMapInterface to store the modules?

This is a very valid question. We did this in an earlier implementation (that can be found [here](https://github.com/lab-cosmo/equisolve/blob/c858bedef4b2799eb445e4c92535ee387224089a/src/equisolve/nn/module_tensor.py#L123C61-L123C71)). As Guilaume explained, since we cannot use `LabelEntriy`s as keys because they are not hashable (and we cannot make them hashable without dropping support for torch versions before 2.0 if I remember correctly), we use this solution of two lists in the same order.

python/metatensor-learn/metatensor/learn/nn/module_map.py
agoscinski: It is quite close to torch. Torch requires also to give input and output dimension, Because we can have different input and output dimensions for each tensor block, we need at least these information for each block. This I can adapt as it was suggested for the output labels https://github.com/lab-cosmo/metatensor/pull/427/files#r1452473183 by merging the two constructors (`__init__` and `from_module`)

```python
class Linear
    def __init__(
        self,
        in_properties:  Union[int, Dict[LabelsEntry, Labels]],
        out_properties: Union[int, Dict[LabelsEntry, Labels]],
        bias: Union[bool, Dict[LabelsEntry, bool]] = True,
        device: torch.device = None,
        dtype: torch.dtype = None,
        many_to_one: bool = True,
    ):

```

What do you think?

python/metatensor-learn/metatensor/learn/nn/module_map.py
Luthaf: > and we cannot make them hashable without dropping support for torch versions before 2.0 if I remember correctly

No, we cannot make them hashable at all. `__hash__` is not in the list of allowed magic methods: https://github.com/pytorch/pytorch/blob/26956980c6953b42e57ad889b87444c4abccec1c/torch/csrc/jit/python/script_init.cpp#L652 

We could add it there, but we won't be able to use it until it reaches a stable release.

python/metatensor-learn/metatensor/learn/nn/module_map.py
abmazitov: > I think giving the option for the end user to specify properties labels if they want to is good (defaulting to `Labels.range("_", len(features))`).
> 
> But agreed on the implication that data would be stored there. How about `out_metadata`? or `out_properties: Tuple[Labels, List[Labels]]` (as a poor man's `Dict[LabelEntry, Labels]`)?

I think that the general issue here is the following
 - If we want to set the properties dimension name for hidden layers, we can't really say much about the nature of this data, except that this is a hidden representation. Thus the hardcoded placeholder name is a good choice IMO
 - If we want to set the name for a last layer only, then it's not really necessary, since we know what type of property we expect to get
 
So the only really useful case in this situation will be a multi-modality model (for instance which predicts both absolute energy and the energy of formation), and we store the output in format of a single tensormap with two different output blocks corresponding to each output type each.

python/metatensor-learn/metatensor/learn/nn/module_map.py
abmazitov: So the only purpose of this method is to provide the user with a safe way of having an access to the modules list?

python/metatensor-learn/metatensor/learn/nn/module_map.py
abmazitov: I agree that in general the different size of the blocks may be an issue, but I feel like in my applications it have never happened. 

Let say, we never want to use the Linear layer on top of the raw spherical expansion coefficients, because we then mix channels and break everything. But usually in both radial and power spectrum, the number of features in each block is fixed, and thus providing only input and output sizes once was working perfectly.

python/metatensor-learn/metatensor/learn/nn/module_map.py
agoscinski: > If we want to set the name for a last layer only, then it's not really necessary, since we know what type of property we expect to get

Since the ModuleMap can allow arbitrary modules, there will be modules that perform operations with tractable property names. You can say the module developer needs to do this in their module implementation, but for that you need to override the forward function around the ModuleMap. Even you create a new Tensor.Map,  since you cannot change the Labels after creation, or you copy the whole forward function from ModulMap  to add the change of the label. I feel like this is more convenient to do. And for the some output module you even said that it is useful to have. I am not sure if in practice people use it, but in principle I can see some useful applications for it.

python/metatensor-learn/metatensor/learn/nn/module_map.py
agoscinski: In the end I might not change it, because I might need to reuse the `position` to also retrieve the correct Labels of out_properties, lets see

python/metatensor-learn/metatensor/learn/nn/module_map.py
agoscinski: > I agree that in general the different size of the blocks may be an issue, but I feel like in my applications it have never happened.

In equivariant learning it happens. But both will ways will be supported. If we have an example for the easy case (as first one) and one for the complex case then it should be fine for the user to understand quickly the more complicated function signature.

python/metatensor-torch/tests/learn/module_map.py
```diff
@@ -0,0 +1,177 @@
...
+        torch.set_default_device(TORCH_KWARGS["device"])
+        torch.set_default_dtype(TORCH_KWARGS["dtype"])
+
+    @pytest.mark.parametrize(
+        "tensor",
+        [
+            random_single_block_no_components_tensor_map(True, True),
+        ],
+    )
+    def test_module_tensor(self, tensor):
```
Luthaf, line 39: the parametrize does not seem to be doing anything here. Is this intended to parametrize the `use_metatensor_torch`?

```suggestion
    @pytest.mark.parametrize("use_metatensor_torch", (True, False))
    def test_module_tensor(self, use_metatensor_torch):
    use_torch=True
    tensor = random_single_block_no_components_tensor_map(use_torch, use_metatensor_torch)
```

Otherwise I would just remove the whole parametrize call (here and in the other tests)

python/metatensor-learn/metatensor/learn/nn/module_map.py
```diff
@@ -0,0 +1,463 @@
...
+        position = self._in_keys.position(key)
+        if position is None:
+            raise KeyError(f"key {key} not found in modules.")
+        module_idx: int = position
+        module: ModuleMapInterface = self[module_idx]
+        return module
+
+    @property
+    def in_keys(self) -> Labels:
+        """ """
```
Luthaf, line 316: this needs some documentation

python/metatensor-learn/tests/module_map.py
```diff
@@ -0,0 +1,143 @@
...
+                ref_values = module(block.values)
+            out_block = out_tensor.block(key)
+            assert torch.allclose(ref_values, out_block.values)
+
+            for parameter, gradient in block.gradients():
+                with torch.no_grad():
+                    ref_gradient_values = module(gradient.values)
+                assert torch.allclose(
+                    ref_gradient_values, out_block.gradient(parameter).values
+                )
```
Luthaf, line 143: could you add a test checking that both ModuleMap and Linear can be compiled with `torch.jit.script`?

python/metatensor-learn/metatensor/learn/nn/module_map.py
```diff
@@ -0,0 +1,463 @@
...
+        It got completely lost because we cannot know in general what the output is.
+        You can add in the initialization of the ModuleMap a TensorMap that contains
+        the intended output Labels.
+    """
+
+    def __init__(
+        self,
+        in_keys: Labels,
+        modules: Sequence[Module],
+        out_properties: Dict[LabelsEntry, Labels] = None
```
Luthaf, line 151: This might fail with torchscript. I was more thinking of a custom `LabelsEntryDict` class that would store a list of values and Labels. It will be a bit of work, so we could do this in a later PR. For now, maybe `out_properties=Optional[List[Labels]]`, with the same number of elements as keys`?

python/metatensor-learn/metatensor/learn/nn/module_map.py
```diff
@@ -0,0 +1,463 @@
...
+        return TensorMap(tensor.keys, out_blocks)
+
+    def forward_block(self, key: LabelsEntry, block: TensorBlock) -> TensorBlock:
+        module = self.get_module(key)
+
+        out_values = module.forward(block.values)
+        if self._out_properties is None:
+            properties = Labels.range("_", out_values.shape[-1])
+        else:
+            properties = self._out_properties[key]
```
Luthaf, line 289: This will not work in general: if the `key` comes from a different Labels than the one used to initialize `self._out_properties`, this will raise KeyNotFound.

python/metatensor-learn/tests/module_map.py
Luthaf: And another test with `out_properties=...`

python/metatensor-torch/tests/learn/utils.py
```diff
@@ -0,0 +1,88 @@
...
+
+    cell_gradient = TensorBlock(
+        values=create_random_array(4, 6, 2),
+        samples=Labels(
+            ["sample", "structure"],
+            create_int32_array([[0, 0], [1, 1], [2, 2], [3, 3]]),
+        ),
+        components=[
+            Labels(
+                ["direction_xx_yy_zz_yz_xz_xy"],
```
Luthaf, line 80:
```suggestion
                ["voigt_index"],
```

(I don't really mind)

python/metatensor-learn/metatensor/learn/nn/module_map.py
abmazitov: Okay, then maybe it's better to just provide a list or tuple of input and output dimensions? 
My point is, creating a tensormap by hands is usually very inconvenient, and it requires the list of dimensions anyway. What do you think of

```py
class Linear(ModuleMap):
    def __init__(self, in_features: Union[List, Tuple], out_features: Union[List, Tuple], bias: Bool = True, device = None, dtype = None):
        pass
```

And then just initialising the set of Linear layers for each block separately?

Luthaf: I'm not sure what you are suggesting here. I agree that building TensorMap by hand is a bit clunky, but it is not required by the latest API suggestion? For `in_properties:  Union[int, Dict[LabelsEntry, Labels]]`, you can use it as

```py
linear = Linear(in_properties=3, out_properties=6)
```

or if you care

```py
linear = Linear(in_properties=3, out_properties=Labels("nn_features", [1, 2, 5, 4]))
```

or if you want different things for each block

```py
linear = Linear(in_properties=[3, 4, 5], out_properties=[Labels(...), Labels(...), Labels(...)])
```

I don't understand what would be inside the `List` or `Tuple` in your suggestion

python/metatensor-learn/metatensor/learn/nn/module_map.py
agoscinski: Yes that was a question here https://github.com/lab-cosmo/metatensor/pull/427#discussion_r1457592018 Then I do it this way

python/metatensor-torch/tests/learn/module_map.py
```diff
@@ -0,0 +1,177 @@
...
+        super().__init__()
+        self._linear = torch.nn.Linear(in_features, out_features)
+        self._activation = Sigmoid()
+        self._last_layer = torch.nn.Linear(out_features, 1)
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        return self._last_layer(self._activation(self._linear(input)))
+
+
+class TestModuleMap:
```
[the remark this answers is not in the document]
agoscinski: In the metatensor-learn tests (for the metatensor-core) tests we use a skipif for the tests
```python
@pytest.mark.skipif(not (HAS_TORCH), reason="requires torch to be run")
class TestModuleMap:
```
One could also add it to ever test, but this requires less code. And I want to mimic the test files.

python/metatensor-torch/tests/learn/module_map.py
agoscinski: Yes there was the ambition in adding more tensor maps with multiple blocks, but I think this should be actually a separate test. I'll remove the parameterize

python/metatensor-learn/metatensor/learn/nn/module_map.py
agoscinski: thanks, got lost in revision, added

python/metatensor-learn/metatensor/learn/nn/module_map.py
agoscinski: Yes that was still something I was waiting to fix until comments from Arslan. Should work now!

python/metatensor-torch/tests/learn/module_map.py
agoscinski: I use a fixture from utils (with an explicit import!) and my plan is to move `random_single_block_no_components_tensor_map` in another PR to `metatensor.operations.test_utils.random_tensor` as discussed in https://github.com/lab-cosmo/metatensor/issues/451

python/metatensor-learn/metatensor/learn/nn/__init__.py
agoscinski: separated the files

python/metatensor-learn/tests/module_map.py
agoscinski: Added tests for `out_properties` and also for `from_weight` in Linear

agoscinski:
Follow up Issues/PRs:
- merge the construction of test tensor maps https://github.com/lab-cosmo/metatensor/issues/451 to then add tests with multiple blocks
- Figure out to add forward function into doc that is TorchScript compatible: TorchScript allows not # added to appear in doc, :inherited-members: is not compatible with torch
But also does not allow to use super in forward https://github.com/pytorch/pytorch/issues/42885
- add weight and bias  property to Linear (will be probably List[nn.Parameter])
- Consider to also accept single Labels for `out_properties` for `ModuleMap.from_module` and `Linear` default constructor. It is also a bit tricky to the get the list check done in a way that does not break in metatensor-torch. I was not sure if this is useful in the first place, so I skipped this.
- add examples for usage of ModuleMap as mentioned in https://github.com/lab-cosmo/metatensor/pull/427#issuecomment-1849001299

agoscinski:
Test failure seems to be due to https://github.com/lab-cosmo/metatensor/issues/460

agoscinski:
Would just use CHANGELOG for the message in the squashed commit
```
adds ModuleMap and Linear modules to metatensor.learn.nn

These modules follow closely torch.nn.ModuleDict and torch.nn.Linear
in PyTorch but they are adapted to work with TensorMap's
```

Luthaf:
> Follow up Issues/PRs:

Could you open any required issue on this repo so it does not stay a comment on this PR?

python/metatensor-torch/tests/learn/module_map.py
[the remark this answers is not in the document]
Luthaf: Since the whole ModuleMap/Linear requires torch anyway, I would just do a `importorskip` on the whole file.

Luthaf (approved):

abmazitov (approved):

PR #433, opened 2023-12-14
Allow `block_from_array` to run on any device

frostedoyster:
At the moment, `block_from_array` only works on CPU

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--433.org.readthedocs.build/en/433/

### DIFF
--- a/metatensor-torch/include/metatensor/torch/labels.hpp
+++ b/metatensor-torch/include/metatensor/torch/labels.hpp
@@ -1,7 +1,6 @@
 #ifndef METATENSOR_TORCH_LABELS_HPP
 #define METATENSOR_TORCH_LABELS_HPP
 
-#include <c10/core/Device.h>
 #include <string>
 #include <vector>
 
--- a/metatensor-torch/include/metatensor/torch/labels.hpp
+++ b/metatensor-torch/include/metatensor/torch/labels.hpp
@@ -63,8 +63,8 @@ public:
     static TorchLabels empty(torch::IValue names);
 
     /// Create Labels with a single dimension with the given name and values in
-    /// the [0, stop) range, on the given device
-    static TorchLabels range(std::string name, int64_t end, torch::IValue device = torch::IValue(torch::kCPU));
+    /// the [0, stop) range
+    static TorchLabels range(std::string name, int64_t end);
 
     /// Create a `LabelsHolder` from a pre-existing `metatensor::Labels`
     explicit LabelsHolder(metatensor::Labels labels);
--- a/metatensor-torch/include/metatensor/torch/labels.hpp
+++ b/metatensor-torch/include/metatensor/torch/labels.hpp
@@ -63,8 +63,8 @@ public:
     static TorchLabels empty(torch::IValue names);
 
     /// Create Labels with a single dimension with the given name and values in
-    /// the [0, stop) range
-    static TorchLabels range(std::string name, int64_t end);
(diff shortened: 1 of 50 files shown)


Luthaf:
You should be able to rebase & pick up the cmake fix from master!

metatensor-torch/include/metatensor/torch/labels.hpp
```diff
@@ -63,8 +63,8 @@ class METATENSOR_TORCH_EXPORT LabelsHolder: public torch::CustomClassHolder {
     static TorchLabels empty(torch::IValue names);
 
     /// Create Labels with a single dimension with the given name and values in
-    /// the [0, stop) range
-    static TorchLabels range(std::string name, int64_t end);
+    /// the [0, stop) range, on the given device
+    static TorchLabels range(std::string name, int64_t end, torch::IValue device = torch::IValue(torch::kCPU));
```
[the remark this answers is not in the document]
frostedoyster: Ok, fair, I've re-done it using `Labels.to()`. My concern is that it will be a bit slower to create the labels and then move them

python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -279,6 +279,20 @@ def eye_like(array, size: int):
...
+def get_device(array) -> Optional[torch_device]:
+    """
+    Returns the device of the array if it is a
+    ``torch.Tensor``, or None if it is a ``numpy.ndarray``.
+    """
+
+    if isinstance(array, TorchTensor):
+        return array.device
+    elif isinstance(array, np.ndarray):
+        return None
```
Luthaf, line 291: most of the code until now returned `"cpu"` for the device of numpy arrays, should this do the same?

python/metatensor-operations/metatensor/operations/block_from_array.py
```diff
@@ -48,14 +49,17 @@ def block_from_array(array) -> TensorBlock:
...
+    properties = Labels.range("property", shape[-1])
 
-    return TensorBlock(
-        values=array,
-        samples=Labels.range("sample", shape[0]),
-        components=components,
-        properties=Labels.range("property", shape[-1]),
-    )
+    device = _dispatch.get_device(array)
+    if device is not None:  # i.e. if array is a torch.Tensor
```
Luthaf, line 60: I don't think we need to guard on `None` here. `Labels.to` does nothing in metatensor-core, so the code should work fine!

python/metatensor-torch/tests/operations/block_from_array.py
```diff
@@ -1,11 +1,12 @@
+import pytest
 import torch
 from packaging import version
 
 import metatensor.torch
 
 
-def check_operation(block_from_array):
-    values = torch.arange(42).reshape(2, 3, 7).to(torch.float64)
+def check_operation(block_from_array, device):
```
Luthaf, line 8: just to be thorough, could you add a test in metatensor-core that creates a block with a `meta` torch tensor, and check that the labels are still on CPU?

metatensor-torch/include/metatensor/torch/labels.hpp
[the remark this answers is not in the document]
Luthaf: should be the same cost: we need the labels data on CPU anyway for libmetatensor.so, and the "on-device" version is always a copy. This will just invert the order of the copies.

PR #436, opened 2023-12-14
Fix slice

### LINKED ISSUE #432: Operations.slice assumes device is cpu/ incompatible with cuda
metatensor.operations.slice that in turn calls metatensor.operations.slice_block throws an error when tensormaps are originally on cuda, as it is not able to directly call .numpy() on the tensormap.block.values. 

For a while I have been using a quick "fix" in `operations/slice.py line 187 ` for the `axis="samples"` for my use case,

```
if axis == "samples":
        # only keep the same names as `labels`
        all_samples = block.samples.view(labels.names)
        # create an arrays of bools indicating which samples indices to keep
        samples_mask = _dispatch.bool_array_like(
            [all_samples.entry(i) in labels for i in range(len(all_samples))],
            block.values.cpu(),
        )
        device = block.values.device
        new_values = _dispatch.mask(block.values**.cpu()**, 0, samples_mask)
        new_samples = Labels(
            block.samples.names,
            _dispatch.mask(block.samples.values, 0, samples_mask),
        )

        new_block = TensorBlock(
            values=new_values.to(device),
            samples=new_samples,
            components=block.components,
            properties=block.properties,
        )

```
We might want to fix this in general (and not just for the specific use case that worked here)

frostedoyster:
Fixes #432 

----
📚 Documentation preview 📚: https://metatensor--436.org.readthedocs.build/en/436/

### DIFF
--- a/python/metatensor-operations/metatensor/operations/_dispatch.py
+++ b/python/metatensor-operations/metatensor/operations/_dispatch.py
@@ -381,6 +381,8 @@ def mask(array, axis: int, mask):
     if isinstance(array, TorchTensor):
         return torch.index_select(array, dim=axis, index=indices)
     elif isinstance(array, np.ndarray):
+        if isinstance(indices, TorchTensor):
+            indices = indices.detach().cpu().numpy()
         return np.take(array, indices, axis=axis)
     else:
         raise TypeError(UNKNOWN_ARRAY_TYPE)
--- a/python/metatensor-operations/tests/slice.py
+++ b/python/metatensor-operations/tests/slice.py
@@ -6,8 +6,10 @@ import pytest
 import metatensor
 from metatensor import Labels, TensorBlock, TensorMap
 
+
 try:
     import torch  # noqa
+
     HAS_TORCH = True
     if torch.cuda.is_available():
         HAS_CUDA = True
--- a/python/metatensor-operations/tests/slice.py
+++ b/python/metatensor-operations/tests/slice.py
@@ -17,6 +17,7 @@ try:
         HAS_CUDA = False
 except ImportError:
     HAS_TORCH = False
+    HAS_CUDA = False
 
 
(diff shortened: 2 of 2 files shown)


python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -381,6 +381,8 @@ def mask(array, axis: int, mask):
     if isinstance(array, TorchTensor):
         return torch.index_select(array, dim=axis, index=indices)
     elif isinstance(array, np.ndarray):
+        if isinstance(indices, TorchTensor):
```
Luthaf, line 384: I don't understand how this happens. I can see us indexing torch values with numpy indices (using classes from metatensor-core), but I don't see how the reverse happens here.

python/metatensor-operations/tests/slice.py
```diff
@@ -515,6 +528,26 @@ def test_slicing_all(tensor):
     )
 
 
+@pytest.mark.skipif(not HAS_CUDA, reason="requires cuda")
+@pytest.mark.skipif(not HAS_TORCH, reason="requires torch")
+def test_slice_different_device():
+    array = torch.empty((6, 5, 7), device="cuda")  # doesn't work with meta
```
Luthaf, line 534:
```suggestion
    array = torch.empty((6, 5, 7), device="cuda")  # doesn't work with device=meta
```

python/metatensor-operations/metatensor/operations/_dispatch.py
frostedoyster: I think that, with slice, we're at some point indexing both labels and the data tensor with the same index tensor. The index tensor is a torch.Tensor, while the labels are a numpy array

Luthaf: ok, that makes sense.

PR #453, opened 2024-01-22
Export load/save as free-standing functions and class members

Luthaf:
Export load/save as free-standing functions and class members

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--453.org.readthedocs.build/en/453/

### DIFF
--- a/python/metatensor-torch/tests/serialization.py
+++ b/python/metatensor-torch/tests/serialization.py
@@ -1,5 +1,6 @@
 import os
 
+import numpy as np
 import torch
 
 import metatensor.torch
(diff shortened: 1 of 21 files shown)


python/metatensor-torch/tests/serialization.py
```diff
@@ -27,18 +28,40 @@ def check_tensor(tensor):
...
-        ),
+    path = os.path.join(
+        os.path.dirname(__file__),
+        "..",
+        "..",
+        "..",
+        "metatensor",
+        "tests",
+        "data.npz",
+    )
```
PicoCentauri, line 39: We could think using pathlib at some point.

PicoCentauri (commented):
As we already discussed: I like it.

python/metatensor-torch/tests/serialization.py
Luthaf: Not sure what this would give us? `/` operator overloading mainly, or is there something else?

I'm not sure I love`/` operator overloading  to be honest =)

PicoCentauri: We could avoid the stacking of  ".." and write it like

```python
    path = Path(__file__).parents[3] / "metatensor" / "tests" / "data.npz"
```

I also don't know if I am a big fan of overloading though.

PR #455, opened 2024-01-24
Labels serialization

### LINKED ISSUE #321: Labels cannot be pickled
Hello, just updated my code to the newest version of equistore (dev333) and found the issue that Labels cannot be pickled (and thus not be saved with pickle.dumps)

Here a small example to replicate the error:

```py
from equistore import Labels
import pickle

l = Labels.empty("_")
pickle.dumps(l)
```

Gives this error:

```
  File "<stdin>", line 1, in <module>
AttributeError: Can't pickle local object 'CDLL.__init__.<locals>._FuncPtr'
```

Luthaf:
Fix #321 by adding `load_labels`, `load_labels_buffer`, `save_labels` and `save_labels_buffer` to all languages API.

Built on top of #453

# Reviewer checklist


Luthaf:
As usual, this is best reviewed one commit at the time!

metatensor-core/CHANGELOG.md
```diff
@@ -44,7 +52,9 @@ a changelog](https://keepachangelog.com/en/1.1.0/) format. This project follows
   dtype used by the data.
 - Offer serialization functionality as member functions (i.e. `TensorMap.load`)
   in addition to the existing free standing functions (i.e. `metatensor.load`) (#453)
-
+- Serialization of labels with `Labels.save()`, `Labels.load`,
```
frostedoyster, line 55:
```suggestion
- Serialization of labels with `Labels.save`, `Labels.load`,
```

metatensor-core/include/metatensor.h
```diff
@@ -831,6 +844,87 @@ struct mts_tensormap_t *mts_tensormap_keys_to_samples(const struct mts_tensormap
...
+/**
+ * Save labels to an in-memory buffer.
+ *
+ * On input, `*buffer` should contain the address of a starting buffer (which
+ * can be NULL) and `*buffer_count` should contain the size of the allocation
+ * (0 if `*buffer` is NULL).
+ *
+ * On output, `*buffer` will contain the serialized data, and `*buffer_count`
+ * the total number of written bytes (which might be less than the allocation
+ * size).
```
frostedoyster, line 904: So it allocates its own memory if it's passed NULL?

metatensor-core/src/c_api/io/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+
+            let new_ptr = unsafe {
+                (self.realloc)(self.realloc_user_data, *self.data, new_size)
+            };
+
+            if new_ptr.is_null() {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::OutOfMemory,
+                    "failed to allocate memory with the realloc callback"
+                ));
```
frostedoyster, line 59: Can we always be sure that a null pointer here means out of memory?

metatensor-torch/src/misc.cpp
```diff
@@ -54,3 +54,20 @@ void metatensor_torch::save(const std::string& path, TorchTensorMap tensor) {
...
+
+TorchLabels metatensor_torch::load_labels(const std::string& path) {
+    return LabelsHolder::load(path);
+}
+
+TorchLabels metatensor_torch::load_labels_buffer(torch::Tensor buffer) {
+    return LabelsHolder::load_buffer(buffer);
+}
+
+void metatensor_torch::save(const std::string& path, TorchLabels labels) {
```
frostedoyster, line 67: How about `save_labels` since this can only save labels?
Also, later we use a `load` function to load `TensorMap`s

metatensor-core/include/metatensor.h
Luthaf: If `buffer` starts as NULL, `realloc` should allocate memory for it yes!

metatensor-core/src/c_api/io/mod.rs
Luthaf: It's not necessarily out of memory, but it indicates a failure to allocate, which will be an out of memory in 90% of the cases.

This is also the best matching error kind I could find in https://doc.rust-lang.org/std/io/enum.ErrorKind.html

metatensor-torch/src/misc.cpp
Luthaf: So I had a discussion here with @PicoCentauri who suggested this approach. We can make the save function take both labels and tensormap, and it does make for a cleaner API I think.

This is harder to do for loading, since overloading on return type is not possible in C++.

I'm not opposed to making the functions more symetric and have this function renamed to `save_labels` everywhere.

metatensor-core/src/c_api/io/mod.rs
frostedoyster: Ok, fair

metatensor-torch/src/misc.cpp
frostedoyster: I missed the overload. I think it's a good solution

frostedoyster (approved):
Thanks a lot for the work!

metatensor-core/src/io/load.rs
```diff
@@ -1,15 +1,45 @@
...
 
 use crate::{TensorMap, TensorBlock, Labels, Error, mts_array_t};
 
-use super::check_for_extra_bytes;
-use super::labels::read_npy_labels;
+use super::{check_for_extra_bytes, PathOrBuffer};
+use super::labels::load_labels;
 use super::npy_header::{Header, DataType};
 
+pub fn looks_like_tensormap_data(mut data: PathOrBuffer) -> bool {
```
agoscinski, line 14: Maybe small doc, not sure if I understand correctly so here as suggestion
```suggestion
/// Extracts from the path or buffer a ZipArchive and considers it to be a TensorMap if this archive contains a "keys.npy" entry.
pub fn looks_like_tensormap_data(mut data: PathOrBuffer) -> bool {
```

metatensor-core/src/io/labels.rs
```diff
@@ -1,17 +1,43 @@
+use std::io::BufReader;
+
 use byteorder::{LittleEndian, ReadBytesExt, BigEndian, WriteBytesExt, NativeEndian};
 
 use super::npy_header::{Header, DataType};
-use super::check_for_extra_bytes;
+use super::{check_for_extra_bytes, PathOrBuffer};
 use crate::{Error, Labels, LabelsBuilder, LabelValue};
 
+pub fn looks_like_labels_data(mut data: PathOrBuffer) -> bool {
```
agoscinski, line 9: Maybe small doc what "looks like" means, not sure when `Header::from_reader(buffer)` is true

metatensor-torch/src/register.cpp
```diff
@@ -215,12 +268,28 @@ TORCH_LIBRARY(metatensor, m) {
...
+        "load_buffer(Tensor buffer) -> __torch__.torch.classes.metatensor.TensorMap",
+        metatensor_torch::load_buffer
+    );
+
+    m.def(
+        "load_labels(str path) -> __torch__.torch.classes.metatensor.Labels",
+        metatensor_torch::load_labels
+    );
+    m.def(
+        "load_labels_buffer(Tensor buffer) -> __torch__.torch.classes.metatensor.Labels",
```
agoscinski, line 287: This looks weird, this function is covered by the metatensor-core Python API with the `load_labels`, so why don't we cover it here too by `load_labels`? Same as we do it for `save_buffer`.

metatensor-core/CHANGELOG.md
agoscinski: Isn't this a valid suggestion to be consistent? (Sorry for being pedantic)

agoscinski (commented):
The PR implements the io utilities to load/save labels from/to a buffer or path, so the serialization of the Labels in the Python API by implementing the Pickle object API and (`__reduce_ex__`), adding constructor, the same goes for TorchScript API. This PR is very similar to the PR https://github.com/lab-cosmo/metatensor/pull/298 for tensor maps.

metatensor-core/CHANGELOG.md
Luthaf: it is a valid suggestion, I did not implement it yet!

metatensor-torch/src/register.cpp
Luthaf: not sure what you mean? This is the same function as the one in `-core`, but registered as a custom TorchScript operation (so models can use it). And we have both `load_labels` and `load_labels_buffer` in both cases.

agoscinski: On the doc preview I don see `load_labels_buffer` in the Python API, maybe it is then just missing in the autofunction .rst? 
https://metatensor--455.org.readthedocs.build/en/455/core/reference/python/io.html#metatensor.load_labels

Luthaf: Oh, right. There is indeed a different API here, `load_labels_buffer` is accessible by giving a file-like object (which can be a buffer) to `load_labels`. We could add a more explicit `load_labels_buffer(buffer: bytes | bytearray | memoryview)` function to the -core API.

I'll see if it is simple enough to include in this PR!

metatensor-core/src/io/load.rs
Luthaf: done!

metatensor-core/src/io/labels.rs
Luthaf: done

metatensor-torch/src/register.cpp
Luthaf: done

agoscinski (approved):
looks good, thanks for the adjustments!

Issue #430, opened 2023-12-12
Atomistic models follow-up


Luthaf:
#405 adds a first definition of atomistic models, here are a couple of things left out of the initial PR that will need to be done:

- [x] Example using the code to export a model & run a simulation with it #431
- [x] Example on how to profile model execution
- [x] Which metadata should be recorded by the model, the discussion is here: https://github.com/lab-cosmo/metatensor/pull/405#discussion_r1373047393

Luthaf:
- [ ] Add support for `todict` and `fromdict` in ASE calculator

Luthaf:
- [ ] pick the best device according to the model & what's available in ASE calculator

Luthaf:
- [ ] handle out of order per-atom energies

PR #472, opened 2024-02-01
Implement todict and fromdict for ASE calculator

Luthaf:
Part of #430.

We added an ASE calculator for metatensor models in #405, leaving `fromdict` and `todict` unimplemented, this PR tries to add these functions (which are standard and required in ASE as far as I understand).

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--472.org.readthedocs.build/en/472/

### DIFF
--- a/python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
+++ b/python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
@@ -59,12 +59,17 @@ class MetatensorCalculator(ase.calculators.calculator.Calculator):
         """
         super().__init__()
 
+        self.parameters = {
+            "check_consistency": check_consistency,
+        }
+
         if isinstance(model, (str, bytes, pathlib.PurePath)):
             if not os.path.exists(model):
                 raise InputError(f"given model path '{model}' does not exist")
 
             check_atomistic_model(model)
             self._model = torch.jit.load(model)
+            self.parameters["model_path"] = str(model)
         elif isinstance(model, torch.jit.RecursiveScriptModule):
             if model.original_name != "MetatensorAtomisticModel":
                 raise InputError(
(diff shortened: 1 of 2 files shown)


python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -88,12 +90,23 @@ def __init__(
 
     def todict(self):
         # used by ASE to save the calculator
-        raise NotImplementedError("todict is not yet implemented")
+        try:
+            model = torch.jit.script(self._model)
+        except RuntimeError as e:
+            raise RuntimeError("could not convert the module to TorchScript") from e
```
sirmarcel, line 96: I'm unfamiliar with this syntax, what does this line do with `e`?

sirmarcel (commented):
As far as I understand this code, which is only medium given my limited understanding of `metatensor` and `torchscript`, this is fine technically.

My only concern is more philosophical: If we consider that `metatensor` will probably at some point deal with models that are "thick" in the sense of having millions of weights, is it a good idea to have an automated facility that dumps the serialised version of these weights into every timestep of a `.traj`? 

As far as I know, `ase` expects `todict` to return something like the parameters of an electronic structure code, rather than a full serialisation of the underlying calculator. So I think the more `ase`-style approach would be to return a dict with the path to the serialised model, rather than the full serialised model. Of course, at the cost of portability... 

So, overall, this is fine to merge in principle, but I would be curious what the intended use case for this is, and see whether this can be better served in some other way.

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
Luthaf: This creates a new exception with a `cause` field, which is then displayed to the user:

```py
try:
    raise Exception("first one")
except Exception as e:
    raise Exception("second one") from e
```

```
  File "tmp.py", line 2, in <module>
    raise Exception("first one")
Exception: first one

The above exception was the direct cause of the following exception:

  File "tmp.py", line 4, in <module>
    raise Exception("second one") from e
Exception: second one
```

Luthaf:
I did not realize this was dumped at every timestep. Agreed that this is not ideal. If the user gave us the path to a model, saving the path again makes sense, not sure how we should handle the case where the user gives a large in-memory model.

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
sirmarcel: Okay -- thanks.

sirmarcel:
Yeah, it looks like this gets called at every timestep: https://gitlab.com/ase/ase/-/blob/master/ase/io/trajectory.py#LC174. I don't think this gets used by particularly serious MD codes, so maybe this is completely fine, but it should probably at least be documented in case someone wonders why their `.traj` files are extremely huge...

sirmarcel:
Okay, I saw your comment now. As far as I know, this is not a required interface for `ase`.  The only thing you need to have is a `parameters` dict (but it doesn't matter particularly what's in there). So one might want to write something informative there, like where we found the model that made up the calculator, but unless we have a pressing need to implement the `todict`/`fromdict` interface, I personally wouldn't bother keeping it around and having it potentially cause issues.

Luthaf:
I've updated the code to store the model path in `todict`, which should solve the size issue. The error message should also point users who need this to work to save the model in a file themself. 

> As far as I know, this is not a required interface for ase.

It is part of the interface and used by some of the code in ASE =/. I agree that we should not be implementing everything just for fun, but the idea is also that this project should be usable by people outside the lab, who might not take the time to report issues if something is broken for them. It seems to me that this code is simple enough to not add too much maintenance work in the future =)

sirmarcel:
Sounds good -- what I meant is that `ase` (as far as I know, as discussed in Slack) doesn't expect these functions to do anything very well-defined, and I think in principle one can get away with only `todict` and not `fromdict`. I think this is a good compromise, we can always adjust further if user needs arise. Once the pipeline is done, this can be merged. Thank you!

Luthaf:
codecov is not behaving today ...

PR #478, opened 2024-02-02
Allow operations to be `torch.jit.save`d

### LINKED ISSUE #410: Internal Torch error when trying to save scripted opteration
With this code:

```py
fn = torch.jit.script(metatensor.torch.sum_over_samples)
torch.jit.save(fn, os.path.join(tmpdir, "exported.pt"))
```

I'm getting the following error:

```
RuntimeError: false INTERNAL ASSERT FAILED at "/Users/runner/work/pytorch/pytorch/pytorch/torch/csrc/jit/api/function_impl.cpp":53, please report a bug to PyTorch. Failed to downcast a Function to a GraphFunction. This probably indicates that the JIT calling context needs a special case on tryToGraphFunction() instead.
```

This also happens for other operations, and does not seems related to `sum_over_samples` specifically. We should prepare a minimal reproduction and submit the corresponding bug to PyTorch.

frostedoyster:
Fix #410

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--478.org.readthedocs.build/en/478/


python/metatensor-operations/metatensor/operations/block_from_array.py
```diff
@@ -50,11 +50,20 @@ def block_from_array(array) -> TensorBlock:
         )
 
     samples = Labels.range("sample", shape[0])
+    Labels(
+        names="sample",
+        values=_dispatch.int_array_like(range(shape[0]), array),
+    )
```
Luthaf, line 56:
```suggestion
    samples = Labels(
        names="sample",
        values=_dispatch.int_array_like(range(shape[0]), array),
    )
```

frostedoyster: Ahh well spotted :)
Eventually it made everything fail so I fixed it

python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -320,10 +320,17 @@ def index_add(output_array, input_array, index):
 def int_array_like(int_list: List[int], like):
     """
     Converts the input list of int to a numpy array or torch tensor
-    based on the type of `like`.
+    based on the device of `like`.
+
+    If the backend is torch and the device is "meta",
+    the device is set to "cpu".
```
Luthaf, line 326: could you explain why this is necessary?

frostedoyster: Otherwise in some cases some Labels are created on the meta device (following the fact that the corresponding block is on meta) and that gives an error

Luthaf: can we give existing `labels.values` instead of the `block.values` to be the `like` argument?

Luthaf: Then could you add a comment saying that we use meta in tests and that it makes this function fails in this case?

Luthaf (approved):

PR #479, opened 2024-02-02
`metatensor-operations` finite difference test for add

agoscinski:
This PR adds utilities for doing finite difference tests for the metatensor-operations and uses them in the simple example of the `add` operation.
This is the first PR addressing issue https://github.com/lab-cosmo/metatensor/issues/84
The idea of the testing utilities originates from issue https://github.com/lab-cosmo/metatensor/issues/469

In the next run on this PR I want to dispatch the testing utilities so they can also be used for torch, but before that I would like to have some general feedback if this is going in a good direction.

# TODOs:
- [x] The documentation is still needs some work. Also would add the testing utils to the doc.
- [x] function `cartesian_linear` needs to be implemented

# Contributor (creator of pull-request) checklist

 - [x] Tests updated (for new features and bugfixes)?
 - [x] Documentation updated (for new features)?
 - [x] Issue referenced (for PRs that solve an issue)?

# Reviewer checklist

 - [ ] CHANGELOG updated with public API or any other important changes?

----
📚 Documentation preview 📚: https://metatensor--479.org.readthedocs.build/en/479/

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1663450013.zip)

PR #480, opened 2024-02-02
Fix #470 non adapted sample indices in gradients during sorting

### LINKED ISSUE #470: `sort` does not adapt samples indices in gradients to the sorted samples in the block the gradients are attached to
```python
import numpy as np
import metatensor
from metatensor import TensorMap, TensorBlock, Labels
block_1 = TensorBlock(
    values=np.array([[3, 5], [1, 2]]),
    samples=Labels(["s"], np.array([[2], [0]])),
    components=[],
    properties=Labels(["p"], np.array([[0], [1]])),
)
block_1.add_gradient(
    parameter="g",
    gradient=TensorBlock(
        values=np.array(
            [[[11, 10],
              [13, 12]],
             [[15, 14],
              [11, 10]],
             [[13, 12],
              [15, 14]]]
        ),
        samples=Labels(
            ["sample", "g"],
            np.array([[0, 1], [0, 2], [1, 1]]),
        ),
        components=[
            Labels(["c"], np.array([[0], [1]])),
        ],
        properties=Labels(["p"], np.array([[0], [1]])),
    ),
)
keys = Labels(names=["_"], values=np.array([[0]]))
tensor = TensorMap(keys, [block_1])
sorted_tensor = metatensor.sort(tensor)
print("tensor")
print(tensor[0].samples)
print(tensor[0].gradient('g').samples)
print()
print("sorted tensor")
print(sorted_tensor[0].samples)
print(sorted_tensor[0].gradient('g').samples)
```
Output:
```bash
tensor
Labels(
    s
    2
    0
)
Labels(
    sample  g
      0     1
      0     2
      1     1
)

sorted tensor
Labels(
    s
    0
    2
)
Labels(
    sample  g
      0     1
      0     2
      1     1
)
```
The gradients assignment to the samples gets swapped in this example. I think this would have been caught by finite difference tests https://github.com/lab-cosmo/metatensor/issues/84. So I look how feasible this is to do for sort.

EDIT: Realised that the sort operation is not really suitable for gradcheck. Will just add regular test

agoscinski:
Fixes #470

- Splitting sorted tensors into an ascending and descending so the fix can be more robustly tested
- Adding dispatched arange_like to create inverse mapping for reasigning samples in gradients

Waiting till https://github.com/lab-cosmo/metatensor/pull/479 is merged to also add finite difference test

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--480.org.readthedocs.build/en/480/

### DIFF
--- a/python/metatensor-operations/metatensor/operations/sort.py
+++ b/python/metatensor-operations/metatensor/operations/sort.py
@@ -4,6 +4,84 @@ from . import _dispatch
 from ._classes import Labels, TensorBlock, TensorMap, torch_jit_is_scripting
 
 
+def _sort_single_gradient_block(
+    block: TensorBlock,
+    gradient_block: TensorBlock,
+    axes: List[str],
+    descending: bool,
+) -> TensorBlock:
+    """
+    Sorts a single gradient tensor block given the tensor block which the gradients are
+    attached to. This function does not check the user input.  This is different from
+    `_sort_single_block` because we need to update the sample differently (since
+    gradient samples are pointers into the values samples).
+    """
+
+    sample_names = gradient_block.samples.names
+    sample_values = gradient_block.samples.values
+
+    component_names: List[List[str]] = []
+    components_values = []
+    for component in gradient_block.components:
+        component_names.append(component.names)
+        components_values.append(component.values)
+
+    property_names = gradient_block.properties.names
+    properties_values = gradient_block.properties.values
+
+    values = gradient_block.values
+    if "samples" in axes:
+        # we first need to get the mapping induced by the sorting in its parent block
+        # so we can change the sample column label entries so it matches the ones of
+        # the parent block
+        block_sample_values = block.samples.values
+        # sample index -> sample labels
+        sorted_idx = _dispatch.argsort_labels_values(
+            block_sample_values, reverse=descending
+        )
+        # obtain inverse mapping sample labels -> sample index
+        sorted_idx_inverse = _dispatch.empty_like(sorted_idx, shape=(len(sorted_idx),))
+        sorted_idx_inverse[sorted_idx] = _dispatch.int_array_like(
+            list(range(len(sorted_idx))), sorted_idx
+        )
+        # adapt sample column in gradient samples to the one of the sorted values of
+        # the gradient_block the gradient is attached to
+        sample_values = _dispatch.copy(sample_values)
+        sample_values[:, 0] = sorted_idx_inverse[sample_values[:, 0]]
+
+        # sort the samples in gradient regularly moving the rows considering all columns
+        sorted_idx = _dispatch.argsort_labels_values(sample_values, reverse=descending)
+        sample_values = sample_values[sorted_idx]
+        values = values[sorted_idx]
+    if "components" in axes:
+        for i, _ in enumerate(gradient_block.components):
+            sorted_idx = _dispatch.argsort_labels_values(
+                components_values[i], reverse=descending
+            )
+            components_values[i] = components_values[i][sorted_idx]
+            values = _dispatch.take(values, sorted_idx, axis=i + 1)
+    if "properties" in axes:
+        sorted_idx = _dispatch.argsort_labels_values(
+            properties_values, reverse=descending
+        )
+        properties_values = properties_values[sorted_idx]
+        values = _dispatch.take(values, sorted_idx, axis=-1)
+
+    samples_labels = Labels(names=sample_names, values=sample_values)
+    properties_labels = Labels(names=property_names, values=properties_values)
+    components_labels = [
+        Labels(names=component_names[i], values=components_values[i])
+        for i in range(len(component_names))
+    ]
+
+    return TensorBlock(
+        values=values,
+        samples=samples_labels,
+        components=components_labels,
+        properties=properties_labels,
+    )
+
+
 def _sort_single_block(
     block: TensorBlock,
     axes: List[str],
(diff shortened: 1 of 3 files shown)


python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -715,3 +715,21 @@ def zeros_like(array, shape: Optional[List[int]] = None, requires_grad: bool = F
         return np.zeros_like(array, shape=shape, subok=False)
     else:
         raise TypeError(UNKNOWN_ARRAY_TYPE)
+
+
+def arange_like(array, start, end, requires_grad: bool = False):
```
Luthaf, line 720: I don't see the point of requires_grad here, since this is intended to work with integer values, right?

agoscinski: yes makes sense, I just copied pasted from zeros_like

agoscinski: removed arange completely and used int_like as you suggested in another PR

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -4,6 +4,79 @@
...
 
 
+def _sort_single_gradient_block(
+    block: TensorBlock,
+    gradient_block: TensorBlock,
+    axes: List[str],
+    descending: bool,
+) -> TensorBlock:
+    """
+    Sorts a single TensorBlock without the user input checking and sorting of gradients
```
Luthaf, line 14:
```suggestion
    Sorts a single gradient TensorBlock. This is different from `_sort_single_block` because we need to update the sample differently (since gradient samples are pointers into the values samples)
```

python/metatensor-operations/metatensor/operations/sort.py
```diff
@@ -4,6 +4,79 @@
...
+    components_values = []
+    for component in gradient_block.components:
+        component_names.append(component.names)
+        components_values.append(component.values)
+
+    property_names = gradient_block.properties.names
+    properties_values = gradient_block.properties.values
+
+    values = gradient_block.values
+    if "samples" in axes:
```
Luthaf, line 30: just to be clear, you don't re-sort the "sample" column but update it according to the values samples sorting. Does this re-sort the other columns in the samples?

python/metatensor-operations/tests/sort.py
```diff
@@ -63,11 +62,12 @@ def tensor():
     )
     # fmt: on
     keys = Labels(names=["key_1", "key_2"], values=np.array([[1, 0], [0, 0]]))
+    # block order is descending
     return TensorMap(keys, [block_2, block_1])
 
 
 @pytest.fixture
-def tensor_sorted():
+def tensor_sorted_ascending():
```
Luthaf, line 70: maybe a small docstring explaining the difference with the other fixture?

python/metatensor-operations/tests/sort.py
```diff
@@ -124,31 +126,87 @@ def tensor_sorted():
     return TensorMap(keys, [block_1, block_2])
 
 
-def test_sort(tensor, tensor_sorted):
+@pytest.fixture
+def tensor_sorted_descending():
```
Luthaf, line 130: same!

python/metatensor-operations/tests/sort.py
```diff
@@ -124,31 +126,87 @@ def tensor_sorted():
...
+    block_1 = TensorBlock(
+        values=np.array([[3, 5], [1, 2]]),
+        samples=Labels(["s"], np.array([[2], [0]])),
+        components=[],
+        properties=Labels(["p"], np.array([[1], [0]])),
+    )
+
+    # fmt: off
```
Luthaf, line 138: this should not be needed

python/metatensor-torch/tests/operations/sort.py
```diff
@@ -65,11 +65,12 @@ def tensor():
     )
     # fmt: on
     keys = Labels(names=["key_1", "key_2"], values=torch.tensor([[1, 0], [0, 0]]))
+    # block order is descending
     return TensorMap(keys, [block_2, block_1])
 
 
 @pytest.fixture
-def tensor_sorted():
+def tensor_sorted_ascending():
```
Luthaf, line 73: same about the docstring

python/metatensor-operations/tests/sort.py
agoscinski: not needed but black makes this really hard to read so I turned it off

python/metatensor-operations/metatensor/operations/sort.py
agoscinski: I updated the in-code doc a little bit

python/metatensor-operations/metatensor/operations/sort.py
agoscinski: updated this doc and embedded suggested change into it

python/metatensor-operations/metatensor/operations/sort.py
Luthaf: sounds good! I had a hard time following some of this

python/metatensor-operations/tests/sort.py
agoscinski: This have strong feelings on. This sort test is already hard to understand as it is, and the black formatting made it much harder.

agoscinski: Unlike other tests, the order must be much easier to read so one can follow what is how moved. I added comments why it is turned off.

PR #486, opened 2024-02-05
Use a local build of the core library in julia

Luthaf:
When working on the julia code we need to have the most up to date version of the core library; which is not always the case when using Metatensor_jll (which can only contain releases). This PR adds a way to build and use a local version of the metatensor-core shared library with julia, and documents it as the main way to interact with the code.

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--486.org.readthedocs.build/en/486/


julia/deps/build_local.jl
```diff
@@ -0,0 +1,68 @@
...
+end
+
+if ismissing(lib_path)
+    error("could not find metatensor library in install directory")
+end
+
+# Save the path to the freshly built library in preferences, it will be checked
+# when loading Metatensor.
+@info "Setting preference to use local build in $(joinpath(project_path, "LocalPreferences.toml"))"
+set_preferences!(
```
agoscinski, line 63: interesting the help page does not show all signatures
```
help?>  set_preferences!
search:        

  set_preferences!(uuid_or_module, prefs::Pair{String,Any}...; export_prefs=false,
                   active_project_only=true, force=false)                                                                                                                            
```
but it internally actually knows it
```
julia> set_preferences!(
           "Metatensor",
           "libmetatensor" => lib_path;
           force=true
       )
ERROR: MethodError: no method matching set_preferences!(::String, ::Pair{String, String}; force::Bool)

Closest candidates are:
  set_preferences!(::Module, ::Pair{String}...; kwargs...)
   @ Preferences /opt/anaconda3/envs/metatensor-dev/share/julia/packages/Preferences/TTEAN/src/Preferences.jl:285
  set_preferences!(::Base.UUID, ::Pair{String}...; export_prefs, active_project_only, kwargs...)
   @ Preferences /opt/anaconda3/envs/metatensor-dev/share/julia/packages/Preferences/TTEAN/src/Preferences.jl:205
  set_preferences!(::String, ::String, Pair{String}...; force)
   @ Preferences /opt/anaconda3/envs/metatensor-dev/share/julia/packages/Preferences/TTEAN/src/Preferences.jl:115

Stacktrace:
 [1] top-level scope
   @ REPL[42]:1
```

agoscinski (approved):
LGTM

julia/README.md
```diff
@@ -5,19 +5,35 @@ This package contains the Julia bindings to the C API of metatensor.
...
-julia
+julia julia/deps/build_local.jl
 ```
 
-From the Julia prompt:
+To setup the environment, start a Julia REPL by running `julia`, and then
 
 ```julia
-] activate .
+] add https://github.com/Luthaf/Metatensor_jll.jl
```
agoscinski, line 25: is this correct?

agoscinski (changes requested):

julia/deps/build_local.jl
```diff
@@ -0,0 +1,68 @@
+using Pkg
+# save the currently active project path to use later when saving preferences
+project_path = dirname(Pkg.project().path)
+
+# Install build time depenedencies
+Pkg.activate(@__DIR__)
+Pkg.instantiate()
+
+
+using Scratch, Preferences, CMake_jll
```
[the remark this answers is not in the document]
Luthaf: Most of the code in the wild is written with `using`, which makes some sense for multiple dispatch. It does make it a lot harder to know where a function/method comes from as you realized.

julia/deps/build_local.jl
Luthaf: Yes, the help page is potentially user generated, but the method dispatch uses the full compiler

agoscinski (approved):

PR #496, opened 2024-02-07
`metatensor-learn`: Set correct device for output properties when `to` method is used from module

### LINKED ISSUE #554: `ModuleMap`'s handling of devices
At the moment, `ModuleMap` takes a device at initialization and there is no way to change it later. However, this does not correspond to idiomatic torch, where modules can be moved from and to devices. Since `Labels` cannot be registered as part of a `torch.nn.Module`, I think the only possibility for now is to move the `Labels` to the correct device during the forward function. Another option would be to override the `to` function of `ModuleMap` so the `Labels` are moved as well, but I do not know how difficult this would be.
@agoscinski @jwa7 @abmazitov @Luthaf

agoscinski:
The `out_properties` labels are not moved to the correct device when using modules `to` function. This affects the `metatensor-learn` nn modules when using it with the TorchScript API, since in `metatensor-core` Labels are anyway always on CPU. So far I only created tests that detect the bug (which does not fail in the CI since one requires two devices). Closes #554.

# TODO:
- [x] fix the issue by overwriting `to` function

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--496.org.readthedocs.build/en/496/


python/metatensor-learn/metatensor/learn/nn/module_map.py
```diff
@@ -341,3 +342,51 @@ def repr_as_module_dict(self) -> str:
             representation += f"  ({key!r}): {self[i]!r}\n"
         representation += ")"
         return representation
+
+    def to(self, *args, **kwargs):
```
Luthaf, line 346: this seems extremely overkill. Why is this not using `super().to()`?

```py
def to(...):
    self._in_keys = ...
    self._out_properties = ...
    
    super().to(...)
```

python/metatensor-learn/tests/module_map.py
```diff
@@ -87,3 +96,84 @@ def test_module_map_single_block_tensor(single_block_tensor_torch, out_propertie
...
+        modules.append(
+            MockModule(
+                in_features=len(single_block_tensor_torch.block(key).properties),
+                out_features=5,
+            )
+        )
+    tensor_module = ModuleMap(
+        single_block_tensor_torch.keys, modules, out_properties=out_properties
+    )
+    tensor_module.to("cpu")
```
Luthaf, line 119: this does nothing, right? Since everything is already on CPU

python/metatensor-learn/tests/module_map.py
```diff
@@ -87,3 +96,84 @@ def test_module_map_single_block_tensor(single_block_tensor_torch, out_propertie
...
+        )
+
+    # labels are always on cpu in metatensor-core so we don't need to move them to
+    # default device
+    tensor_module = ModuleMap(
+        single_block_tensor_torch.keys, modules, out_properties=out_properties
+    )
+    tensor_module.to("cpu")
+    single_block_tensor_torch = single_block_tensor_torch.to(device="cpu")
+    with torch.no_grad():
```
Luthaf, line 155: why no_grad? Seems useless here

python/metatensor-learn/tests/module_map.py
```diff
@@ -87,3 +96,84 @@ def test_module_map_single_block_tensor(single_block_tensor_torch, out_propertie
...
+        out_tensor = tensor_module(single_block_tensor_torch)
+    assert out_tensor.device.type == "cpu"
+
+    # cpu -> cuda
+    torch.set_default_device("cpu")
```
Luthaf, line 160: why a new module map instead of moving the one we already have back to CPU?

python/metatensor-learn/tests/module_map.py
```diff
@@ -87,3 +96,84 @@ def test_module_map_single_block_tensor(single_block_tensor_torch, out_propertie
...
+    # labels are always on cpu in metatensor-core so we don't need to move them to
+    # default device
+    tensor_module = ModuleMap(
+        single_block_tensor_torch.keys, modules, out_properties=out_properties
+    )
+    tensor_module.to("cpu")
+    single_block_tensor_torch = single_block_tensor_torch.to(device="cpu")
+    with torch.no_grad():
+        out_tensor = tensor_module(single_block_tensor_torch)
+    assert out_tensor.device.type == "cpu"
```
Luthaf, line 157: this should also check that the module's parameters have been moved as well

python/metatensor-learn/tests/module_map.py
```diff
@@ -87,3 +96,84 @@ def test_module_map_single_block_tensor(single_block_tensor_torch, out_propertie
...
+    "out_properties", [None, [Labels(["a", "b"], np.array([[1, 1]]))]]
+)
+@pytest.mark.skipif(not HAS_CUDA, reason="requires cuda")
+def test_cuda_module_map_to(single_block_tensor_torch, out_properties):  # noqa F811
+    """
+    We set the correct default device for initialization and check if the module this
+    works once the default device has been changed. This catches cases where the default
+    device or a hard coded device is used for tensor created within the forward
+    function.
+    """
```
Luthaf, line 136: I would also like a test for moving dtypes

python/metatensor-torch/tests/learn/module_map.py
```diff
@@ -84,6 +93,70 @@ def test_module_map_single_block_tensor(single_block_tensor, out_properties):
                 assert out_block.gradient(parameter).properties == out_properties[0]
 
 
+@pytest.mark.parametrize(
+    "out_properties", [None, [Labels(["a", "b"], torch.tensor([[1, 1]]))]]
+)
+@pytest.mark.skipif(not HAS_CUDA, reason="requires cuda")
+def test_cuda_module_map_to(single_block_tensor, out_properties):  # noqa F811
```
Luthaf, line 100: similar comments here as for the other tests

Luthaf (commented):
I would also like to see a check on a TorchScript-compiled ModuleMap, to ensure we can call this function from C++ when moving a trained model to GPU.

python/metatensor-learn/metatensor/learn/nn/module_map.py
frostedoyster: Ok! But we'll need one more line to extract the device

python/metatensor-learn/tests/module_map.py
frostedoyster: Yes, you're right, I will try meta

python/metatensor-learn/tests/module_map.py
frostedoyster: My bad

frostedoyster:
@Luthaf I'm having a lot of trouble using `super()` in torchscript. I had a look around but I can't understand if it's supported.
Still to be done: the dtype tests.
Also, what is the expected behavior for the `meta` device? Are the labels supposed to stay on CPU?

Luthaf:
> @Luthaf I'm having a lot of trouble using super() in torchscript. I had a look around but I can't understand if it's supported.

Right, it's possible that it's not. I'll check if one can override `to` in C++ in a different way.

>  Also, what is the expected behavior for the meta device? Are the labels supposed to stay on CPU?

For `metatensor.torch.Labels`, we have two copies of the values. One is internal to metatensor and always on CPU, but not accessible by the user. And the other is a `torch.Tensor` with the same data, which is moved between devices. I think creating the Labels on CPU and moving to meta should work, but not 100% sure. Since the data inside the tensor matter, maybe we can not use meta for tests here.

Luthaf:
So, slight issue here: the code for `Module::to` in C++ does not use TorchScript: https://github.com/pytorch/pytorch/blob/eb1d6ed9f9d9731401b04382f526a64e6d27b6e6/torch/csrc/jit/api/module.cpp#L182-L192

Only parameters and buffers are moved. So we might have to register the labels/blocks/tensormap values as buffers if we want them to be properly moved around all the time. I'll give it a try. Hopefully we should be able to provide functions to make this easier:

```py
def register_buffers(module: torch.nn.Module, name: str, labels: Labels)

# this would create sub-buffers `<name>_samples`/`<name>_properties`/`<name>_values`
def register_buffers(module: torch.nn.Module, name: str, block: TensorBlock)

def register_buffers(module: torch.nn.Module, name: str, tensor: TensorMap)
```

frostedoyster:
Ok, thanks a lot. For your information: the `meta` tests work for `metatensor`, but not `metatensor-torch`

Luthaf:
> For your information: the meta tests work for metatensor, but not metatensor-torch

That's kinda expected: `Labels.to(device)` does nothing in `metatensor`

Luthaf:
Ok, this is proving a lot harder than expected.

Overriding `to` is not an option, since we can not use it the new version from C++. Only parameters and buffers are moved from C++.

The issue with the `register_labels_as_buffers` idea above is that after moving to a new device, the buffer no longer share data with the labels, so while the buffer is moved to the device by `module.to(device)` or `module.cuda()`, the labels are not.

I then tried to add  pre-forward hook that would automatically move the data if the buffer has been moved, but these hooks require to statically know the full set of parameters of `forward`, which we can't do inside a generic `register_labels_as_buffers`.

I'm reaching the conclusion that we will need to move the labels/blocks/tensormap inside forward, adding a check if the labels device/dtype match the input device/dtype and moving if it does not. This is not ideal since we need to explicitly do it for each new custom Module, but it seems like the only solution.

frostedoyster:
Ok, fair. I'll be working on it

frostedoyster (approved):

python/metatensor-learn/tests/module_map.py
```diff
@@ -87,3 +93,126 @@ def test_module_map_single_block_tensor(single_block_tensor_torch, out_propertie
...
+    assert tensor_module._in_keys.device == "cpu"
+    if out_properties is not None:
+        for label in tensor_module._out_properties:
+            assert label.device == "cpu"
+
+    tensor_module.to("meta")
+
+    # at this point, the parameters should have been moved,
+    # but the input keys and output properties should still be on cpu
+    for parameter in tensor_module.parameters():
```
Luthaf, line 130:
```suggestion
    assert len(list(tensor_module.parameters())) > 0
    for parameter in tensor_module.parameters():
```

python/metatensor-learn/tests/module_map.py
```diff
@@ -87,3 +93,126 @@ def test_module_map_single_block_tensor(single_block_tensor_torch, out_propertie
...
+                out_features=5,
+            )
+        )
+
+    tensor_module = ModuleMap(single_block_tensor_torch.keys, modules)
+    tensor_module(single_block_tensor_torch)
+
+    tensor_module = tensor_module.to(torch.float64)
+    single_block_tensor_torch = single_block_tensor_torch.to(dtype=torch.float64)
+    tensor_module(single_block_tensor_torch)
```
Luthaf, line 218: please add an assert that the output has the right dtype before and after the `to` call

python/metatensor-learn/tests/module_map.py
```diff
@@ -87,3 +93,126 @@ def test_module_map_single_block_tensor(single_block_tensor_torch, out_propertie
...
+
+
+@pytest.mark.parametrize(
+    "out_properties", [None, [Labels(["a", "b"], np.array([[1, 1]]))]]
+)
+@pytest.mark.skipif(not HAS_CUDA, reason="requires cuda")
+@pytest.mark.filterwarnings(
+    "ignore:.*If you are using Pytorch and need the labels to also be on cuda.*"
+)
+def test_module_map_cuda(single_block_tensor_torch, out_properties):  # noqa F811
```
Luthaf, line 157: should this test and the meta one be merged in a single one? It seems like duplicated code

python/metatensor-learn/tests/module_map.py
```diff
@@ -87,3 +93,126 @@ def test_module_map_single_block_tensor(single_block_tensor_torch, out_propertie
...
+    assert tensor_module._in_keys.device == "cpu"
+    if out_properties is not None:
+        for label in tensor_module._out_properties:
+            assert label.device == "cpu"
+
+    tensor_module.to("cuda")
+
+    # at this point, the parameters should have been moved,
+    # but the input keys and output properties should still be on cpu
+    for parameter in tensor_module.parameters():
```
Luthaf, line 183:
```suggestion
    assert len(list(tensor_module.parameters())) > 0
    for parameter in tensor_module.parameters():
```

python/metatensor-torch/tests/learn/module_map.py
```diff
@@ -84,6 +93,75 @@ def test_module_map_single_block_tensor(single_block_tensor, out_properties):
...
+    assert tensor_module._in_keys.device.type == "cpu"
+    if out_properties is not None:
+        for label in tensor_module._out_properties:
+            assert label.device.type == "cpu"
+
+    tensor_module.to("cuda")
+
+    # at this point, the parameters should have been moved,
+    # but the input keys and output properties should still be on cpu
+    for parameter in tensor_module.parameters():
```
Luthaf, line 128:
```suggestion
    assert len(list(tensor_module.parameters())) > 0
    for parameter in tensor_module.parameters():
```

python/metatensor-torch/tests/learn/module_map.py
```diff
@@ -84,6 +93,75 @@ def test_module_map_single_block_tensor(single_block_tensor, out_properties):
...
+    assert tensor_module._in_keys.device.type == "cpu"
+    if out_properties is not None:
+        for label in tensor_module._out_properties:
+            assert label.device.type == "cpu"
+
+    single_block_tensor = single_block_tensor.to(device="cuda")
+    out_tensor = tensor_module(single_block_tensor)
+    assert out_tensor.device.type == "cuda"
+
+    # the input keys and output properties should be on cuda
```
Luthaf, line 140:
```suggestion
    # the input keys and output properties should now be on cuda
```

python/metatensor-torch/tests/learn/module_map.py
```diff
@@ -84,6 +93,75 @@ def test_module_map_single_block_tensor(single_block_tensor, out_properties):
...
+                out_features=5,
+            )
+        )
+
+    tensor_module = ModuleMap(single_block_tensor.keys, modules)
+    tensor_module(single_block_tensor)
+
+    tensor_module = tensor_module.to(torch.float64)
+    single_block_tensor = single_block_tensor.to(torch.float64)
+    tensor_module(single_block_tensor)
```
Luthaf, line 162: same comment about asserts for the dtype of the ouput

python/metatensor-torch/tests/learn/module_map.py
```diff
@@ -111,3 +189,49 @@ def test_torchscript_module_map(single_block_tensor):
     buffer.seek(0)
     torch.jit.load(buffer)
     buffer.close()
+
+
+@pytest.mark.parametrize(
+    "out_properties", [None, [Labels(["a", "b"], torch.tensor([[1, 1]]))]]
+)
+@pytest.mark.skipif(not HAS_CUDA, reason="requires cuda")
+def test_torchscript_module_map_cuda(out_properties, single_block_tensor):
```
Luthaf, line 198: can this test be merged with the other one?

python/metatensor-torch/tests/learn/module_map.py
```diff
@@ -111,3 +189,49 @@ def test_torchscript_module_map(single_block_tensor):
     buffer.seek(0)
     torch.jit.load(buffer)
     buffer.close()
+
+
+@pytest.mark.parametrize(
```
Luthaf, line 194: I'm not sure it is worth parametrizing all these tests on `out_properties`. Just one version with `out_properties` defined to something should be plenty to check this part of the code.

Luthaf (commented):
Looks good overall, I have a couple of improvements suggestions for the tests and it should be good to go!

Luthaf:
Closing in favor of #595

Issue #482, opened 2024-02-02
Speeding up dataset and dataloader


frostedoyster:
I'm wondering if it's possible to torchscript the indexing and collating in the Datasets. It made a huge difference to me: from super-slow training to relatively fast
@jwa7 @agoscinski @abmazitov @Luthaf

jwa7:
Sounds good! Do you have example code that is half ready to go? If so we can start there and iterate together :)

frostedoyster:
I only have the code in metatensor-models.
https://github.com/lab-cosmo/metatensor-models/blob/df96bb062ddf9e7ee2ff6ed620f35e788487ec6d/src/metatensor/models/utils/data/dataset.py#L20-L21
We could aim to just compile these functions. I fear compiling the whole index functions within the Datasets and the collate functions would be much harder

ceriottm:
Can you profile to figure out what is making the current implementation that slow?

frostedoyster:
Completely in RAM

frostedoyster:
https://gist.github.com/frostedoyster/eb82bdf463c05cfd04cb6b652940d624
https://gist.github.com/frostedoyster/27a1d45eb350e0baf51e62d34dd26463

With these simple examples, I get 5x on `join` and 30x on `slice` by compiling. I'll try if I can profile

frostedoyster:
I don't know if join can be sped up without compiling. Regarding slice,

```
all_samples = block.samples.view(labels.names).to_owned()
# create an arrays of bools indicating which samples indices to keep
_, position, _ = all_samples.intersection_and_mapping(labels)
samples_mask = _dispatch.where(position != -1, True, False)
```

This would work in principle to substitute the list above, but `to_owned()` fails because some entries can be repeated in the view, and owned labels can't have the same entry multiple times. For now, I think the easiest option is to compile

Luthaf:
I'm surprised by how much of a different torch script makes. From my understanding, it should mainly be replacing the overhead of running the Python, interpreter with the overhead of running the TorchScript interpreter. Even if TorchScript is a lot simpler than Python, that's still quite a large difference!

> This would work in principle to substitute the slow list above, but to_owned() fails because some entries can be repeated in the view, and owned labels can't have the same entry multiple times. For now, I think the easiest option is to compile

We could make it so the right hand side of `intersection_and_mapping` does not need to be owned, only the left hand  side one. But this might require fairly intrusive changes so I'd go for compiling anyway in the mean time!

[...]

PR #504, opened 2024-02-09
Automatically use torch.jit.script for all operations

Luthaf:
This should help with #482, and torch script compatibility was already a requirement anyway.

The diff is a bit larger than I'd like because when using 

```
@torch.jit.script
def foo():
    ...
```

all functions called inside `foo` must already be defined.

As a small side change, I renamed `_classes` to `_backend` which feels more descriptive.

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--504.org.readthedocs.build/en/504/


frostedoyster (commented):
Is there a way to use the non-scripted classes? (E.g. for better error messages?)
EDIT: E.g. "re-run with METATENSOR_OPERATIONS_TORCHSCRIPT=0 for a better error message"

agoscinski:
```
Is there a way to use the non-scripted classes? (E.g. for better error messages?)
EDIT: E.g. "re-run with METATENSOR_OPERATIONS_TORCHSCRIPT=0 for a better error message
```
Maybe this can be put into CONTRIBUTING.rst into the `Running tests`? Otherwise looks good. PR seems very simple, just creates a lot of line noise because of change of order of the functions which is needed for torch script

python/metatensor-operations/metatensor/operations/_backend.py
```diff
@@ -16,12 +16,17 @@ def torch_jit_is_scripting():
     return False
 
 
+def torch_jit_script(function):
+    return function
```
PicoCentauri, line 20: Maybe we should a small comment why we need this function. On a first glance this just looks like a useless unity operation.

python/metatensor-operations/metatensor/operations/reduce_over_samples.py
```diff
@@ -437,12 +428,15 @@ def sum_over_samples_block(
     [[ 4  7 10]
      [17 19 21]]
     """
+    if isinstance(sample_names, str):
+        sample_names = [sample_names]
```
PicoCentauri, line 432: This conversion done in the private functions? I think we can skip this here or do we need this due to torchscript?

PicoCentauri (commented):
Two minor questions. Very good test cleanup as well.

Luthaf:
> Is there a way to use the non-scripted classes? (E.g. for better error messages?)

I think we can use the normal PyTorch mechanism for this: https://pytorch.org/docs/stable/jit.html#disable-jit-for-debugging

I'll add a line to Contributing.rst

python/metatensor-operations/metatensor/operations/_backend.py
Luthaf: Good point

python/metatensor-operations/metatensor/operations/reduce_over_samples.py
Luthaf: It is not: the private function want `List[str]`, not `Union[List[str], str]`. I can change them to do this though!

PicoCentauri: Adding this to the private functions looks for consistent to me. The public functions are just as wrappers for the private ones setting the parameters to the private ones. Sorry for the extra work.

frostedoyster (approved):

PR #511, opened 2024-02-12
Share `sphinx_gallery_conf` within documentation building

PicoCentauri:
I also added some explanations why we are using two sphinx projects. Let me know if this makes things a bit clearer.

I had to rename the folder `scripts/sphinx-gallery` to `gallery` to not infer with the actual sphinx gallery project. I am not sure about the name though.

----
📚 Documentation preview 📚: https://metatensor--511.org.readthedocs.build/en/511/

### DIFF
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -4,7 +4,6 @@ import sys
 from datetime import datetime
 
 from sphinx.domains.c import CObject
-from sphinx_gallery.sorting import FileNameSortKey
 
 
 # When importing metatensor-torch, this will change the definition of the classes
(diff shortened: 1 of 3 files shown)


docs/src/conf.py
```diff
@@ -17,6 +16,17 @@
 
 ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", ".."))
 sys.path.append(os.path.join(ROOT, "docs", "extensions"))
+sys.path.append(os.path.join(ROOT, "docs", "scripts"))
+
+
+# We use a second (pseudo) sphinx project located in `docs/scipt/gallery` to run the
+# examples and generate the actual outout for our shinx-gallery. This is necessary
+# because here we have to set `METATENSOR_IMPORT_FOR_SPHINX` to TRUE to allow the
```
Luthaf, line 24:
```suggestion
# because here we have to set `METATENSOR_IMPORT_FOR_SPHINX` to`"1"` to allow the
```

docs/src/conf.py
```diff
@@ -17,6 +16,17 @@
...
+# correct generation of the class and function docstrings which are seperate from the
+# actual code.
+#
+# We the same sphinx gallery confoguartion as in pseudo project
+from gallery.conf import sphinx_gallery_conf  # noqa: E402
```
Luthaf, line 29: I'd put `sys.path.append(os.path.join(ROOT, "docs", "scripts"))` closer to this line, to indicate they are related

Luthaf (commented):
> I am not sure about the name though.

How about somthing like `execute_examples` / `generate_example` / `run_examples` / ... ? 

We could also move the code a bit to have `<metatensor>/docs/run_examples/{conf.py, generate-examples.py}` and remove one level of nesting

PicoCentauri:
Okay I think we won against RTD for now again :-)

Luthaf (approved):

PR #513, opened 2024-02-13
More `metatensor.learn.nn` building blocks

jwa7:
**Description**

More `metatensor.learn.nn` building blocks:

- `LayerNorm`: applies a layer norm to all blocks
- `InvariantLayerNorm`: applies a layer norm only to invariant blocks at specified key indices. Applies the identity operation to all others, thus maintaining equivariance.
- `EquiLinear`: specification of `bias` to only be applied to invariant blocks
- `Tanh`: tanh transformations on each block
- `InvariantTanh`: only on invariant blocks
- `Sequential`: build sequential models, accepting any `metatensor.learn.nn.ModuleMap` as args.

Docstrings in place and unit tests to come, but feedback welcome in the meantime @agoscinski @abmazitov @Luthaf 

**Usage**

```py
import metatensor
from metatensor.learn import nn as nn

X = metatensor.load("path/to/X.npz")
Y = metatensor.load("path/to/Y.npz")

keys = X.keys
in_features = [len(X[key].properties) for key in keys]
out_features = [len(Y[key].properties) for key in keys]
invariant_key_idxs = [i for i, key in enumerate(keys) if key["spherical_harmonics_l"] == 0]

model = nn.Sequential(
        in_keys,
        nn.LayerNorm(
            in_keys=in_keys,
            normalized_shape=[
                len(tensor.block(key).properties) for key in invariant_key_idxs
            ],
            dtype=torch.float64,
        ),
        nn.Linear(
            in_keys=in_keys,
            in_features=in_features,
            out_features=4,
            bias=True,
            dtype=torch.float64,
        ),
        nn.Tanh(in_keys=in_keys),
        nn.Linear(
            in_keys=in_keys,
            in_features=4,
            out_features=1,
            out_properties=[Y.block(key).properties for key in in_keys],
            bias=True,
            dtype=torch.float64,
        ),
    )

model(X)
>>> TensorMap with 8 blocks
keys: spherical_harmonics_l  species_center
                0                  1
                1                  1
                      ...
                2                  14
                3                  14
```

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--513.org.readthedocs.build/en/513/


python/metatensor-learn/metatensor/learn/nn/layer_norm.py
```diff
@@ -0,0 +1,321 @@
...
+            raise ValueError(
+                "`bias` must have same length as `in_keys`, but"
+                f" len(bias) != len(in_keys) [{len(bias)} !="
+                f" {len(in_keys)}]"
+            )
+
+        # Build module list
+        modules = []
+        for i in range(len(in_keys)):
+            module = torch.nn.LayerNorm(
```
abmazitov, line 148: Shouldn't we pass the `elementwise_affine`, `bias`, `device` and `dtype` to the `torch.nn.LayerNorm` constructor?

jwa7: Good spot, thanks!

ceriottm:
As I mentioned today, I think with similar effort you could implement an EquiLayerNorm that also works on non-invariant layers. The only important thing is that the same scaling is applied to all `components` for one property. I think it'd be worth doing this in this PR - let me know if I'm missing something or you want to discuss this.

python/metatensor-learn/metatensor/learn/nn/layer_norm.py
```diff
@@ -0,0 +1,324 @@
...
+            )
+            modules.append(module)
+
+        super().__init__(in_keys, modules, out_properties)
+
+
+class EquiLayerNorm(ModuleMap):
+    """
+    Construct a module map with only layer norm modules applied to the invariant
+    blocks, and the identity applied to covariant blocks.
```
agoscinski, line 164: Would add a reference, because some stuff is not clear without the torch doc for example `elementwise_affine`

```suggestion
    blocks, and the identity applied to covariant blocks. Please refer to the
    `torch.nn.LayerNorm documentation <https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html>`_ 
    for a more detailed explanation of the parameters
```

python/metatensor-learn/metatensor/learn/nn/layer_norm.py
```diff
@@ -0,0 +1,324 @@
...
+        properties are relabeled using Labels.range.
+    """
+
+    def __init__(
+        self,
+        in_keys: Labels,
+        normalized_shape: Union[
+            Union[int, List[int], torch.Size],
+            Union[List[int], List[List[int]], List[torch.Size]],
+        ],
```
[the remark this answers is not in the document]
Luthaf: this type feels extremely complex. Why is it needed? Also, I'm fairly sure that `Union[Union[T, U], Union[V, W]]` is exactly the same as `Union[T, U, V, W]` anyway.

python/metatensor-learn/metatensor/learn/nn/layer_norm.py
Luthaf: Please use 
```
:py:class`torch.nn.LayerNorm`
```

To generate the link, this way we won't have issues if torch changes its URL scheme.

Luthaf (commented):
this seems to be missing some tests that the code can be used with TorchScript

python/metatensor-learn/tests/sequential.py
```diff
@@ -0,0 +1,114 @@
...
+        ),
+    )
+
+    prediction = model(tensor)
+    assert metatensor.equal_metadata(
+        prediction, tensor, check=["samples", "components"]
+    )
+
+
+def test_sequential_equi_mlp(tensor, wigner_d_real):
```
Luthaf, line 65: this does not feel relevant here: Sequential has no knowledge of equivariance

python/metatensor-learn/tests/sequential.py
```diff
@@ -0,0 +1,114 @@
...
+            in_keys=in_keys,
+            in_features=4,
+            out_features=1,
+            bias=True,
+            dtype=torch.float64,
+        ),
+    )
+
+    prediction = model(tensor)
+    assert metatensor.equal_metadata(
```
Luthaf, line 60: I think this should also check that the values are the same, at least for one block.

tox.ini
```diff
@@ -136,6 +136,8 @@ deps =
     {[testenv]packaging_deps}
     {[testenv]testing_deps}
     torch=={env:METATENSOR_TESTS_TORCH_VERSION:2.2.*}
+    scipy
```
Luthaf, line 139:
```suggestion
    
    ; used for checking equivariance
    scipy
```

python/metatensor-learn/metatensor/learn/nn/layer_norm.py
```diff
@@ -0,0 +1,361 @@
...
+            self.weight = Parameter(torch.empty(in_features, **factory_kwargs))
+            if bias:
+                self.bias = Parameter(torch.empty(in_features, **factory_kwargs))
+            else:
+                self.register_parameter("bias", None)
+        else:
+            self.register_parameter("weight", None)
+            self.register_parameter("bias", None)
+
+        self.reset_parameters()
```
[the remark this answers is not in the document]
jwa7: Not entirely sure, but wanted to keep the base implementation as similar to the torch native as possible - i.e. copied and pasted, but minimally modified

python/metatensor-learn/metatensor/learn/nn/layer_norm.py
```diff
@@ -0,0 +1,361 @@
...
+    :param mean: bool or list of bool. Whether or not to subtract the mean over the
+        properties (final) dimension of each invariant block of the input passed to
+        :py:meth:`forward`. If passed as a list, must have the same length as :param
+        invariant_key_idxs:.
+    """
+
+    def __init__(
+        self,
+        in_keys: Labels,
+        invariant_key_idxs: List[int],
```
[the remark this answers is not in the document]
jwa7: Resolved without change after discussion

python/metatensor-learn/tests/sequential.py
jwa7: Check what values are the same sorry? I have no reference data here, should I write a regression test?

python/metatensor-learn/tests/_rotation_utils.py
```diff
@@ -0,0 +1,208 @@
+"""
+Class for generating real Wigner-D matrices, and using them to rotate ASE frames
+and TensorMaps of density coefficients in the spherical basis.
```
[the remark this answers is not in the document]
jwa7: I think there's a difference in the Euler angle convention used to do the rotations between ASE (XZX) and scipy (ZYZ) and I didn't get round to a conversion between them so hadn't cross checked. The equivariance checks and the norm preservation seem fine, though.

python/metatensor-learn/metatensor/learn/nn/linear.py
```diff
@@ -3,121 +3,198 @@
...
+                device=device,
+                dtype=dtype,
             )
             modules.append(module)
 
         super().__init__(in_keys, modules, out_properties)
 
     @classmethod
-    def from_weights(cls, weights: TensorMap, bias: Optional[TensorMap] = None):
+    def from_weights(
```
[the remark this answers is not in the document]
jwa7: `from_weights` removed re: discussion. See #585

python/metatensor-torch/tests/learn/tanh.py
```diff
@@ -0,0 +1,57 @@
+import pytest
+import torch
+
+import metatensor.torch
+from metatensor.torch.learn.nn import InvariantTanh, Tanh
+
+from .._tests_utils import _test_module_save_load, _test_scripted_module_output
+from ._tests_utils import random_single_block_no_components_tensor_map
```
Luthaf, line 8: I don't love this =/ I would rather have some code duplication in tests but make them extremely easy to read & as simple as possible. Do you have a strong feeling here?

python/metatensor-learn/tests/sequential.py
Luthaf: I meant writing the same sequential for normal torch tensor & sending a single block through. But with random initialization, this might be a lot of complexity and not much advantages, so we can maybe do without

python/metatensor-torch/tests/learn/tanh.py
Luthaf: I'm talking about `_test_module_save_load` and `_test_scripted_module_output`, not `random_single_block_no_components_tensor_map` (for now)

jwa7: When I was writing and editing the tests (i.e. changing the io buffer context manager) it was rather inconvenient to change all the tests and do copy-paste-editing of all the variable names etc. As such it felt appropriate to remove redundancy in this way. 

I'd argue that these two tests are simple enough that one can easily understand what they do by looking at them once in `.._tests_utils`, and so slight loss in readability makes future developments and edits easier. Also the reason they are in `.._tests_utils` is that they are the same test for all learn modules, but also in principle for all operations too, and could be applied here as well. Thoughts?

Luthaf (approved):
I cleaned up the set of commits, this should be good to go!

PR #533, opened 2024-02-29
Move unit validation and conversion to C++

Luthaf:
This allow to check for valid unit on metadata construction time instead of later

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--533.org.readthedocs.build/en/533/


PicoCentauri (commented):
Nice! Maybe you can add a test for an error raise if one of the modules with wrong units or quantities is initialized. This is what the rewrite started.

Maybe unrelated to this PR: how to we define models with infinite cutoff? "cutoff=-1"?

Luthaf:
> Maybe unrelated to this PR: how to we define models with infinite cutoff? "cutoff=-1"?

`interaction_range = infinity`

PicoCentauri:
> > Maybe unrelated to this PR: how to we define models with infinite cutoff? "cutoff=-1"?
> 
> `interaction_range = infinity`

as a string? `interaction_range = "infinity"`

Luthaf:
no, this is a float, that you can set in the `ModelCapabilities`: https://lab-cosmo.github.io/metatensor/latest/atomistic/reference/models.html#metatensor.torch.atomistic.ModelCapabilities.interaction_range

metatensor-torch/src/atomistic/model.cpp
```diff
@@ -731,3 +759,117 @@ torch::jit::Module metatensor_torch::load_atomistic_model(
...
+    {"energy", Quantity{/* name */ "energy", /* baseline */ "ev", {
+        {"ev", 1.0},
+        {"mev", 1000.0},
+        {"hartree", 0.03674932247495664},
+        {"kcal/mol", 23.060548012069496},
+        {"kj/mol", 96.48533288249877},
+    }}},
+};
+
+void metatensor_torch::validate_quantity(const std::string& quantity) {
```
PicoCentauri, line 839: Am I reading this correctly that either a quantity is known or empty string, otherwise we throw an error? This means if I am doing `"dipole"` is not suppported anymore. I like it like this.

We should adjust the docstring of `metatensor.torch.atomistic.ModelOutput` for quantity and point to a list with supported quantities and units. Maybe create a page at `docs/src/atomistic` ?

PicoCentauri:
like `torch.inf`?

Luthaf:
or just `float("inf")`, or `1.0 / 0.0`, or `math.inf` ...

metatensor-torch/src/atomistic/model.cpp
Luthaf: yes, that's how it works for now. I'm not sure I love it, since it might prevent people to add units to their models if they are not known (and they might not come back & add units later). Not sure what the best solution is here. Maybe only sending a warning for unknown quantity (but error on unknown unit) would be better?

> point to a list with supported quantities and units.

It is documented in `unit_conversion_factor`: https://metatensor--533.org.readthedocs.build/en/533/atomistic/reference/models/index.html#metatensor.torch.atomistic.unit_conversion_factor

I could move it out of the function and link to it from `ModelOutput` and friends.

PicoCentauri: > Maybe only sending a warning for unknown quantity (but error on unknown unit) would be better?

Yes this is what I had in mind when I did the checks for `metatensor-models`

> It is documented in unit_conversion_factor: https://metatensor--533.org.readthedocs.build/en/533/atomistic/reference/models/index.html#metatensor.torch.atomistic.unit_conversion_factor


Ahh I found it. Maybe moving it out in propper clean table is better. I think this is quite important that we document what quantities we support.

Luthaf: Done!

PR #537, opened 2024-03-04
use new type convention in System error message

PicoCentauri:

 - ~[ ] Documentation updated (for new features)?~
 - ~[ ] Issue referenced (for PRs that solve an issue)?~

# Reviewer checklist


metatensor-torch/src/atomistic/system.cpp
```diff
@@ -342,17 +342,17 @@ SystemHolder::SystemHolder(torch::Tensor types, torch::Tensor positions, torch::
...
             "`positions` must be a 2 dimensional tensor, got a tensor with " +
             std::to_string(positions_.sizes().size()) + " dimensions"
         );
     }
 
-    if (positions_.size(0) != n_atoms || positions_.size(1) != 3) {
+    if (positions_.size(0) != n_types || positions_.size(1) != 3) {
         C10_THROW_ERROR(ValueError,
-            "`positions` must be a (n_atoms x 3) tensor, got a tensor with shape [" +
+            "`positions` must be a (n_types x 3) tensor, got a tensor with shape [" +
```
Luthaf, line 355: `n_types` is a bit ambiguous in the error message: is this the number of different types or the shape of the `types` array? The previous `n_atoms` was a bit clearer to me, even if how we determine `n_atoms` was not extremely clear. How about
```suggestion
            "`positions` must be a (len(types) x 3) tensor, got a tensor with shape [" +
```

PicoCentauri: Sounds good!

Luthaf (approved):

PR #551, opened 2024-03-13
Add ModelCapabilities::dtype, picked by the model

Luthaf:
The dtype is picked by the model, and the engine has to convert data on input/output to whatever it needs (similar to how different devices are handled).

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--551.org.readthedocs.build/en/551/

### DIFF
--- a/metatensor-torch/CHANGELOG.md
+++ b/metatensor-torch/CHANGELOG.md
@@ -17,6 +17,21 @@ a changelog](https://keepachangelog.com/en/1.1.0/) format. This project follows
 #### Removed
 -->
 
+### metatensor-torch C++
+
+#### Added
+
+- `ModelCapabilities::dtype`, used by the model to communicate the dtype it
+  wants to use for inputs and outputs.
+
+### metatensor-torch Python
+
+#### Added
+
+- `ModelCapabilities.dtype`, used by the model to communicate the dtype it
+  wants to use for inputs and outputs.
+
+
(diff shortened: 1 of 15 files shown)


Luthaf:
> I thought he plan was the model basically has to accept the dtype the model gets from the engine and try work with it as best as it can?

Err, I understood it the other way around ^_^. The model decide what it wants, and the engine must adapt around it (like for device).

This might be a bit different for metatensor-models, where the "engine" is also the one creating the model. But at this point capabilities don't exist yet.

PicoCentauri:
> > I thought he plan was the model basically has to accept the dtype the model gets from the engine and try work with it as best as it can?
> 
> Err, I understood it the other way around ^_^. The model decide what it wants, and the engine must adapt around it (like for device).

Okay, I really thought the other way around based on the discussion on slack

Screenshot 2024-03-18 at 16 38 08

But, I see that this might be hard for a model that use hardcoded `dtypes` like PET. So we can use the `dtype` as `capabilities `

> This might be a bit different for metatensor-models, where the "engine" is also the one creating the model. But at this point capabilities don't exist yet.

Yes, there the user picks the `dtype` and in the end we would inject this into the new `capabilities` that you have implemented.

Luthaf:
There was some misunderstanding with Marcel, because it seems that in jax models can do automatic casting if the weights don't match the inputs. Torch can not do that, hence why it makes sense to pick dtype once and stay with it.

PicoCentauri (approved):
Okay, I get it now! LGTM then.

PR #553, opened 2024-03-15
Adapt the code for Numpy v2.0

Luthaf:
With these changes, tests that do not require torch pass when using numpy 2.0.0b1. Tests that are based on torch will have to wait.

To prevent issues in the mean time, I also pinned dependencies to `numpy<2`

# Reviewer checklist


python/metatensor-core/metatensor/tensor.py
```diff
@@ -732,12 +732,19 @@ def _list_or_str_to_array_c_char(strings: Union[str, Sequence[str]]):
     return c_strings
 
 
+def _is_numpy_int(value):
+    return np.can_cast(value, np.int32, casting="same_kind")
+
+
```
jwa7, line 738: To me this looks slightly misleading - is being castable to a numpy int and actually being a numpy int the same thing, at least for what this is practically used for?

jwa7 (commented):
Looks good! Just one question

python/metatensor-core/metatensor/tensor.py
Luthaf: with how the function is used yes, but it's a fair point on the function by itself. I'll rename to `_can_cast_to_numpy_int()`

jwa7: Ok sounds good! Happy for you to merge once done

jwa7 (approved):

PR #557, opened 2024-03-25
Parallelize compilation with cmake

frostedoyster:
We use cmake's `--parallel` flag to speed up compilation for Python builds

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--557.org.readthedocs.build/en/557/

### DIFF
--- a/python/metatensor-core/setup.py
+++ b/python/metatensor-core/setup.py
@@ -102,7 +102,7 @@ class cmake_ext(build_ext):
             check=True,
         )
         subprocess.run(
-            ["cmake", "--build", build_dir, "--target", "install"],
+            ["cmake", "--build", build_dir, "--parallel", "--target", "install"],
             check=True,
         )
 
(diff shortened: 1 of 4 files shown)


python/metatensor-core/setup.py
```diff
@@ -102,7 +102,7 @@ def run(self):
             check=True,
         )
         subprocess.run(
-            ["cmake", "--build", build_dir, "--target", "install"],
+            ["cmake", "--build", build_dir, "--parallel", "--target", "install"],
```
Luthaf, line 105: this does not do anything, since we are not compiling any C++ code. Does not hurt to have it though!

Luthaf:
That's nice: CI for metatensor-torch Python tests went from 10min to 7min!

I'll see if we can do the same for some other tests

python/metatensor-core/setup.py
frostedoyster: Ah I assume it just calls the Rust compiler?

Luthaf (approved):

PR #569, opened 2024-04-04
Load TorchScript extensions used by a model

### LINKED ISSUE #512: Facilities to load torch extensions for a model
We have some code to save torch extensions when exporting an atomistic model with 

```py
wrapper = MetatensorAtomisticModel(...)

wrapper.export("exported-model.pt", collect_extensions="path/to/a/folder")
# stores all extensions in `path/to/a/folder`
```

We should also provide some facilities to load a folder containing all these extensions, from both Python and C++.

----

This should remove the need for something like 

> We have to import rascaline.torch even though it is not used explicitly in this tutorial. The SOAP-BPNN model contains compiled extensions and therefore the import is required.

in metatensor-models ASE tutorials: https://lab-cosmo.github.io/metatensor-models/latest/examples/ase/run_ase.html

Luthaf:
Fix #512 

This PR adds functionality to load the extensions a model uses (in practise defined as the extensions that where loaded when exporting a model) after the fact. This will enable using a model using rascaline without doing an `import rascaline.torch` before; or in LAMMPS.

An extensions loaded from `<site-package>/some/path.so` will be recorded as `some/path.so`, and we will try to load it from `<extensions_directory>/some/path.so`, `<current_directory>/some/path.so` and just as `path` (letting the OS try to find this lib, which enables stuff like `LD_LIBRARY_PATH` to work).

### Unresolved

Should we allow for multiple `extensions_directory`? So a user could add a Python's `<site-packages>` with the right set of packages to the list, allowing to use some newer versions of the extensions together with the files that where collected during the model export?

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--569.org.readthedocs.build/en/569/

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -100,7 +100,7 @@ jobs:
     name: ${{ matrix.name }} (torch v${{ matrix.torch-version }})
     strategy:
       matrix:
-        torch-version: ['1.11', '1.12', '1.13', '2.0', '2.1', '2.2']
+        torch-version: ['1.12', '1.13', '2.0', '2.1', '2.2']
         arch: ['arm64', 'x86_64']
         os: ['ubuntu-22.04', 'macos-11', 'macos-14', 'windows-2019']
         exclude:
@@ -110,7 +110,6 @@ jobs:
           # no arm64-windows build
           - {os: windows-2019, arch: arm64}
           # arm64-macos is only supported for torch >= 2.0
-          - {os: macos-14, arch: arm64, torch-version: '1.11'}
           - {os: macos-14, arch: arm64, torch-version: '1.12'}
           - {os: macos-14, arch: arm64, torch-version: '1.13'}
         include:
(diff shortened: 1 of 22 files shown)


.github/workflows/build-wheels.yml
```diff
@@ -100,7 +100,7 @@ jobs:
     name: ${{ matrix.name }} (torch v${{ matrix.torch-version }})
     strategy:
       matrix:
-        torch-version: ['1.11', '1.12', '1.13', '2.0', '2.1', '2.2']
```
PicoCentauri, line 103: I the minimal torch version listed somewhere in the docs and/or in the `pyproject.toml`?

python/metatensor-torch/tests/atomistic/ase_calculator.py
```diff
@@ -27,162 +22,36 @@
 }
 
 
-class LennardJones(torch.nn.Module):
```
PicoCentauri, line 30: Is this intended to be removed?

Luthaf: yes! This code now lives in https://github.com/Luthaf/metatensor-lj-test/, which is installed by tox.

.github/workflows/build-wheels.yml
Luthaf: I don't think so, I can add it. Should I pull this commit into a separate PR?

PicoCentauri: Yes another PR is fine. If you want create an issue that we don't forget.

PicoCentauri (approved):

.github/workflows/build-wheels.yml
Luthaf: Too late, I've done it in this one =)

PR #570, opened 2024-04-04
More units for length and energy

Luthaf:
Add more known units for length and energy, including most of the (non-combined) units from https://docs.lammps.org/units.html

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--570.org.readthedocs.build/en/570/


frostedoyster:
A random question: why Hartree and not Ha?
I just tried to train on QM9 and "Ha" didn't work

Luthaf:
No real reason! I'm trying to limit how many variations of the same unit we have. Right now `nm` and `nanometer` is the only duplicated unit, I could even remove that. 

From a user point of view, trying Ha and getting an error telling to use Hartree should not be too bad, right?

Luthaf:
Actually, I changed my mind. It will be more useful (in particular for integration with MD engines) to support a larger set of units (for example plumed uses `nm`, `A`, `Bohr`). So I've added a bunch of alternative unit spelling, and we can add more to it later.

Luthaf:
> A random question: why Hartree and not Ha?

Is `Ha` an abbreviation used anywhere? I could not find it online.

frostedoyster:
You're right. It seems to be E_h everywhere

frostedoyster (approved):

PR #564, opened 2024-04-02
Mark metatensor-torch wheels as python-independant

Luthaf:
Since we do not directly link to libpython.so (only through libtorch_python), the wheels we produce do not depend on the version of Python used to create them.

This would be a first step toward distributing metatensor_torch wheels on PyPI, using a similar approach to https://github.com/lab-cosmo/sphericart/pull/106#issuecomment-2031600480

# Contributor (creator of pull-request) checklist

 - [ ] ~Tests updated (for new features and bugfixes)?~
 - [ ] ~Documentation updated (for new features)?~
 - [ ] ~Issue referenced (for PRs that solve an issue)?~

# Reviewer checklist

 - [ ] CHANGELOG updated with public API or any other important changes?

----
📚 Documentation preview 📚: https://metatensor--564.org.readthedocs.build/en/564/

PR #571, opened 2024-04-04
Build metatensor-torch for all supported torch versions

Luthaf:
This tries to build metatensor_torch wheels we can publish on PyPI, following #564.

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--571.org.readthedocs.build/en/571/

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -24,24 +24,24 @@ jobs:
           - name: x86_64 Linux
             os: ubuntu-22.04
             rust-target: x86_64-unknown-linux-gnu
-            cibw_arch: x86_64
-          - name: Aarch64 Linux
+            cibw-arch: x86_64
+          - name: arm64 Linux
             os: ubuntu-22.04
             rust-target: aarch64-unknown-linux-gnu
-            cibw_arch: aarch64
+            cibw-arch: aarch64
           - name: x86_64 macOS
-            os: macos-11
+            os: macos-12
             rust-target: x86_64-apple-darwin
-            cibw_arch: x86_64
-          - name: M1 macOS
+            cibw-arch: x86_64
+          - name: arm64 macOS
             os: macos-14
             rust-target: aarch64-apple-darwin
-            cibw_arch: arm64
+            cibw-arch: arm64
           - name: x86_64 Windows
             os: windows-2019
             # TODO: add a 32-bit windows builder?
             rust-target: x86_64-pc-windows-msvc
-            cibw_arch: AMD64
+            cibw-arch: AMD64
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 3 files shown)


Luthaf:
Ok, this seems to be working! Using the wheels from the latest build, 

```
$ pip install torch
$ pip install ./metatensor_core-0.2.0.dev35-py3-none-macosx_11_0_arm64.whl
$ pip install --no-deps ./metatensor_torch-0.4.0.dev32-py3-none-macosx_11_0_arm64.whl

$ python -c "import torch; import metatensor.torch; print(torch.__version__)"
2.2.2

$ pip install torch==2.1
$ python -c "import torch; import metatensor.torch; print(torch.__version__)"
2.1.0
```

The same metatensor_torch wheel can be loaded with multiple torch versions!

----

The files sizes are not horrible either:

```
$ ls -lah metatensor_torch*
-rw-r--r--@ 1 guillaume  staff   1,9M  5 avr 14:51 metatensor_torch-0.4.0.dev32-py3-none-macosx_11_0_arm64.whl
-rw-r--r--@ 1 guillaume  staff   4,1M  5 avr 14:51 metatensor_torch-0.4.0.dev32-py3-none-macosx_11_0_x86_64.whl
-rw-r--r--@ 1 guillaume  staff   3,4M  5 avr 14:51 metatensor_torch-0.4.0.dev32-py3-none-manylinux2014_aarch64.manylinux_2_17_aarch64.whl
-rw-r--r--@ 1 guillaume  staff   3,6M  5 avr 14:51 metatensor_torch-0.4.0.dev32-py3-none-manylinux2014_x86_64.manylinux_2_17_x86_64.whl
-rw-r--r--@ 1 guillaume  staff   2,5M  5 avr 14:51 metatensor_torch-0.4.0.dev32-py3-none-win_amd64.whl
```

----

I think the main impact is going to be on CI time. We went from ~30 CI jobs to 63 CI jobs; and building the arm64-linux wheels uses CPU emulation and is very slow (around ~20min to build the torch wheels). We could either

- not run this for all PR, only run it on release + on a schedule
- not build arm64 wheels for metatensor_torch (only metatensor_core). Users would fallback on the sdist and have to build from sources
- some other idea?

frostedoyster:
Does this work with torch + cuda?

Luthaf:
It should, since we are only linking to torch_cpu which should be around as well. Can you try quickly, using the right wheel from here: https://github.com/lab-cosmo/metatensor/actions/runs/8571334454/artifacts/1388709143?

Luthaf:
It seems that the wheels still contain a reference to metatensor-core by path (`/project/python/metatensor-core`), this should be fixed.

.github/workflows/build-wheels.yml
```diff
@@ -67,36 +67,192 @@ jobs:
...
-        if: startsWith(github.ref, 'refs/tags/metatensor-core-v')
-        uses: softprops/action-gh-release@v1
+  build-torch-wheels:
+    runs-on: ${{ matrix.os }}
+    name: ${{ matrix.name }} (torch v${{ matrix.torch-version }})
+    strategy:
+      matrix:
+        torch-version: ['1.11', '1.12', '1.13', '2.0', '2.1', '2.2']
+        arch: ['arm64', 'x86_64']
+        os: ['ubuntu-22.04', 'macos-11', 'macos-14', 'windows-2019']
```
PicoCentauri, line 100: Shouldn't there be another loop over python versions?

.github/workflows/build-wheels.yml
```diff
@@ -67,36 +67,192 @@ jobs:
...
+            arch: arm64
+            rust-target: aarch64-apple-darwin
+            cibw-arch: arm64
+          - name: x86_64 Windows
+            os: windows-2019
+            arch: x86_64
+            rust-target: x86_64-pc-windows-msvc
+            cibw-arch: AMD64
+          # add the right python version for each torch version
+          - {torch-version: '1.11', python-version: '3.10', cibw-python: 'cp310-*'}
```
PicoCentauri, line 139: Can't you run this with Py3.12? If yes does this work? I thought we have to build the wheels for every Python major version.

PicoCentauri (commented):
looks like it could work 🤣

.github/workflows/build-wheels.yml
Luthaf: no, we only need to build the code once per torch version thanks to https://github.com/lab-cosmo/metatensor/pull/564

We just pick one python version that's compatible with the torch version for the build below.

.github/workflows/build-wheels.yml
Luthaf: no, there are no pre-built torch 1.11 for python 3.12. The versions here are the highest Python versions for which there is a prebuilt torch wheel

.github/workflows/build-wheels.yml
PicoCentauri: I see.

.github/workflows/build-wheels.yml
PicoCentauri: Okay I understand.

PicoCentauri (approved):

PR #578, opened 2024-04-10
Allow to override custom data inside a System

Luthaf:
This enable workflows where a first part of the calculation generates some data that will be used as input in a second part of the calculations (e.g. soap model to generate charges then used by lode)

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--578.org.readthedocs.build/en/578/

### DIFF
--- a/python/metatensor-torch/metatensor/torch/atomistic/documentation.py
+++ b/python/metatensor-torch/metatensor/torch/atomistic/documentation.py
@@ -135,7 +135,7 @@ class System:
         Get all the neighbors lists options registered with this :py:class:`System`
         """
 
-    def add_data(self, name: str, data: TensorBlock):
+    def add_data(self, name: str, data: TensorBlock, override: bool = False):
         """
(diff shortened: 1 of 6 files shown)


python/metatensor-torch/metatensor/torch/atomistic/documentation.py
```diff
@@ -144,6 +144,8 @@ def add_data(self, name: str, data: TensorBlock):
 
         :param name: name of the custom data
         :param data: values of the custom data
+        :param override: if true, allow this function to override existing data with the
```
PicoCentauri, line 147: I don't know if we follow this but we could link to the Python docs here.

```suggestion
        :param override: if :py:obj:`True`, allow this function to override existing data with the
```

Luthaf: TBH, I'm not sure what we (or the users) gain by linking to the docs of `True`/`False`/`None`. Happy to put that inside backticks for a better rendering

PR #590, opened 2024-04-26
Add support for torch v2.3

Luthaf:
This is not doing much, just build wheels for torch v2.3 and using it for tests on CI

----
📚 Documentation preview 📚: https://metatensor--590.org.readthedocs.build/en/590/

### DIFF
--- a/python/metatensor-core/tests/_tests_utils.py
+++ b/python/metatensor-core/tests/_tests_utils.py
@@ -1,7 +1,6 @@
-"""Utility functions to be used across all test files.
+"""Utility functions to be used across all test files"""
 
-These functions are by design no pytest fixtures to avoid a confusing global import.
-"""
+import os
 
 import numpy as np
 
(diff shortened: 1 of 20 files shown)


python/metatensor-core/tests/_tests_utils.py
```diff
@@ -144,3 +143,15 @@ def tensor_zero_len_block():
             )
         ],
     )
+
+
+def can_use_mps_backend():
+    import torch
+
+    return (
+        # Github Actions M1 runners don't have a GPU accessible
```
PicoCentauri, line 152: This is a scandal TBH that they don't allow us to use the GPU. But I understand why.

Luthaf: I think this is actually because CI is running on VM and not on raw hardware, and the VM can't access the GPU in general.

PR #592, opened 2024-04-26
Record the compatible torch version in wheel METADATA

Luthaf:
We now set `install_requires`/`Requires-Dist:` fields to the specific set of torch version a given metatensor-torch wheel is compatible with. This happen both when building a wheel locally (we set `torch == x.y.*`) and when merging wheels on CI (we merge the individual requirements above into `torch >=1.12,<2.4`)

# Reviewer checklist

----
📚 Documentation preview 📚: https://metatensor--592.org.readthedocs.build/en/592/

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -359,9 +359,8 @@ jobs:
         uses: softprops/action-gh-release@v1
         with:
           files: |
-            wheels/metatensor-core-*.tar.gz
             wheels/cxx/metatensor-core-cxx-*.tar.gz
-            wheels/metatensor_core-*.whl
+            wheels/metatensor_core-*
           prerelease: ${{ contains(github.ref, '-rc') }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -371,9 +370,8 @@ jobs:
         uses: softprops/action-gh-release@v1
         with:
           files: |
-            wheels/metatensor-torch-*.tar.gz
             wheels/cxx/metatensor-torch-cxx-*.tar.gz
-            wheels/metatensor_torch-*.whl
+            wheels/metatensor_torch-*
           prerelease: ${{ contains(github.ref, '-rc') }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -383,8 +381,7 @@ jobs:
         uses: softprops/action-gh-release@v1
         with:
           files: |
-            wheels/metatensor-operations-*.tar.gz
-            wheels/metatensor_operations-*.whl
+            wheels/metatensor_operations-*
           prerelease: ${{ contains(github.ref, '-rc') }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


Luthaf:
This seems to be working, see for example https://github.com/lab-cosmo/metatensor/actions/runs/8879679437/job/24379229869?pr=592

The METADATA file contains 

```
[...]
Requires-Python: >=3.8
Description-Content-Type: text/x-rst
License-File: LICENSE
License-File: AUTHORS
Requires-Dist: torch >=2.0,<2.3
```

Luthaf:
With the wheel on test.pypi.org, this seems to work:

```
$ pip install --extra-index-url https://test.pypi.org/simple/ --pre metatensor-torch
Successfully installed metatensor-torch-0.5.0.dev14 torch-2.2.2
```

```
$ pip install -U torch
Installing collected packages: torch
  Attempting uninstall: torch
    Found existing installation: torch 2.2.2
    Uninstalling torch-2.2.2:
      Successfully uninstalled torch-2.2.2
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
metatensor-torch 0.5.0.dev14 requires torch<2.3,>=2.0, but you have torch 2.3.0 which is incompatible.
```

```
$ pip uninstall metatensor-torch
$ pip install --extra-index-url https://test.pypi.org/simple/ --pre metatensor-torch
Installing collected packages: torch, metatensor-torch
  Attempting uninstall: torch
    Found existing installation: torch 2.3.0
    Uninstalling torch-2.3.0:
      Successfully uninstalled torch-2.3.0

Successfully installed metatensor-torch-0.5.0.dev14 torch-2.2.2
```

We can then document the `--no-binary` option to install with another torch version in the documentation.

.github/workflows/build-wheels.yml
```diff
@@ -244,14 +247,24 @@ jobs:
...
+          # unpack all single torch versions wheels in the same directory
+          mkdir dist/unpacked
+          find dist -name "*.whl" -print -exec python -m wheel unpack --dest dist/unpacked/ {} ';'
 
-          mkdir unpacked
-          find . -name "*.whl" -print -exec python -m wheel unpack --dest unpacked/ {} ';'
+          sed -i "s/Requires-Dist: torch ==.*/$MERGED_TORCH_REQUIRE/" dist/unpacked/metatensor_torch-*/metatensor_torch-*.dist-info/METADATA
 
+          echo "\n\n METADATA = \n\n"
+          cat dist/unpacked/metatensor_torch-*/metatensor_torch-*.dist-info/METADATA
```
PicoCentauri, line 263: Do we want to keep thos debug information?

I think it is useful but just to be sure.

PicoCentauri (commented):
It looks like that this will work. I am afraid to ask but, is there a test that we could add to make sure we don't break this.

If torch has some builds from their `master` branch it maybe we can use these.
- Install the master branch, assuming that their version number is above the one we support
- install metatensor
- check that installation is succesfull and we have the latest torch version that we support.
-

.github/workflows/build-wheels.yml
Luthaf: yes, this is just printed in the CI so it might be worth keeping around.

Luthaf:
>  I am afraid to ask but, is there a test that we could add to make sure we don't break this.

I could add a test, but it would only run whenever the wheel build CI runs (so not on every PR)

Luthaf:
Actually, adding a test might not be possible, since we would need to install the version of metatensor-core that was created during the same CI run, but the requirements currently contain something like `Requires-Dist: metatensor-core @ file:///<...>/metatensor/python/metatensor-core` (because we want a local install to behave differently from an installation from PyPI). 

I'll try to think more about it.

PicoCentauri:
I think it may also be fine if we leave this like this for now. If we break this logic at some point we can come back and try to invent a test.

PR #616, opened 2024-05-13
Fix syntax in wheels comment workflow

Luthaf:
` is used for string formatting in JS.

This should fix the error in https://github.com/lab-cosmo/metatensor/actions/runs/9060508099


PicoCentauri (approved):
Ahh my bad. Thanks for the fix!

PR #618, opened 2024-05-13
Also set CMAKE_INSTALL_LIBDIR in metatensor-sys

Luthaf:
Same as #617 but when building the code to be used from Rust (i.e. rascaline)

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1497635853.zip)


Luthaf:
This will be released as v0.1.8, only on `crates.io` (the code would be identical everywhere else)

PR #623, opened 2024-05-17
Allow to access `wrapped_module` of exported models

frostedoyster:
This feature would be very useful to access information about the underlying model

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1760815821.zip)

### DIFF
--- a/python/metatensor-torch/metatensor/torch/atomistic/model.py
+++ b/python/metatensor-torch/metatensor/torch/atomistic/model.py
@@ -233,6 +233,14 @@ class MetatensorAtomisticModel(torch.nn.Module):
     >>> with tempfile.TemporaryDirectory() as directory:
     ...     wrapped.save(os.path.join(directory, "constant-energy-model.pt"))
     ...
+
+    .. py:attribute:: module
+        :type: ModelInterface
+
+    The torch module wrapped by this :py:class:`MetatensorAtomisticModel`.
+
+    Reading from this attribute is safe, but modifying it is not recommended,
+    unless you are familiar with the implementation of the model.
     """
 
     # Some annotation to make the TorchScript compiler happy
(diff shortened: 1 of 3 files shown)


Luthaf:
This does not work unfortunately, since `torch.nn.Module` is not a valid TorchScript type. Can you access `exported_model._module` after loading an exported model? If yes, we could rename it to `self.wrapped_module` or something similar, and document that reading this field is fine, but writing to it is not.

frostedoyster:
Yes, I was able to access `._module`. Let's do that then. Perhaps we should also remove `wrapped_module()` from the documentation though, since it's not accessible anymore once the model is exported

Luthaf:
Yes, that makes sense

Luthaf:
@frostedoyster what's the status on this? Do you still want to finish it or should we close?

docs/src/atomistic/reference/models/export.rst
```diff
@@ -16,3 +16,4 @@ define some information about the model, using :py:class:`ModelMetadata` and
 
 .. autoclass:: metatensor.torch.atomistic.MetatensorAtomisticModel
     :members:
+    :exclude-members: wrapped_module
```
Luthaf, line 19:
```suggestion
```

And did you check that `module` appears in the docs?

frostedoyster: No, and my way of including it also seems to break torchscript

Luthaf: instead of `ModelInterface`, you can try to annotate with `torch.nn.Module`

frostedoyster:
It also doesn't work. We could remove the annotation or try something more sophisticated (what though?)

Luthaf:
Does it work without any annotation?

Luthaf:
alternatively, you can document a member without going through a docstring, but instead adding something like

```rst
  <...> class docstring <...>
.. py:attribute:: module
    :type: ModelInterface
    
    <...> attribute docstring <...>
```

I'm not 100% sure on the syntax, I'm trying to understand https://www.sphinx-doc.org/en/master/usage/domains/python.html#directive-py-class

frostedoyster:
No, it doesn't work without the annotation (`module=None`)

docs/src/atomistic/reference/models/export.rst
Luthaf: This seems to work, could you add a test that we can access the module after a full export/save/load cycle? And remove the line above in the docs as well as the `wrappedmodule` function?

frostedoyster:
It passes the tests now but when I download the docs I see wrapped_module() and I don't see the new attribute

PR #625, opened 2024-05-17
Add `save()`, deprecate `.export()` for atomistic models

PicoCentauri:
As discussed today the name `export` for saving an atomistic model is a bit misleading. I changed this the better one called `save` and deprecate the old one.

# Reviewer checklist


Luthaf:
The CI failure seems relevant: we should update the tests to use `.save` as well

python/examples/atomistic/1-export-atomistic-model.py
```diff
@@ -215,17 +217,45 @@ def forward(
...
+# the API of the objects is the same!
+
+print("type wrapper:", type(wrapper))
+print("type atomistic_model:", type(atomistic_model))
+
+# %%
+#
+# Now that we have an exported model that we can load, the next tutorial will show how
+# you can use such a model to run `Molecular Dynamics`_ simulation using the Atomic
+# Simulating Environment (`ASE`_).
```
Luthaf, line 258: Users are not supposed to call the models like this, especially in the first tutorial.

python/metatensor-torch/metatensor/torch/atomistic/model.py
```diff
@@ -425,6 +425,24 @@ def forward(
...
+        .. warning::
+            :py:meth:`export` is deprecated. Use :py:meth:`save` instead.
+
+        :param file: where to save the model. This can be a path or a file-like object.
+        :param collect_extensions: if not None, all currently loaded PyTorch extension
+            will be collected in this directory. If this directory already exists, it
+            is removed and re-created.
+        """
+        warnings.warn(
+            message="`export()` method is deprecated\nUse `save()` method instead.",
```
Luthaf, line 437:
```suggestion
            message="`export()` is deprecated, use `save()` instead",
```

python/examples/atomistic/1-export-atomistic-model.py
PicoCentauri: Yeah probably you are right. Should we move it somewhere else? At least for me it these lines where helpful to understand what is going on.

PicoCentauri:
> The CI failure seems relevant: we should update the tests to use `.save` as well

Yes, I will look into this.

python/examples/atomistic/1-export-atomistic-model.py
Luthaf: I would put something like this in a tutorial for people writing a new "engine" that would call arbitrary models.

PicoCentauri:
Ahh okay the actual error is somewhere else. I see.

python/examples/atomistic/1-export-atomistic-model.py
Luthaf: No, we don't have one of these tutorial yet. I'm happy to fix the reshaping error, but I think it is more important for this tutorial to be clear and not confusing. If we can find a way to do this while also testing that the model actually runs, that would be ideal!

PicoCentauri: Agree. Adding a test might be a good idea. I will add one.

python/metatensor-torch/exported-model.pt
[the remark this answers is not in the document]
PicoCentauri: woopsiii

Luthaf (approved):

PR #629, opened 2024-05-23
Make `LayerNorm` sample independent

jwa7:
Fix the implementation of `LayerNorm` and `InvariantLayerNorm` to calculate the mean and variance of blocks over all dimensions except the samples, such that transformations are applied to samples independently. The current implementation calculates these statistics over the samples, thus applying a BatchNorm-esque transformation.

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1534277939.zip)

### DIFF
--- a/python/metatensor-learn/tests/layer_norm.py
+++ b/python/metatensor-learn/tests/layer_norm.py
@@ -83,3 +83,19 @@ def test_equivariance(tensor, wigner_d_real):
     fRx = f(Rx)  # f(R . x)
 
     assert metatensor.allclose(fRx, Rfx, atol=1e-10, rtol=1e-10)
+
+
+def test_layer_norm_indep_samples():
+    """
+    Tests that the LayerNorm is applied to each sample independently.
+    """
+    layer_norm_module = _LayerNorm(
+        in_features=10, elementwise_affine=True, dtype=torch.float64
+    )
+    tensor = torch.randn(5, 3, 10)
+    norm = layer_norm_module(tensor)
+
+    # Apply to each sample independently and check consistency
+    for i in range(tensor.shape[0]):
+        norm_i = layer_norm_module(tensor[i : i + 1])
+        assert torch.allclose(norm[i], norm_i, atol=1e-10, rtol=1e-10)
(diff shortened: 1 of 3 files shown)


python/metatensor-learn/tests/layer_norm.py
```diff
@@ -83,3 +83,19 @@ def test_equivariance(tensor, wigner_d_real):
     fRx = f(Rx)  # f(R . x)
 
     assert metatensor.allclose(fRx, Rfx, atol=1e-10, rtol=1e-10)
+
+
+def test_layer_norm_indep_samples():
```
Luthaf, line 88:
```suggestion
def test_layer_norm_independent_samples():
```

Luthaf (commented):
This looks alright, but it might be worth to get someone more familiar with LayerNorm to also check that this behaves as expected. Maybe @frostedoyster or @spozdn can help here?

jwa7:
> The change is an improvement over the previous design, but I think the resulting models will not be equivariant

Yes in this case `LayerNorm` is not equivariance-preserving, but none of the base modules (i.e. `Tanh`, `Linear`) are in general. The `InvariantLayerNorm` is however, by applying a `LayerNorm` to invariants only, and the identity operator to covariants (in a similar fashion to `InvariantTanh`). 

> I think the correct thing to do would be to apply the layernorm just to the properties

@Luthaf what are your thoughts on this re: our discussion last week? As implemented we are contracting over components and properties at the moment when calculating the mean/std. I seem to remember there being a reason for not contracting over just properties - what was it again?

frostedoyster:
Ah, nvm then, it looks good to me

Luthaf:
> As implemented we are contracting over components and properties at the moment when calculating the mean/std. I seem to remember there being a reason for not contracting over just properties - what was it again?

So my idea was that when we start using this for covariant LayerNorm, we will set the mean to 0 (or ignore the mean in normalization); and we need to divide all components by the same value to preserve equivariance. So we need to compute the statistics over all properties and all components, producing only one standard deviation per sample (and not one per component)

jwa7:
Makes sense - good as is then?

PR #632, opened 2024-05-24
Allow non-identifiers as fields in Dataset

### LINKED ISSUE #621: `metatensor.learn.Dataset` does not accept field which are not valid Python identifiers
In `metatensor.atomsitic`, we're requiring non-standard outputs to be `something::output_name`, but this is not compatible with `metatensor.learn.Dataset`, which works with named tuples.

Luthaf:
Fix #621 

We use a custom `namedtuple` function to allow storing data where the name is not a python identifier. These fields can be accessed by their numeric index `sample[3]` or using a string name `sample["some weird name"]`. Previous code should work in the exact same way.

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1543863592.zip)

### DIFF
--- a/python/metatensor-learn/metatensor/learn/data/dataset.py
+++ b/python/metatensor-learn/metatensor/learn/data/dataset.py
@@ -109,6 +109,38 @@ class Dataset(_BaseDataset):
     ID. This data field for a given sample is then only lazily loaded into memory when
     the :py:meth:`Dataset.__getitem__` method is called.
 
+    >>> # create a Dataset with only lists
+    >>> dataset = Dataset(num=[1, 2, 3], string=["a", "b", "c"])
+    >>> dataset[0]
+    Sample(num=1, string='a')
+    >>> dataset[2]
+    Sample(num=3, string='c')
+
+    >>> # create a Dataset with callables for lazy loading of data
+    >>> def call_me(sample_id: int):
+    ...     # this could also read from a file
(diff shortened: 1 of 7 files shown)


jwa7 (commented):
Looks good! And thanks for adding examples to the docstrings.

Do you think that we should mention the `from_dict()` method in the main class docstring(s), or do we rely on the user looking for the method in the docs if they want to use non-identifiers?

python/metatensor-learn/metatensor/learn/data/dataset.py
```diff
@@ -242,14 +288,70 @@ def __init__(self, sample_id: List, **kwargs):
...
         data = {"sample_id": sample_id}
         data.update(kwargs)
         super().__init__(data=data, size_arg_name="sample_id", size=len(sample_id))
 
         self._sample_id = sample_id
-        self._sample_id_to_idx = {smpl_id: idx for idx, smpl_id in enumerate(sample_id)}
+        self._sample_id_to_idx = {sample: i for i, sample in enumerate(sample_id)}
+
+    @classmethod
+    def from_dict(cls, data: Dict[str, Any], sample_id: List):
```
jwa7, line 301:
```py
def from_dict(cls, data: Dict[str, Any], sample_id: List) -> "IndexedDataset":
```

PR #640, opened 2024-05-29
added len of a block

DavideTisi:
add a __len__ function to the block.
I found it annoying not to be able to do len(block). 
It is not revolutionary so if you do not like you can also not merge it.

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1558537230.zip)

### DIFF
--- a/metatensor-torch/include/metatensor/torch/block.hpp
+++ b/metatensor-torch/include/metatensor/torch/block.hpp
@@ -59,7 +59,7 @@ public:
         return this->labels(0);
     }
 
-    // get the len of the samples
+    /// get the len of the samples
     int64_t len() const {
         return this->labels(0)->count();
     }
--- a/metatensor-torch/include/metatensor/torch/block.hpp
+++ b/metatensor-torch/include/metatensor/torch/block.hpp
@@ -61,7 +61,7 @@ public:
 
     // get the len of the samples
     int64_t len() const {
-        return this->labels(0).count();
+        return this->labels(0)->count();
     }
 
     /// Access the component `Labels` for this block.
--- a/metatensor-torch/include/metatensor/torch/block.hpp
+++ b/metatensor-torch/include/metatensor/torch/block.hpp
@@ -61,7 +61,7 @@ public:
 
     // get the len of the samples
     int64_t len() const {
-        return this->labels(0).size(0);
+        return this->labels.size(0);
     }
 
     /// Access the component `Labels` for this block.
--- a/metatensor-torch/include/metatensor/torch/block.hpp
+++ b/metatensor-torch/include/metatensor/torch/block.hpp
@@ -59,6 +59,11 @@ public:
         return this->labels(0);
     }
 
+    // get the len of the samples
+    int64_t len() const {
+        return this->labels(0).size(0);
+    }
+
     /// Access the component `Labels` for this block.
     ///
     /// The entries in these labels describe intermediate dimensions of the
--- a/metatensor-torch/include/metatensor/torch/block.hpp
+++ b/metatensor-torch/include/metatensor/torch/block.hpp
@@ -61,7 +61,7 @@ public:
 
     // get the len of the samples
     int64_t len() const {
-        return this->labels.size(0);
+        return this->labels(0).count();
     }
 
     /// Access the component `Labels` for this block.
--- a/metatensor-torch/include/metatensor/torch/block.hpp
+++ b/metatensor-torch/include/metatensor/torch/block.hpp
@@ -66,9 +66,7 @@ public:
 
     /// Get the shape of the values Tensor
     at::IntArrayRef shape() const{
-        auto _shape  = this->block_.values_shape();
-        std::vector<int64_t> shapei64(_shape.begin(), _shape.end());
-        at::IntArrayRef shape(shapei64);
+        at::IntArrayRef shape(this->shapei64);
         return shape ;
     }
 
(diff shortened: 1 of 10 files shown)


Luthaf:
Length of a block makes sense. As a related function, do we want to expose `block.shape` as well?

python/metatensor-core/metatensor/block.py
```diff
@@ -182,6 +182,12 @@ def __reduce__(self):
             "TensorMap first"
         )
 
+    def __len__(self):
+        """
+        get len of the values array
```
Luthaf, line 187:
```suggestion
        Get the length of the values stored in this block (i.e. the number of samples in the block)
```

Luthaf (commented):
we need the same function to be available for the TorchScript backend. This means adding the function in C++ `TensorBlockHolder` and registering it with TorchScript.

DavideTisi:
> we need the same function to be available for the TorchScript backend. This means adding the function in C++ `TensorBlockHolder` and registering it with TorchScript.

ok but why this is not done for `__len__` in tensorMap ?

Luthaf:
It is also done for `TensorMap.__len__`: https://github.com/lab-cosmo/metatensor/blob/bf8c0130961b61d9a0dd10c676cf8570f611cc57/metatensor-torch/src/register.cpp#L213

Although this is only exposed to TorchScript and not to C++. We could change this!

DavideTisi:
yes i deleted my comment almost immediately when i saw that lines

DavideTisi:
now it should be done

metatensor-torch/include/metatensor/torch/block.hpp
```diff
@@ -59,11 +59,19 @@ class METATENSOR_TORCH_EXPORT TensorBlockHolder: public torch::CustomClassHolder
...
     }
 
-    /// get the len of the samples
+    /// Get the len of the samples
     int64_t len() const {
         return this->labels(0)->count();
     }
 
+    /// Get the shape of the values Tensor
+    torch::Tensor shape() const{
```
Luthaf, line 68: this should be returned as the same type as `torch::Tensor::sizes`, which I think is either `at::IntArrayRef` or `at::IntArray`.

metatensor-torch/include/metatensor/torch/block.hpp
```diff
@@ -59,11 +59,19 @@ class METATENSOR_TORCH_EXPORT TensorBlockHolder: public torch::CustomClassHolder
         return this->labels(0);
     }
 
-    /// get the len of the samples
+    /// Get the len of the samples
```
Luthaf, line 62:
```suggestion
    /// Get the length of this block, i.e. the number of samples
```

python/metatensor-core/metatensor/block.py
```diff
@@ -182,13 +182,22 @@ def __reduce__(self):
...
         (i.e. the number of samples in the block)
         """
         return len(self.values)
 
+    @property
+    def shape(self):
+        """
+        Get the shape of the values stored in this block
+        (i.e. the number of samples, of components, of properties
+        in the block)
```
Luthaf, line 197:
```suggestion
        Get the shape of the values  array in this block.
```

python/metatensor-core/tests/blocks.py
```diff
@@ -41,6 +41,14 @@ def block_components():
     )
 
 
+def test_len(block):
```
Luthaf, line 44: could you also add the same tests in metatensor-torch?

metatensor-torch/include/metatensor/torch/block.hpp
DavideTisi: which i cannot use, because it fails I think it is not TorchScriptable

DavideTisi:
the shape code was wrong 
`at::IntArrayRef` is a reference so it need to reference to some variable that do not goes out of scope. 
so I create a shape variable for the class

it should be correct now, at least the test works

metatensor-torch/include/metatensor/torch/block.hpp
Luthaf: Yes, but you can use `this->values().sizes()` instead. 

`values_shape()` is here for integration with metatensor (which uses `vector<size_t>`)

metatensor-torch/tests/block.cpp
```diff
@@ -1,4 +1,5 @@
 #include <torch/torch.h>
+#include <iostream>
```
Luthaf, line 2:
```suggestion
```

metatensor-torch/include/metatensor/torch/block.hpp
```diff
@@ -170,6 +181,8 @@ class METATENSOR_TORCH_EXPORT TensorBlockHolder: public torch::CustomClassHolder
     /// If this TensorBlock contains gradients, these are gradients w.r.t. this
     /// parameter
     std::string parameter_;
+    /// Shape
+    std::vector<int64_t> shapei64 ;
```
Luthaf, line 185: I don't think we need an extra class variable

Luthaf (commented):
This just needs a couple small things

- adding the new function to metatensor/torch/documentation.py
- adding a test in metatensor-torch Python package (to check that register is working fine)
- adding the functions around [here](https://github.com/lab-cosmo/metatensor/blob/ec6ee61ebb6eab326462386171f5cd777f12a3c4/python/metatensor-torch/tests/block.py#L323) to check that the code is compatible with TorchScript

Luthaf (approved):
I'll merge once CI is happy!

Issue #642, opened 2024-05-31
`metatensor.torch.sort` is broken


frostedoyster:
I leave it to @ppegolo for more details

DavideTisi:
ahahah it is two day that I am try to fix/improve sort

jwa7:
Broken for me too! Here's an MWE:

```py
import torch 
import metatensor.torch as mts

tensor = mts.TensorMap(
    keys=mts.Labels(
        names=["a", "b"],
        values=torch.tensor([[2, 1], [1, 5]]),
    ),
    blocks=[
        mts.TensorBlock(
            values=torch.randn(3, 2),
            samples=mts.Labels(
                names=["s1", "s2", "s3"],
                values=torch.tensor(
                    [
                        [0, 1, 2],
                        [2, 3, 4],
                        [1, 5, 7],
                    ]
                )
            ),
            components=[],
            properties=mts.Labels(
                names=["p1", "p2"],
                values=torch.tensor(
                    [
                        [100, 0],
                        [5, 7000],
                    ]
                )
            )
        ),
        mts.TensorBlock(
            values=torch.randn(3, 3),
            samples=mts.Labels(
                names=["s1", "s2", "s3"],
                values=torch.tensor(
                    [
                        [0, 2, 2],
                        [0, 1, 2],
                        [1, 5, 7],
                    ]
                ),
            ),
            components=[],
            properties=mts.Labels(
                names=["p1", "p2"],
                values=torch.tensor(
                    [
                        [5, 10],
                        [5, 5],
                        [5, 6],
                    ]
                )
            ),
        )
    ]
)
# Sort
sorted = mts.sort(tensor, "samples")

# All metadata is the same
assert tensor.keys == sorted.keys
assert tensor.block(0).samples == sorted.block(0).samples
assert tensor.block(0).properties == sorted.block(0).properties
assert torch.all(tensor.block(0).values == sorted.block(0).values)
```

ppegolo:
Thanks Joe, that's exactly the same issue I'm facing.

jwa7:
I'm guessing the error is here, in `_dispatch.argsort_labels_values()`?

```py
    if isinstance(labels_values, TorchTensor):
        # torchscript does not support sorted for List[List[int]]
        # so we temporary do this trick. this will be fixed with issue #366
        max_int = torch.max(labels_values)
        idx = torch.sum(
            max_int ** torch.arange(labels_values.shape[1]) * labels_values, dim=1
        )
[...]

PR #644, opened 2024-06-01
Feature/sort by

DavideTisi:
I added the possibility to sort only by some columns in a specific axis, using `name`

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1570536484.zip)

### DIFF
--- a/python/metatensor-operations/metatensor/operations/_dispatch.py
+++ b/python/metatensor-operations/metatensor/operations/_dispatch.py
@@ -113,6 +113,26 @@ def allclose(
         raise TypeError(UNKNOWN_ARRAY_TYPE)
 
 
+def argsort(values, axis=-1, reverse: bool = False):
+    """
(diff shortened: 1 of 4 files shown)


DavideTisi:
I open the pr to trigger the CI

DavideTisi:
I am opening this PR to understand what we want to do. 
This PR has two parts (and it fails in both):
- one try to fix #642, @jwa7 example (that I added to the tests) now works, but the torch algo is not a real solution there are other case in which it fails
- two the main reason I started this pr. I wanted a "sort by name" function basically instead of sorting everything it sorts following the column with `name=name`

Luthaf:
I think this can be rebased to only contain the `sort_by` functionality now that #647 is merged.

DavideTisi:
Yes sure

Luthaf:
Closing for now, will be re-opened when ready!

PR #647, opened 2024-06-08
fix the bug with `metatensor.torch.sort`

### LINKED ISSUE #642: `metatensor.torch.sort` is broken

DavideTisi:
fix #642

Fix the bug with `metatensor.torch.sort` #642.
The algorithm is not very smart but is constricted by `TorchScript`.
I also added the test to the torch part

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1589839581.zip)

### DIFF
--- a/python/metatensor-operations/metatensor/operations/_dispatch.py
+++ b/python/metatensor-operations/metatensor/operations/_dispatch.py
@@ -125,6 +125,13 @@ def argsort_labels_values(labels_values, reverse: bool = False):
     """
     if isinstance(labels_values, TorchTensor):
         if labels_values.shape[1] == 1:
+            # Index is appended at the end to get
+            # the indices corresponding to the
+            # sorted values. Because we append the indices at
+            # the end and since metadata is unique, we do not affect
+            # the sorted order. Using `torch.argsort` is not an option
+            # because only on Tensors and sorting a Tensor is
+            # different than sorting a List(Tuple)
             data = [(int(row[0]), i) for i, row in enumerate(labels_values)]
             return torch.tensor(
                 [i[-1] for i in sort_list_2(data, reverse=reverse)],
@@ -881,6 +888,7 @@ def zeros_like(array, shape: Optional[List[int]] = None, requires_grad: bool = F
 def sort_list_2(
     data: List[Tuple[int, int]], reverse: bool = False
 ) -> List[Tuple[int, int]]:
+    # Using the `reverse` flag of sorted does not work with TorchScript
     if reverse:
         return list(sorted(data))[::-1]
     else:
@@ -891,6 +899,7 @@ def sort_list_2(
 def sort_list_3(
     data: List[Tuple[int, int, int]], reverse: bool = False
 ) -> List[Tuple[int, int, int]]:
+    # Using the `reverse` flag of sorted does not work with TorchScript
     if reverse:
         return list(sorted(data))[::-1]
     else:
@@ -901,6 +910,7 @@ def sort_list_3(
 def sort_list_4(
     data: List[Tuple[int, int, int, int]], reverse: bool = False
 ) -> List[Tuple[int, int, int, int]]:
+    # Using the `reverse` flag of sorted does not work with TorchScript
     if reverse:
         return list(sorted(data))[::-1]
     else:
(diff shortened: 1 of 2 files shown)


python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -124,13 +124,132 @@ def argsort_labels_values(labels_values, reverse: bool = False):
...
-        # so we temporary do this trick. this will be fixed with issue #366
-        max_int = torch.max(labels_values)
-        idx = torch.sum(
-            max_int ** torch.arange(labels_values.shape[1]) * labels_values, dim=1
-        )
-        return torch.argsort(idx, dim=-1, descending=reverse)
+        if labels_values.shape[1] == 1:
+            data = [(int(row[0]), i) for i, row in enumerate(labels_values)]
+            return torch.tensor(
+                [i[-1] for i in sort_list_2(data, reverse=reverse)],
```
Luthaf, line 130: instead of adding `i` everywhere, can you do an argsort on the list of tuple?

DavideTisi: doesn't argsort exist only in numpy or similar?

python/metatensor-torch/tests/operations/sort.py
```diff
@@ -24,3 +26,402 @@ def test_save():
...
+def tensor_sorted_ascending():
+    """
+    This is the `tensor` fixture sorted in ascending order how it should be returned
+    when applying metatensor.operations.sort with `descending=False` option.
+    """
+    block_1 = TensorBlock(
+        values=torch.tensor(
+            [
+                [1, 2],
+                [3, 5],
```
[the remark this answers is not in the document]
DavideTisi: done

python/metatensor-operations/metatensor/operations/_dispatch.py
Luthaf: https://pytorch.org/docs/stable/generated/torch.argsort.html

Luthaf: although this might only work with `Tensor`

python/metatensor-operations/metatensor/operations/_dispatch.py
```diff
@@ -756,3 +875,95 @@ def zeros_like(array, shape: Optional[List[int]] = None, requires_grad: bool = F
...
+        return list(sorted(data))
+
+
+@torch_jit_script
+def sort_list_10(
+    data: List[Tuple[int, int, int, int, int, int, int, int, int, int]],
+    reverse: bool = False,
+) -> List[Tuple[int, int, int, int, int, int, int, int, int, int]]:
+    if reverse:
+        return list(sorted(data))[::-1]
```
Luthaf, line 967: Python's sorted has a `reverse` flag as well: https://docs.python.org/3/library/functions.html#sorted

DavideTisi: not working with torch script for some reason

python/metatensor-operations/metatensor/operations/_dispatch.py
DavideTisi: exactly sorting the tensor is not the same as sorting a list of tuples

Luthaf: ok, all good then. Might be worth to add a comment there so I don't ask the question again next time I review this code =)

python/metatensor-operations/metatensor/operations/_dispatch.py
Luthaf: crap. Ok, this is worth a comment as well then

python/metatensor-operations/metatensor/operations/_dispatch.py
DavideTisi: Something like: 
"# Index is appended at the end to get the indices corresponding to the
  # sorted values. Because we append the indices at the end and since metadata
  # is unique, we do not affect the sorted order.
  # Using torch.argsort is not an option because only on Tensors and sorting a Tensor is 
  # different than sorting a List(Tuple)"
  
  ?

Luthaf: that works!

Luthaf (approved):

Issue #650, opened 2024-06-12
Passing ensemble predictions to MD engines


frostedoyster:
Given #648, I thought it was a good idea to open an issue for the forwarding of ensemble predictions to MD engines, which we will need relatively soon.

The design could be as simple as having a new optional property name (something like `ensemble_member`) for any registered output in `metatensor.torch.atomistic`, and then passing this to the metatensor interface to the MD engine, which can decide how to handle it.
i-PI should be well-equipped for this. ASE is also very customizable so we should be able to make the ensemble predictions available to the user relatively easily. I don't know LAMMPS very well, but perhaps we could code something in our driver, or perhaps leave it alone and error out if ensemble predictions are received.

@Luthaf @ceriottm

ceriottm:
In i-PI this should be implemented at the level of the metatensor driver, but yes should be feasible. Davide and Matthias are best positioned to help.

Luthaf:
> The design could be as simple as having a new optional property name (something like ensemble_member) for any registered property in metatensor.torch.atomistic, and then passing this to the metatensor interface to the MD engine, which can decide how to handle it.

I'm not sure I understand what you mean here. If we are adding a new property name to the output, this might have to be a different output. I would also not do something where the behavior changes on the property names, since this makes the whole thing more complex to explain to users and engine developers alike.

----

My favorite solution here would be to do another output name alltogether, with its own specification. So for example `ensemble_energy`, which is similar to `energy` except it has multiple properties per sample. For maximal compatibility, we should also encourage any model that can do `ensemble_energy` to also have a `energy` output, that will be used by any engine that don't know or don't care about ensembles.

----

In general, what's the use case for this? Do we actually need to full ensemble of predictions, or will this always be used to compute the mean and standard deviation for uncertainty quantification purposes?

frostedoyster:
@ceriottm can expand, but passing all the individual predictions is necessary.
And yes, I meant an optional property that may or may not be there. It's more complicated, but it would give us ensembles for free on any property that we might register later on

Luthaf:
Should we keep this open to track implementation in the relevant engines?

PR #654, opened 2024-06-16
Add `energy_ensemble` as a new standard output for atomistic models

frostedoyster:
This PR implements ensembles of energies as a new standard output for atomistic models (#650). It also makes the necessary changes to the ASE calculator to make this property (which is not standard in ASE) available to the user. In addition, the dipole is also registered as a new property and added to the ASE calculator

TODO: actual conversion factors for dipoles

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1719201793.zip)

### DIFF
--- a/docs/src/atomistic/outputs.rst
+++ b/docs/src/atomistic/outputs.rst
@@ -15,6 +15,8 @@ encouraged to come together, define the metadata they need and add a new section
 to this page.
 
 
+.. _energy:
+
 Energy
 ^^^^^^
 
@@ -54,6 +56,8 @@ have the following metadata:
     - ``"energy"``
     - the energy must have a single property dimension named ``"energy"``, with
       a single entry set to ``0``.
+
+.. _energy-gradients:
 
 Energy gradients
 ----------------
@@ -119,3 +123,43 @@ The following gradients can be defined and requested with
     - ``["xyz_1", "xyz_2"]``
     - Both ``"xyz_1"`` and ``"xyz_2"`` have values ``[0, 1, 2]``, and correspond
       to the two axes of the 3x3 strain matrix :math:`\epsilon`.
+
+
+Energy ensemble
+^^^^^^^^^^^^^^^
+
+An ensemble of energies is associated with the ``"energy_ensemble"`` key in the
+model outputs, and must have the following metadata:
+
+.. list-table:: Metadata for energy ensemble output
+  :widths: 2 3 7
+  :header-rows: 1
+
+  * - Metadata
+    - Names
+    - Description
+
+  * - keys
+    - same as `Energy`_
+    - same as `Energy`_
+
+  * - samples
+    - same as `Energy`_
+    - same as `Energy`_
+
+  * - components
+    - same as `Energy`_
+    - same as `Energy`_
+
+  * - properties
+    - ``"ensemble_member"``
+    - the energy ensemble must have a single property dimension named
+      ``"ensemble_member"``, with entries ranging from 0 to the number of
+      members of the ensemble minus one.
+
+
+Energy ensemble gradients
+-------------------------
+
+The gradient metadata for energy ensemble is the same as for the ``energy``
+output (see `Energy gradients`_).
--- a/docs/src/atomistic/outputs.rst
+++ b/docs/src/atomistic/outputs.rst
@@ -129,7 +129,11 @@ Energy ensemble
 ^^^^^^^^^^^^^^^
 
 An ensemble of energies is associated with the ``"energy_ensemble"`` key in the
-model outputs, and must have the following metadata:
+model outputs. Such ensembles are sometimes used to perform uncertainty
+quantification, using multiple prediction to estimate an error on the mean
+prediction.
+
+Energy ensembles must have the following metadata:
 
 .. list-table:: Metadata for energy ensemble output
   :widths: 2 3 7
@@ -152,10 +156,10 @@ model outputs, and must have the following metadata:
     - same as `Energy`_
 
   * - properties
-    - ``"ensemble_member"``
+    - ``"energy"``
     - the energy ensemble must have a single property dimension named
-      ``"ensemble_member"``, with entries ranging from 0 to the number of
-      members of the ensemble minus one.
+      ``"energy"``, with entries ranging from 0 to the number of members of the
+      ensemble minus one.
 
 
 Energy ensemble gradients
--- a/metatensor-torch/src/atomistic/model.cpp
+++ b/metatensor-torch/src/atomistic/model.cpp
@@ -136,7 +136,8 @@ ModelOutput ModelOutputHolder::from_json(std::string_view json) {
 /******************************************************************************/
 
 std::unordered_set<std::string> KNOWN_OUTPUTS = {
-    "energy"
+    "energy",
+    "energy_ensemble"
 };
 
 void ModelCapabilitiesHolder::set_outputs(torch::Dict<std::string, ModelOutput> outputs) {
--- a/python/metatensor-torch/metatensor/torch/atomistic/outputs.py
+++ b/python/metatensor-torch/metatensor/torch/atomistic/outputs.py
@@ -44,30 +44,46 @@ def _check_outputs(
             )
 
         if name == "energy":
-            _check_energy(systems, request, selected_atoms, energy=value)
+            _check_energy_like(
+                "energy",
+                value,
+                systems,
+                request,
+                selected_atoms,
+            )
         elif name == "energy_ensemble":
-            _check_energy_ensemble(
-                systems, request, selected_atoms, energy_ensemble=value
+            _check_energy_like(
+                "energy_ensemble",
+                value,
+                systems,
+                request,
+                selected_atoms,
             )
         else:
             # this is a non-standard output, there is nothing to check
             continue
 
 
-def _check_energy(
+def _check_energy_like(
+    name: str,
+    value: TensorMap,
     systems: List[System],
     request: ModelOutput,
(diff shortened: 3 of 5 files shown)


frostedoyster:
Regarding the tests, it seems that `energy` and `energy_ensemble` are being ordered differently on Windows x)

metatensor-torch/src/atomistic/model.cpp
```diff
@@ -1032,6 +1034,29 @@ static std::unordered_map<std::string, Quantity> KNOWN_QUANTITIES = {
         {"J", "Joule"},
         {"Ry", "Rydberg"},
     }}},
+    {"energy_ensemble", Quantity{/* name */ "energy", /* baseline */ "eV", {
```
Luthaf, line 1037: This should not be needed, `energy_ensemble` should be able to re-use the `energy` quantity and associated units

metatensor-torch/src/atomistic/model.cpp
```diff
@@ -1032,6 +1034,29 @@ static std::unordered_map<std::string, Quantity> KNOWN_QUANTITIES = {
...
+        {"Joule", 1.60218e-19},
+        {"Rydberg", 0.07349864435130857},
+    }, {
+        // alternative names
+        {"J", "Joule"},
+        {"Ry", "Rydberg"},
+    }}},
+    {"dipole", Quantity{/* name */ "dipole", /* baseline */ "D", {
+        {"Debye", 1.0},
+        {"Coulomb-meter", 1000.0},
```
Luthaf, line 1052:
```suggestion
        {"Coulombmeter", 1000.0},
```

metatensor-torch/src/atomistic/model.cpp
```diff
@@ -1032,6 +1034,29 @@ static std::unordered_map<std::string, Quantity> KNOWN_QUANTITIES = {
...
+        {"atomic units", 0.03674932247495664},
+    }, {
+        // alternative names
+        {"D", "Debye"},
+        {"C-m", "Coulomb-meter"},
```
Luthaf, line 1057:
```suggestion
        {"Cm", "Coulombmeter"},
```

This seems to be the SI unit here

metatensor-torch/src/atomistic/model.cpp
```diff
@@ -1032,6 +1034,29 @@ static std::unordered_map<std::string, Quantity> KNOWN_QUANTITIES = {
...
+        {"Rydberg", 0.07349864435130857},
+    }, {
+        // alternative names
+        {"J", "Joule"},
+        {"Ry", "Rydberg"},
+    }}},
+    {"dipole", Quantity{/* name */ "dipole", /* baseline */ "D", {
+        {"Debye", 1.0},
+        {"Coulomb-meter", 1000.0},
+        {"atomic units", 0.03674932247495664},
```
Luthaf, line 1053: I'm not sure I love having a space inside the unit =/

Luthaf (commented):
This is changing way too many things at the same time. Could you have one PR for energy-ensemble, one PR for dipole and one PR changing the ASE calculator? 

The new outputs need code to check that the data matches the specification in https://github.com/lab-cosmo/metatensor/blob/master/python/metatensor-torch/metatensor/torch/atomistic/outputs.py

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -73,6 +74,12 @@ def __init__(
             running, defaults to False.
         :param device: torch device to use for the calculation. If ``None``, we will try
             the options in the model's ``supported_device`` in order.
+        :param properties_to_store: list of model outputs to store as results of the ASE
+            calculator at every step. This is useful when you want to store properties
+            that are not used in the propagation of the dynamics and/or are not standard
+            ASE properties ('energy', 'forces', 'stress', 'stresses', 'dipole',
+            'charges', 'magmom', 'magmoms', 'free_energy', 'energies'). These properties
+            will be available as ``atoms.calc.results['<stored_property>']``.
```
Luthaf, line 82: Is this typical in ASE? The idea for properties not part of the standard was to use the `run_model` function.

In general, this does not seems to be required for dipole, since ASE already supports them as property.

metatensor-torch/src/atomistic/model.cpp
Luthaf, line 1057: Should this be `C.m`? That's what https://en.wikipedia.org/wiki/Electric_dipole_moment uses

metatensor-torch/src/atomistic/model.cpp
```diff
@@ -1032,6 +1034,29 @@ static std::unordered_map<std::string, Quantity> KNOWN_QUANTITIES = {
...
+        {"kcal/mol", 23.060548012069496},
+        {"kJ/mol", 96.48533288249877},
+        {"Joule", 1.60218e-19},
+        {"Rydberg", 0.07349864435130857},
+    }, {
+        // alternative names
+        {"J", "Joule"},
+        {"Ry", "Rydberg"},
+    }}},
+    {"dipole", Quantity{/* name */ "dipole", /* baseline */ "D", {
```
Luthaf, line 1050: do we want to call this `electric_dipole`?

frostedoyster:
Ok, I will open different PRs

sirmarcel:
Thx Guillaume & Filippo. Feel free to re-assign me for ASE changes or ensemble stuff. 👍

docs/src/atomistic/outputs.rst
```diff
@@ -119,3 +119,45 @@ The following gradients can be defined and requested with
     - ``["xyz_1", "xyz_2"]``
     - Both ``"xyz_1"`` and ``"xyz_2"`` have values ``[0, 1, 2]``, and correspond
       to the two axes of the 3x3 strain matrix :math:`\epsilon`.
+
+
+Energy ensemble
+^^^^^^^^^^^^^^^
+
+An ensemble of energies is associated with the ``"energy_ensemble"`` key in the
+model outputs, and must have the following metadata:
```
Luthaf, line 128: I'm split whether we should duplicate the information that is identical to `energy` here, or say "keys: same as `energy`", "samples: same as `energy`", ... and only specify the bit that changes.

Do you have a preference?

python/metatensor-torch/metatensor/torch/atomistic/outputs.py
```diff
@@ -172,3 +176,96 @@ def _check_energy(
...
+def _check_energy_ensemble(
+    systems: List[System],
+    request: ModelOutput,
+    selected_atoms: Optional[Labels],
+    energy_ensemble: TensorMap,
+):
+    """Check the "energy_ensemble" output metadata"""
+    if energy_ensemble.keys != Labels("_", torch.tensor([[0]])):
+        raise ValueError(
+            "invalid keys for 'energy' output: expected `Labels('_', [[0]])`"
```
Luthaf, line 190:
```suggestion
            "invalid keys for 'energy_ensemble' output: expected `Labels('_', [[0]])`"
```

python/metatensor-torch/metatensor/torch/atomistic/outputs.py
```diff
@@ -172,3 +176,96 @@ def _check_energy(
...
+        raise ValueError(
+            "invalid properties for 'energy_ensemble' output: expected a Labels "
+            "object with the name 'ensemble_member' and the values being the indices "
+            "of the ensemble members from 0 to the number of ensemble members minus "
+            "one."
+        )
+
+    if len(energy_ensemble_block.gradients()) != 0:
+        raise ValueError(
+            "invalid gradients for 'energy' output: no gradients should be present."
```
Luthaf, line 270:
```suggestion
            "invalid gradients for 'energy_ensemble' output: no gradients should be present."
```

Luthaf: Do we actually want to disallow the possibility to have forward gradients for `energy_ensemble`?

docs/src/atomistic/outputs.rst
frostedoyster: I would definitely prefer "same as energies"

python/metatensor-torch/metatensor/torch/atomistic/outputs.py
frostedoyster: Yeah... they make no sense... unless we give them a different meaning.
For example, the uncertainties on the forces are not the derivatives of the uncertainties on the energies

Luthaf: I mean, the output is not called "energy_uncertainties" but "energy_ensemble", so the positions gradients would be an ensemble of forces. It would then be up to the engine to do what they want with it.

I'm also fine disallowing it for now, but then we should mention it in the standard.

docs/src/atomistic/outputs.rst
Luthaf: let's do this then

python/metatensor-torch/metatensor/torch/atomistic/outputs.py
frostedoyster: You're right, I got confused. The gradients for `energy_ensemble` are well-defined and it's a good idea to allow them as explicit gradients

docs/src/atomistic/outputs.rst
frostedoyster: Should be done

Luthaf:
This looks good. Should we also have a test checking some example model that returns an ensemble of energies?

Issue #655, opened 2024-06-16
ASE interface re-runs the model multiple times at each step


frostedoyster:
This is how ASE calls calculators to get properties:

```
    def get_property(self, name, atoms=None, allow_calculation=True):
        if name not in self.implemented_properties:
            raise PropertyNotImplementedError('{} property not implemented'
                                              .format(name))

        if atoms is None:
            atoms = self.atoms
            system_changes = []
        else:
            system_changes = self.check_state(atoms)
            if system_changes:
                self.reset()
        if name not in self.results:
            if not allow_calculation:
                return None
            self.calculate(atoms, [name], system_changes)

        if name not in self.results:
            # For some reason the calculator was not able to do what we want,
            # and that is OK.
            raise PropertyNotImplementedError('{} not present in this '
                                              'calculation'.format(name))

        result = self.results[name]
        if isinstance(result, np.ndarray):
            result = result.copy()
        return result
```

Essentially one property at a time. This means that our model will be called once to get the forces at each MD step. If the user wants to know the potential energy, the calculator will be called again. If the simulation is NPT and a stress is required, this generates another call. Dipoles would be yet another call and so on.

This is a consequence of the fact that our calculator stores as `results` only the requested property. Therefore no caching happens and the calculator has to be queried again.

frostedoyster:
A custom `get_property` for our calculator might be one way to solve the issue

Luthaf:
We could also store the energy/energies in `results` whenever we are computing forces/stress. We could also always store the stress when computing forces (and vice versa). The only one I would not do here would be to computed forces/stress if the user only asked for energy (since this is quite a bit more expensive).

Overriding `get_property` might work, but I would rather only implement what ASE wants us to implement, to prevent the code from breaking in the next release.

sirmarcel:
Heya, I've been called in to comment on this. What's the problem here? As far as I'm aware the expected way to deal with this is to write everything you possibly can into `results` (in the `calculate` method) so it doesn't have to be recomputed. So a `get_forces` should populate `results` with the potential energy (and maybe the stress also).

Luthaf:
Fixed by #655

PR #658, opened 2024-06-18
Store additional properties in ASE calculator

frostedoyster:
These changes allow for a user to calculate and store non-standard properties in the ASE calculator. For relatively experienced users, this is also a fix for #655 (one can set `"energy", "forces", "stress"` as `properties_to_store`, these will be calculated at each step, and then the inherited `get_property` function will not re-trigger a calculation but just read from the `calc.results` dict).

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1723565709.zip)

### DIFF
--- a/python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
+++ b/python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
@@ -253,6 +253,25 @@ class MetatensorCalculator(ase.calculators.calculator.Calculator):
             system_changes=system_changes,
         )
 
+        # In the next few lines, we decide which properties to calculate among energy,
+        # forces and stress. In addition to the requested properties, we calculate the
+        # energy if any of the three is requested, as it is an intermediate step in the
+        # calculation of the other two. We also calculate the forces if the stress is
+        # requested, and vice-versa. The overhead for the latter operation is also
+        # small, assuming that the majority of the model computes forces and stresses
+        # by backpropagation as opposed to forward-mode differentiation.
+        calculate_energy = (
+            "energy" in properties
+            or "energies" in properties
+            or "forces" in properties
+            or "stress" in properties
+        )
+        calculate_energies = "energies" in properties
+        calculate_forces = "forces" in properties or "stress" in properties
+        calculate_stress = "stress" in properties or "forces" in properties
+        if "stresses" in properties:
+            raise NotImplementedError("'stresses' are not implemented yet")
+
         with record_function("ASECalculator::prepare_inputs"):
             outputs = _ase_properties_to_metatensor_outputs(properties)
             capabilities = self._model.capabilities()
@@ -268,11 +287,11 @@ class MetatensorCalculator(ase.calculators.calculator.Calculator):
             )
 
             do_backward = False
-            if "forces" in properties:
+            if calculate_forces:
                 do_backward = True
                 positions.requires_grad_(True)
 
-            if "stress" in properties:
+            if calculate_stress:
                 do_backward = True
 
                 strain = torch.eye(
@@ -283,9 +302,6 @@ class MetatensorCalculator(ase.calculators.calculator.Calculator):
                 positions.retain_grad()
 
                 cell = cell @ strain
-
-            if "stresses" in properties:
-                raise NotImplementedError("'stresses' are not implemented yet")
 
             run_options = ModelEvaluationOptions(
                 length_unit="angstrom",
@@ -335,14 +351,14 @@ class MetatensorCalculator(ase.calculators.calculator.Calculator):
 
             self.results = {}
 
-            if "energies" in properties:
+            if calculate_energies:
                 energies_values = energies.detach().reshape(-1)
                 energies_values = energies_values.to(device="cpu").to(
                     dtype=torch.float64
                 )
                 self.results["energies"] = energies_values.numpy()
 
-            if "energy" in properties:
+            if calculate_energy:
                 energy_values = energy.detach()
                 energy_values = energy_values.to(device="cpu").to(dtype=torch.float64)
                 self.results["energy"] = energy_values.numpy()[0, 0]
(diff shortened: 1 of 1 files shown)


Luthaf:
Why did you go with this approach instead of automatically storing energy/energies when computing the forces and stress? The fact that one has to explicitly set these is not ideal, and I feel like a solution to #655 that works without explicit user action would be nicer.

frostedoyster:
This was not intended as a fix for #655 (although it is if one knows the internals). It's just to propagate potentially more outputs to the user

Luthaf:
Ok, the initially intended API for doing outputs outside of ASE specification was the [`run_model`](https://docs.metatensor.org/latest/atomistic/reference/ase.html#metatensor.torch.atomistic.ase_calculator.MetatensorCalculator.run_model) function. Why is it not suitable in your case?

frostedoyster:
It looks to me like it recomputes everything...
IMO the only way not to recompute if one wants additional properties is to save stuff to the `calculator.results` dict, because that is what's checked to avoid recomputations (if the Atoms object hasn't changed)

Luthaf:
> It looks to me like it recomputes everything...

No, because this function is **not** integrated with the rest of ASE interface (`get_property` & friends). This should be used like this:

```py
calculator = MetatensorCalculator(...)
outputs = {...}

calculator.run_model(atoms, outputs)
```

Since the user is explicitly calling this function themself, it will only recompute if the users call it again manually.

---

In general there are two separate issues IMO: #655 is about storing multiple that ASE knows about (storing energy when computing forces, etc.); while `run_model` is about computing properties ASE does not know about (i.e. `mtt::last_layer_features`, `nmr_shielding`, etc.).

We need a fix for #655, but I don't think that `properties_to_store` is the best solution. I would rather do this automatically without any user intervention.

frostedoyster:
The fact that it is not integrated with the rest of ASE is why the recomputations are necessary. If I need a quantity at some step of MD, I have to recompute

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -352,18 +368,54 @@ def calculate(
...
 
-            if "stress" in properties:
+            if "stress" in properties_to_calculate:
                 stress_values = -strain.grad.reshape(3, 3) / atoms.cell.volume
                 stress_values = stress_values.to(device="cpu").to(dtype=torch.float64)
                 self.results["stress"] = _full_3x3_to_voigt_6_stress(
                     stress_values.numpy()
                 )
 
+    def request_properties_every_n_steps(
```
Luthaf, line 383: I still don't think this should be part of the same object. The calculator should follow the ASE API for calculators, and we should provide a separate tool (class, function, …) to handle calculations that have to run at other times.

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -73,6 +74,7 @@ def __init__(
         extensions_directory=None,
         check_consistency=False,
         device=None,
+        properties_to_store: Optional[List[str]] = None,
```
Luthaf, line 77:
```suggestion
        extra_metatensor_outputs: Optional[Dict[str, ModelOutput]] = None,
```

I would rename this and handle it a bit differently from "normal" properties to make it clearer that this mechanism is only intended for things which are *not* already known ASE properties. 

IMO the data should also be stored as TensorMap in `calc.results["extra_metatensor_outputs"]["..."]`, again to make it clear we are operating outside of the ASE interface.

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
frostedoyster: Ok, I can extract it into a standalone function

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
frostedoyster: I agree that it would be good in principle but it breaks the ASE caching mechanism and it turns it de facto into a different calculator with no re-use (which is pointless because we already have the function to call the model separately)

Luthaf: > but it breaks the ASE caching mechanism

How? Especially if we only allow things ASE has no knowledge about in `extra_metatensor_outputs`?

> it turns it de facto into a different calculator with no re-use

No? There will be a single model call to compute both "energy", "forces" and e.g. "last_layer_features". Any weight reuse is then up to the model.

bananenpampe:
Hello, 

Yuxuan and I have ported a preliminary chemical shielding model to an ase-MetatensorCalculator.
This model returns atom centered non standard properties, namely the chemical shieldings.
I would feel like, that the most natural way to compute these properties woule be like this (ideally even without having to specify the domain name itself):

```python
calc = MetatensorCalculator("model.pt")
frame.set_calculator(calc)
cs_iso = frame.get_properties(["mtt::cs_iso"])
```

However, currently it requires doing this, which I find not so intuitive:

```python
output = calc.run_model(frame, outputs)
cs_iso = output['mtt::cs_iso'].block(0).values.detach().numpy()
```

I personally feel like the results should be stored in the results dict as numpy arrays, that way models can be used from users outside the metatensor universe.

Luthaf:
This PR would remove the need to explicitly call `run_model` and instead put the data in a dict, but you would still have to provide the `ModelOutput` and you would get the results as `TensorMap`.

> I personally feel like the results should be stored in the results dict as numpy arrays, that way models can be used from users outside the metatensor universe.

This is not possible for fully generic metatensor outputs, since they could have multiple keys/blocks, and then this function would not know what to do with them.

I agree that doing things like this would be nice, but it does not fit in the ASE calculator framework. We could however define a new API that could do what you are asking *for standardized outputs*, with something like this

```py
calc = MetatensorCalculator("model.pt")
cs_iso = calc.get_metatensor_property(atoms, "scalar_chemical_shielding")

assert isinstance(cs_iso, np.ndarray)
```

This requires first defining what the standard outputs are and what `get_metatensor_property` should do for these outputs.
For example, what does this function do if there are multiple blocks in the output? What does it do if there are missing samples? How does it differentiate per atom or per structure outputs? 

I'll open a separate issue to discuss this further, this PR is mainly aimed at things that should be computed together with the energy/forces.

Luthaf (approved):

Issue #402, opened 2023-10-12
Release on conda-forge


Luthaf:
In addition to PyPI, we should also provide the release on [conda forge](https://conda-forge.org/)! 

I think we should have the following packages over there:

- metatensor: empty package that depends on most/all of the other packages below
- metatensor-core
- metatensor-core-python
- metatensor-operations-python
- metatensor-torch
- metatensor-torch-python

Ideally, `metatensor-core-python` and `metatensor-torch-python` would not have their own copy of the shared libraries from `metatensor-core` and `metatensor-torch` but use the shared libraries for the other packages. That requires some changes to the code, so we might leave it to a future version.

PicoCentauri:
Yes, as I said I can help with that. You mentioned a repo that we can use as reference for the rust building?

Luthaf:
We should not have to build the Rust code directly, `metatensor-core` should be built by CMake. The only point will be to install a rust compiler when building the code, but that should be doable.

I might have mentioned chemfiles where I do the separation between a pure C++ package, and then a Python package which depends on the first one and uses the shared library from the pure C++ package.

Luthaf:
libmetatensor-torch is now on conda-forge: https://github.com/conda-forge/libmetatensor-torch-feedstock

Luthaf:
The python packages are all on conda-forge:

- https://github.com/conda-forge/python-metatensor-core-feedstock/
- https://github.com/conda-forge/python-metatensor-torch-feedstock/
- https://github.com/conda-forge/python-metatensor-operations-feedstock/
- https://github.com/conda-forge/python-metatensor-learn-feedstock/

PR #664, opened 2024-06-21
Add OpenMM-torch interface

frostedoyster:
Adds a MD interface to OpenMM via OpenMM-torch

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1670702696.zip)

⚙️ [Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1624015907.zip)

### DIFF
--- a/docs/src/atomistic/reference/openmm.rst
+++ b/docs/src/atomistic/reference/openmm.rst
@@ -7,9 +7,7 @@ OpenMM integration
 which can be used to load a :py:class:`MetatensorAtomisticModel` into an
 ``openmm.Force`` object able to calculate forces on an ``openmm.System``.
 
-.. autoclass:: metatensor.torch.atomistic.openmm_interface.get_metatensor_force
-    :show-inheritance:
-    :members:
+.. autofunction:: metatensor.torch.atomistic.openmm_force.get_metatensor_force
 
 In order to run simulations with ``metatensor.torch.atomistic`` and ``OpenMM``,
 we recommend installing ``OpenMM`` from conda, using
--- a/docs/src/atomistic/reference/openmm.rst
+++ b/docs/src/atomistic/reference/openmm.rst
@@ -1,0 +1,41 @@
+OpenMM integration
+==================
+
+.. py:currentmodule:: metatensor.torch.atomistic
+
+:py:mod:`openmm_interface` contains the ``get_metatensor_force`` function,
+which can be used to load a :py:class:`MetatensorAtomisticModel` into an
+``openmm.Force`` object able to calculate forces on an ``openmm.System``.
+
+.. autoclass:: metatensor.torch.atomistic.openmm_interface.get_metatensor_force
+    :show-inheritance:
+    :members:
+
+In order to run simulations with ``metatensor.torch.atomistic`` and ``OpenMM``,
+we recommend installing ``OpenMM`` from conda, using
+``conda install -c conda-forge openmm-torch nnpops``. Subsequently,
+metatensor can be installed with ``pip install metatensor[torch]``, and a minimal
+script demonstrating how to run a simple simulation is illustrated below:
+
+.. code-block:: python
+
+    import openmm
+    from metatensor.torch.atomistic.openmm_interface import get_metatensor_force
+
+    # load an input geometry file
+    topology = openmm.app.PDBFile('input.pdb').getTopology()
+    system = openmm.System()
+    for atom in topology.atoms():
+        system.addParticle(atom.element.mass)
+    
+    # get the force object from an exported model saved as 'model.pt'
+    force = get_metatensor_force(system, topology, 'model.pt')
+    system.addForce(force)
+
+    integrator = openmm.VerletIntegrator(0.001)
+    platform = openmm.Platform.getPlatformByName('CUDA')
+    simulation = openmm.app.Simulation(topology, system, integrator, platform)
+    simulation.context.setPositions(openmm.app.PDBFile('input.pdb').getPositions())
+    simulation.reporters.append(openmm.app.PDBReporter('output.pdb', 100))
+
+    simulation.step(1000)
--- a/python/metatensor-operations/CHANGELOG.md
+++ b/python/metatensor-operations/CHANGELOG.md
@@ -17,10 +17,6 @@ follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Removed
 -->
 
-### Changed
-
-- We now require Python >= 3.9
-
 ## [Version 0.2.2](https://github.com/lab-cosmo/metatensor/releases/tag/metatensor-operations-v0.2.2) - 2024-06-19
 
 ### Fixed
--- a/python/metatensor-torch/metatensor/torch/atomistic/openmm_force.py
+++ b/python/metatensor-torch/metatensor/torch/atomistic/openmm_force.py
@@ -13,9 +13,9 @@ from metatensor.torch.atomistic import (
 
 
 try:
+    import NNPOps.neighbors
     import openmm
     import openmmtorch
-    from NNPOps.neighbors import getNeighborPairs
 
     HAS_OPENMM = True
 except ImportError:
@@ -38,9 +38,25 @@ def get_metatensor_force(
     path: str,
     extensions_directory: Optional[str] = None,
     forceGroup: int = 0,
-    atoms: Optional[Iterable[int]] = None,
+    selected_atoms: Optional[Iterable[int]] = None,
     check_consistency: bool = False,
-) -> openmm.System:
+) -> openmm.Force:
+    """
+    Create an OpenMM Force from a metatensor atomistic model.
+
+    :param system: The OpenMM System object.
+    :param topology: The OpenMM Topology object.
+    :param path: The path to the exported metatensor model.
+    :param extensions_directory: The path to the extensions for the model.
+    :param forceGroup: The force group to which the force should be added.
+    :param selected_atoms: The indices of the atoms on which non-zero forces should
+        be computed (e.g., the ML region). If None, forces will be computed for all
+        atoms.
+    :param check_consistency: Whether to check various consistency conditions inside
+        the model.
+
+    return: The OpenMM Force object.
+    """
 
     if not HAS_OPENMM:
         raise ImportError(
@@ -52,15 +68,15 @@ def get_metatensor_force(
 
     # Get the atomic numbers of the ML region.
     all_atoms = list(topology.atoms())
-    atomic_numbers = [atom.element.atomic_number for atom in all_atoms]
+    atomic_types = [atom.element.atomic_number for atom in all_atoms]
 
-    if atoms is None:
+    if selected_atoms is None:
         selected_atoms = None
     else:
         selected_atoms = Labels(
             names=["system", "atom"],
             values=torch.tensor(
-                [[0, selected_atom] for selected_atom in atoms],
+                [[0, selected_atom] for selected_atom in selected_atoms],
                 dtype=torch.int32,
             ),
         )
@@ -69,8 +85,8 @@ def get_metatensor_force(
 
         def __init__(
             self,
-            model: torch.jit._script.RecursiveScriptModule,
-            atomic_numbers: List[int],
+            model,
+            atomic_types: List[int],
             selected_atoms: Optional[Labels],
             check_consistency: bool,
         ) -> None:
@@ -78,7 +94,7 @@ def get_metatensor_force(
 
             self.model = model
             self.register_buffer(
-                "atomic_numbers", torch.tensor(atomic_numbers, dtype=torch.int32)
+                "atomic_types", torch.tensor(atomic_types, dtype=torch.int32)
             )
             self.evaluation_options = ModelEvaluationOptions(
                 length_unit="nm",
@@ -125,26 +141,23 @@ def get_metatensor_force(
 
             # create System
             system = System(
-                types=self.atomic_numbers,
+                types=self.atomic_types,
                 positions=positions,
                 cell=cell,
             )
             system = _attach_neighbors(system, self.requested_neighbor_list)
 
-            energy = (
-                self.model(
-                    [system],
-                    self.evaluation_options,
-                    check_consistency=self.check_consistency,
-                )["energy"]
-                .block()
-                .values.reshape(())
+            outputs = self.model(
+                [system],
+                self.evaluation_options,
+                check_consistency=self.check_consistency,
             )
+            energy = outputs["energy"].block().values.reshape(())
             return energy
 
     metatensor_force = MetatensorForce(
         model,
-        atomic_numbers,
+        atomic_types,
         selected_atoms,
         check_consistency,
     )
@@ -175,11 +188,11 @@ def _attach_neighbors(
         cell = system.cell
 
     # Get the neighbor pairs, shifts and edge indices.
-    neighbors, interatomic_vectors, _, _ = getNeighborPairs(
-        system.positions,
-        requested_nl_options.engine_cutoff("nm"),
-        -1,
-        cell,
+    neighbors, interatomic_vectors, _, _ = NNPOps.neighbors.getNeighborPairs(
+        positions=system.positions,
+        cutoff=requested_nl_options.engine_cutoff("nm"),
+        max_num_pairs=-1,
+        box_vectors=cell,
     )
     mask = neighbors[0] >= 0
     neighbors = neighbors[:, mask]
--- a/python/metatensor-torch/metatensor/torch/atomistic/openmm_force.py
+++ b/python/metatensor-torch/metatensor/torch/atomistic/openmm_force.py
@@ -68,6 +68,9 @@ def get_metatensor_force(
 
     model = load_atomistic_model(path, extensions_directory=extensions_directory)
 
+    # Print the model's metadata
+    print(model.metadata().print())
+
     # Get the atomic numbers of the ML region.
     all_atoms = list(topology.atoms())
     atomic_types = [atom.element.atomic_number for atom in all_atoms]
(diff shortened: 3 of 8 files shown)


python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
```diff
@@ -0,0 +1,161 @@
...
+        isPeriodic = (
+            topology.getPeriodicBoxVectors() is not None
+        ) or system.usesPeriodicBoundaryConditions()
+        force.setUsesPeriodicBoundaryConditions(isPeriodic)
+        force.setForceGroup(forceGroup)
+
+        system.addForce(force)
+
+
+MLPotential.registerImplFactory("metatensor", MetatensorPotentialImplFactory)
```
Luthaf, line 161: this will fail if we don't have OpenMM installed.

Luthaf (commented):
Why is this using openmm-ml instead of openmm-torch directly? My understanding of the former was that it is supposed to provide specific models, not a whole new set of models?

----

This will also need tests, documentation and examples. And I guess this required #402 to be done first as well, since OpenMM relies heavily on conda.

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
```diff
@@ -0,0 +1,161 @@
...
+        **args,
+    ) -> None:
+
+        if not HAS_OPENMM:
+            raise ImportError(
+                "Could not import openmm. If you want to use metatensor with "
+                "openmm, please install openmm-ml with conda."
+            )
+
+        model = load_atomistic_model(self.path)  # TODO: extensions_directory
```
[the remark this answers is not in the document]
frostedoyster: Gone

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
```diff
@@ -0,0 +1,161 @@
...
+
+        # Get the atomic numbers of the ML region.
+        all_atoms = list(topology.atoms())
+        atomic_numbers = [atom.element.atomic_number for atom in all_atoms]
+
+        # TODO: Set up selected_atoms as a Labels object
```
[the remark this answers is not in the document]
frostedoyster: It was done indeed

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
frostedoyster: Gone

docs/src/atomistic/reference/openmm.rst
```diff
@@ -0,0 +1,41 @@
...
+
+.. py:currentmodule:: metatensor.torch.atomistic
+
+:py:mod:`openmm_interface` contains the ``get_metatensor_force`` function,
+which can be used to load a :py:class:`MetatensorAtomisticModel` into an
+``openmm.Force`` object able to calculate forces on an ``openmm.System``.
+
+.. autoclass:: metatensor.torch.atomistic.openmm_interface.get_metatensor_force
+    :show-inheritance:
+    :members:
```
Luthaf, line 12:
```suggestion
.. autofunction:: metatensor.torch.atomistic.openmm_interface.get_metatensor_force
```

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
```diff
@@ -0,0 +1,235 @@
...
+
+def get_metatensor_force(
+    system: openmm.System,
+    topology: openmm.app.Topology,
+    path: str,
+    extensions_directory: Optional[str] = None,
+    forceGroup: int = 0,
+    atoms: Optional[Iterable[int]] = None,
+    check_consistency: bool = False,
+) -> openmm.System:
```
Luthaf, line 42: this needs a docstring & doc for all parameters

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
```diff
@@ -0,0 +1,235 @@
...
+    HAS_OPENMM = False
+
+
+def get_metatensor_force(
+    system: openmm.System,
+    topology: openmm.app.Topology,
+    path: str,
+    extensions_directory: Optional[str] = None,
+    forceGroup: int = 0,
+    atoms: Optional[Iterable[int]] = None,
```
Luthaf, line 40:
```suggestion
    selected_atoms: Optional[Iterable[int]] = None,
```

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
```diff
@@ -0,0 +1,235 @@
...
+                dtype=torch.int32,
+            ),
+        )
+
+    class MetatensorForce(torch.nn.Module):
+
+        def __init__(
+            self,
+            model: torch.jit._script.RecursiveScriptModule,
+            atomic_numbers: List[int],
```
Luthaf, line 72:
```suggestion
            atomic_types: List[int],
```

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
```diff
@@ -0,0 +1,235 @@
...
+
+            if self.requested_neighbor_list is None:
+                return system
+
+            cell: Optional[torch.Tensor] = None
+            if not torch.all(system.cell == 0.0):
+                cell = system.cell
+
+            # Get the neighbor pairs, shifts and edge indices.
+            neighbors, interatomic_vectors, _, _ = getNeighborPairs(
```
Luthaf, line 119:
```suggestion
            neighbors, interatomic_vectors, _, _ = NNPOps.neighbors.getNeighborPairs(
```

for clarity

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
```diff
@@ -0,0 +1,235 @@
...
+                system.positions,
+                self.requested_neighbor_list.engine_cutoff("nm"),
+                -1,
```
Luthaf, line 122: could you give the name of this argument? (and maybe others as well)

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
```diff
@@ -0,0 +1,235 @@
...
+                cell,
+            )
+            mask = neighbors[0] >= 0
```
Luthaf, line 125: will there negative values in `neighbors[0]` if `n_pairs=-1` above?

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
```diff
@@ -0,0 +1,235 @@
...
+
+            energy = (
+                self.model(
+                    [system],
+                    self.evaluation_options,
+                    check_consistency=self.check_consistency,
+                )["energy"]
+                .block()
+                .values.reshape(())
+            )
```
Luthaf, line 214:
```suggestion
            outputs = self.model(
                [system],
                self.evaluation_options,
                check_consistency=self.check_consistency,
            )
            energy = outputs["energy"].block().values.reshape(())
```

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
```diff
@@ -0,0 +1,235 @@
...
+) -> openmm.System:
+
+    if not HAS_OPENMM:
+        raise ImportError(
+            "Could not import openmm and/or nnpops. If you want to use metatensor with "
+            "openmm, please install openmm-torch and nnpops with conda."
+        )
+
+    model = load_atomistic_model(path, extensions_directory=extensions_directory)
+
```
Luthaf, line 51: Could you print `model.metadata()` to the log somewhere? This prints information about the model. 

Also, is there a mechanism to register papers to cite in OpenMM? If so, we should register the model references.

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
```diff
@@ -0,0 +1,235 @@
...
+                [[0, selected_atom] for selected_atom in atoms],
+                dtype=torch.int32,
+            ),
+        )
+
+    class MetatensorForce(torch.nn.Module):
+
+        def __init__(
+            self,
+            model: torch.jit._script.RecursiveScriptModule,
```
Luthaf, line 71:
```suggestion
            model,
```

We don't really need the types in `__init__`, and this one is private

Luthaf (commented):
I could not find how this code handle dtype/device? The model might be on a different device than the positions/cell.

For the dtype, the model might require a specific one and the system should be converted to this dtype

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
frostedoyster: Yes, I think that NL returns a huge Tensor with all pairs, where only a certain number are not -1, i.e., masked

frostedoyster:
Typical code to start a simulation would look like this:

```
platform = openmm.Platform.getPlatformByName('CUDA')
properties = {'Precision': 'double'}
simulation = openmm.app.Simulation(topology, system, integrator, platform, properties)
```

The device is handled by OpenMM torch (it moves the model to the desired device).
The dtype is more tricky, as it is not handled by OpenMM (I think they're trying to allow model developers to do things in mixed precision, so they won't call `to(dtype)`). At the moment, if the precision is wrong, the model will crash at runtime, and if the user matches the dtypes correctly, the model will run.

Let me know what you think the best solution is (leave as is, check at each step, etc)

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
frostedoyster: I think they have some MD loggers, but I don't think they're accessible from our function (they're attached to a `Simulation` object that is independent). Literally zero clue regarding the citation mechanism

frostedoyster: Perhaps the best solution is to `print()` and whatever

Luthaf:
> Let me know what you think the best solution is (leave as is, check at each step, etc)

The way we are doing this in other engines is to always convert from the engine dtype to the model dtype for inputs, and the other way around for outputs. We could add a warning about mismatched dtype if we do the conversion, but I would rather allow the code to run instead of crashing.

python/metatensor-torch/metatensor/torch/atomistic/openmm_interface.py
Luthaf: I understand this behavior if we are setting `max_n_pairs=300`: if there are fewer pairs, the other ones are filled with -1. You are saying this is also the case when giving `max_n_pairs=-1`?

frostedoyster: Yes, I tried a few tests (with `max_n_pairs=-1`) and the mask is still necessary. I think they're always returning a number of pairs that is a multiple of 4. E.g. if you have 27 pairs, 28 will be returned and one of them is masked with center=neighbor=-1

frostedoyster:
Printing metadata doesn't work because `::print()` is not registered it seems, I'll see if I can fix it

docs/src/atomistic/reference/openmm.rst
```diff
@@ -0,0 +1,39 @@
...
+:py:mod:`openmm_interface` contains the ``get_metatensor_force`` function,
+which can be used to load a :py:class:`MetatensorAtomisticModel` into an
+``openmm.Force`` object able to calculate forces on an ``openmm.System``.
+
+.. autofunction:: metatensor.torch.atomistic.openmm_force.get_metatensor_force
+
+In order to run simulations with ``metatensor.torch.atomistic`` and ``OpenMM``,
+we recommend installing ``OpenMM`` from conda, using
+``conda install -c conda-forge openmm-torch nnpops``. Subsequently,
+metatensor can be installed with ``pip install metatensor[torch]``, and a minimal
```
Luthaf, line 15: Since there is #665, should we recommend building metatensor-torch from sources for now? Something like ``pip install --no-binary=metatensor-torch metatensor[torch]``

python/metatensor-operations/CHANGELOG.md
```diff
@@ -17,10 +17,6 @@ follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Removed
 -->
 
-### Changed
```
Luthaf, line 20: this removal looks like a merge issue

python/metatensor-torch/metatensor/torch/atomistic/openmm_force.py
```diff
@@ -0,0 +1,286 @@
...
+    check_consistency: bool = False,
+) -> openmm.Force:
+    """
+    Create an OpenMM Force from a metatensor atomistic model.
+
+    :param system: The OpenMM System object.
+    :param topology: The OpenMM Topology object.
+    :param path: The path to the exported metatensor model.
+    :param extensions_directory: The path to the extensions for the model.
+    :param forceGroup: The force group to which the force should be added.
```
Luthaf, line 56: should we link to OpenMM docs for force group here? It might not be clear to everyone.

docs/src/atomistic/reference/openmm.rst
```diff
@@ -0,0 +1,39 @@
...
+.. py:currentmodule:: metatensor.torch.atomistic
+
+:py:mod:`openmm_interface` contains the ``get_metatensor_force`` function,
+which can be used to load a :py:class:`MetatensorAtomisticModel` into an
+``openmm.Force`` object able to calculate forces on an ``openmm.System``.
+
+.. autofunction:: metatensor.torch.atomistic.openmm_force.get_metatensor_force
+
+In order to run simulations with ``metatensor.torch.atomistic`` and ``OpenMM``,
+we recommend installing ``OpenMM`` from conda, using
```
Luthaf, line 13: you should mention than NNPOps is a required dependency as well in here

python/metatensor-torch/metatensor/torch/atomistic/openmm_force.py
```diff
@@ -0,0 +1,286 @@
...
+        atomic_types,
+        selected_atoms,
+        check_consistency,
+    )
+
+    # torchscript everything
+    module = torch.jit.script(metatensor_force)
+
+    # create the OpenMM force
+    force = openmmtorch.TorchForce(module)
```
Luthaf, line 194: so `model.to(device)` is called by TorchForce? Could you leave a comment if this is the case, so future readers don't wonder what's happening?

python/metatensor-torch/tests/atomistic/openmm_force.py
```diff
@@ -0,0 +1,186 @@
...
+        cutoff=CUTOFF,
+        sigma=SIGMA,
+        epsilon=EPSILON,
+        length_unit="Angstrom",
+        energy_unit="eV",
+        with_extension=False,
+    )
+
+
+def model_different_units():
```
Luthaf, line 39: this might not be so interesting to run here, I did it in ASE to ensure our code did unit conversions right, but we don't have to do it for all interfaces.

python/metatensor-torch/tests/atomistic/openmm_force.py
```diff
@@ -0,0 +1,186 @@
...
+    if atoms.pbc.any():
+        system.setDefaultPeriodicBoxVectors(
+            atoms.cell[0] * openmm.unit.angstrom,
+            atoms.cell[1] * openmm.unit.angstrom,
+            atoms.cell[2] * openmm.unit.angstrom,
+        )
+    force = get_metatensor_force(system, topology, model_path, check_consistency=True)
+    system.addForce(force)
+    integrator = openmm.VerletIntegrator(0.001 * openmm.unit.picoseconds)
+    platform = openmm.Platform.getPlatformByName("CPU")
```
Luthaf, line 86: should we try to run this test on CUDA if it is available?

python/metatensor-torch/tests/atomistic/openmm_force.py
```diff
@@ -0,0 +1,186 @@
...
+    simulation = openmm.app.Simulation(
+        topology, system, integrator, platform, properties
+    )
+    simulation.context.setPositions(openmm.app.PDBFile(structure_path).getPositions())
+    state = simulation.context.getState(getForces=True)
+    openmm_forces = (
+        state.getForces(asNumpy=True).value_in_unit(
+            openmm.unit.ev / (openmm.unit.angstrom * openmm.unit.mole)
+        )
+        / 6.0221367e23
```
Luthaf, line 97: maybe put this in a constant on top?

Luthaf: also, why are you doing `* openmm.unit.mole` above, just to divide again here?

python/metatensor-torch/tests/atomistic/openmm_force.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+try:
+    import NNPOps  # noqa: F401
+    import openmm
+    import openmmtorch  # noqa: F401
+
+    HAS_OPENMM = True
+except ImportError:
+    HAS_OPENMM = False
```
Luthaf, line 20: Since we want to skip the whole file if one of the dependency is missing, we should use importskip:

```suggestion
openmm = pytest.importskip("openmm")

pytest.importskip("NNPOps")
pytest.importskip("openmmtorch")
```

Luthaf (commented):
Looks good! We should find a way to run the tests for this. I guess the main issue is that openmm requires conda for installation, right? 

Either we keep the code in this repo, and we find a way to use conda from tox; or we move the code to another repo, with its own CI setup.

Maybe adding a section in `tox.ini` for "tests that require conda", and in there creating a conda env by calling conda explicitly could work.

docs/src/atomistic/reference/openmm.rst
frostedoyster: Sounds good

Luthaf:
We'll need to move this to https://github.com/metatensor/metatomic

HaoZeke:
Closing as per discussions with @frostedoyster

PR #668, opened 2024-06-28
Do not export models on device

frostedoyster:
If one exports a model on a device and then tries to load it when the device is not available anymore, you will get
`cpp_module = torch._C.import_ir_module(cu, os.fspath(f), map_location, _extra_files, _restore_shapes)  # type: ignore[call-arg]`
`RuntimeError: No CUDA GPUs are available`

As a result, we should always save on CPU and then it is the responsibility of whoever loads the model to change the device if needed

# Reviewer checklist


Luthaf:
Could you add this fix to the changelog?

Luthaf (approved):

PR #671, opened 2024-07-03
Allow loading npz files which use DEFLATE compression

### LINKED ISSUE #469: When loading tensor map that requires `use_numpy=True` the error message is unclear
When I load the tensor map
```python
In [1]: import metatensor
In [2]: metatensor.load("./python/metatensor-operations/tests/data/qm7-power-spectrum.npz")
MetatensorError                           Traceback (most recent call last)
Cell In[2], line 1
----> 1 metatensor.load("./python/metatensor-operations/tests/data/qm7-power-spectrum.npz")

File /opt/anaconda3/envs/metatensor-dev/lib/python3.11/site-packages/metatensor/io.py:72, in load(file, use_numpy)
     70 else:
     71     if isinstance(file, (str, pathlib.Path)):
---> 72         return load_custom_array(file, create_numpy_array)
     73     else:
     74         # assume we have a file-like object
     75         buffer = file.read()

File /opt/anaconda3/envs/metatensor-dev/lib/python3.11/site-packages/metatensor/io.py:118, in load_custom_array(path, create_array)
    114     path = bytes(path)
    116 ptr = lib.mts_tensormap_load(path, mts_create_array_callback_t(create_array))
--> 118 return TensorMap._from_ptr(ptr)

File /opt/anaconda3/envs/metatensor-dev/lib/python3.11/site-packages/metatensor/tensor.py:121, in TensorMap._from_ptr(ptr)
    118 @staticmethod
    119 def _from_ptr(ptr):
    120     """Create a tensor map from a pointer owning its data"""
--> 121     _check_pointer(ptr)
    122     obj = TensorMap.__new__(TensorMap)
    123     obj._lib = _get_library()

File /opt/anaconda3/envs/metatensor-dev/lib/python3.11/site-packages/metatensor/status.py:48, in _check_pointer(pointer)
     46     raise MetatensorError(last_error()) from e
     47 else:
---> 48     raise MetatensorError(last_error())

MetatensorError: serialization format error: unsupported Zip archive: Compression method not supported: at 'keys.npy'
```
Correct usage 
```python
metatensor.load("./python/metatensor-operations/tests/data/qm7-power-spectrum.npz", use_numpy=True)
```
But I cannot understand from the first error message that this is the issue. I am not sure what the right solution would be, but anything would help. For example, checking the file ending and add something to the error message if `npz` is identified (or just always add this to the message). Or if this fails trying to also load it with `use_numpy` option to give user more informative error message.

Luthaf:
This does not change file saving, we are still only using STORED in this case.

Fix #469.

This is mainly intended to remove the surprising behavior as seen in the issue, where a file that should be loaded can only be loaded through numpy and not metatensor-core. Such files should be rare, since one has to go out of their way to create it (saving with metatensor/numpy, then loading as a normal npz and re-saving with different settings).

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1663633276.zip)

⚙️ [Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1664257799.zip)

### DIFF
--- a/python/metatensor-core/tests/serialization.py
+++ b/python/metatensor-core/tests/serialization.py
@@ -80,6 +80,30 @@ def test_load(use_numpy, memory_buffer, standalone_fn):
     assert gradient.values.shape == (59, 3, 5, 3)
 
 
+@pytest.mark.parametrize("use_numpy", (True, False))
+def test_load_deflate(use_numpy):
+    # This file was saved using DEFLATE to compress the different ZIP archive members
+    path = os.path.join(
+        os.path.dirname(__file__),
+        "..",
+        "..",
+        "metatensor-operations",
+        "tests",
+        "data",
+        "qm7-power-spectrum.npz",
+    )
+
+    tensor = metatensor.load(path, use_numpy=use_numpy)
+
+    assert isinstance(tensor, TensorMap)
+    assert tensor.keys.names == [
+        "center_type",
+        "neighbor_1_type",
+        "neighbor_2_type",
+    ]
+    assert len(tensor.keys) == 17
+
+
 # using tmpdir as pytest-built-in fixture
 # https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures
 @pytest.mark.parametrize("use_numpy", (True, False))
--- a/metatensor-core/Cargo.toml
+++ b/metatensor-core/Cargo.toml
@@ -24,7 +24,7 @@ smallvec = {version = "1", features = ["union"]}
 # implementation of the NPZ serialization format
 byteorder = {version = "1"}
 num-traits = {version = "0.2", default-features = false}
-zip = {version = "0.6", default-features = false}
+zip = {version = "0.6", default-features = false, features = ["deflate"]}
 
 [build-dependencies]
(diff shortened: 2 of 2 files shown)


python/metatensor-core/tests/serialization.py
```diff
@@ -80,6 +80,30 @@ def test_load(use_numpy, memory_buffer, standalone_fn):
     assert gradient.values.shape == (59, 3, 5, 3)
 
 
+@pytest.mark.parametrize("use_numpy", (True, False))
+def test_load_deflate(use_numpy):
+    # This file was saved using DEFLATE to compress the different ZIP archive members
```
PicoCentauri, line 85: I am just wondering how we were able to load this file before without supporting _deflate_?

Luthaf: we could only load it with `use_numpy=True`, which does support compression. Instead of this test, I could also do `use_numpy=False` when loading data in operations tests. What do you think?

PicoCentauri: Yes makes sense. Now I see the parametrize.

PicoCentauri (approved):

PR #673, opened 2024-07-04
Serialization for `System`

frostedoyster:
Implements serialization for `metatensor.torch.atomistic.System`

# Reviewer checklist

### DIFF
--- a/metatensor-torch/include/metatensor/torch/atomistic/system.hpp
+++ b/metatensor-torch/include/metatensor/torch/atomistic/system.hpp
@@ -255,6 +255,12 @@ public:
     /// Implementation of `__str__` and `__repr__` for Python
     std::string str() const;
 
+    /// Serialize a `System` to a JSON string.
+    std::string to_json() const;
+
+    /// Load a serialized `System` from a JSON string.
+    static System from_json(std::string_view json);
+
 private:
     struct nl_options_compare {
         bool operator()(const NeighborListOptions& a, const NeighborListOptions& b) const {
--- a/metatensor-torch/src/atomistic/system.cpp
+++ b/metatensor-torch/src/atomistic/system.cpp
@@ -789,3 +789,104 @@ std::string SystemHolder::str() const {
 
     return result.str();
 }
+
+template<typename scalar_t>
+nlohmann::json tensor_to_vector_string(const torch::Tensor& tensor) {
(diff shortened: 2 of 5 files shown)


frostedoyster:
@Luthaf is this what you were thinking of?
If not, does torch provide something like Tensor to json and back?

Luthaf:
I would rather not use JSON for this, or at the very least only use JSON for the container and use a binary serialization format for the data (i.e. the torch Tensor and the TensorBlock inside the system).

We should be able to access whatever format torch is using internally to save Tensor, and use serialization to a buffer to save metatensor data.

For the container format, JSON is fine, but not ideal. The overall thing might look like

```json
{
    "positions": "VERY LONG BINARY STRING ...6251690352651719",
    "cell": "VERY LONG BINARY STRING ...6251690352651719",
    "types": "VERY LONG BINARY STRING ...6251690352651719",
    "neighbors": [
        {"options": "JSON_OF_OPTIONS", "values": "VERY LONG BINARY STRING ...6251690352651719"},
        {"options": "JSON_OF_OPTIONS", "values": "VERY LONG BINARY STRING ...6251690352651719"}
    ],
    "data": {
        "name-1": "VERY LONG BINARY STRING ...6251690352651719",
        "name-2": "VERY LONG BINARY STRING ...6251690352651719",
    }
}
```

The main problem then is that encoding a binary string to JSON might require some tranformations of the string to escape some of them.

An alternative would be to go for [ubjson](https://en.wikipedia.org/wiki/UBJSON), which should be supported by our JSON library. We might need to be careful to make sure the serialization uses a proper binary format to store our binary data (they seem to use list of integers in this case). The main drawback here is that we would now have another, different serialization format.

frostedoyster:
Hmm ok, in that case I think it's better if you do it because it looks complicated to me.

Another possibility is to merge something like this for now and change it later. The main use of this would be loading from disk for a large dataset and, although this approach is potentially slow, I'm hoping that it will still be faster than the model that is being trained (data loading can happen asynchronously and in parallel, see `num_workers` and related for the torch Dataloader)

Luthaf:
After thinking more about it, I think System serialization should use the same format as the rest of metatensor, i.e. npz (ZIP of npy files). There are a couple of implementation for both the ZIP and npy formats in C++, and we could even use a manual implementation for the NPY part.

Luthaf:
Replaced by #797

PR #674, opened 2024-07-05
Fix documentation for printing `ModelMetadata`

frostedoyster:
The documentation at https://docs.metatensor.org/latest/atomistic/reference/models/metadata.html
says there is a `ModelMetadata.print()` method, but it doesn't work at the moment because it's not registered

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1672240911.zip)

### DIFF
--- a/python/metatensor-torch/metatensor/torch/atomistic/documentation.py
+++ b/python/metatensor-torch/metatensor/torch/atomistic/documentation.py
@@ -385,7 +385,12 @@ class ModelEvaluationOptions:
 
 
 class ModelMetadata:
-    """Metadata about a specific exported model"""
+    """
+    Metadata about a specific exported model
+
+    This class implements the __str__ and __repr__ methods, so its representation can be
+    easily printed, logged, inserted into other strings, etc.
+    """
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


frostedoyster:
IDK, works locally

metatensor-torch/src/register.cpp
```diff
@@ -382,6 +382,7 @@ TORCH_LIBRARY(metatensor, m) {
         )
         .def("__repr__", &ModelMetadataHolder::print)
         .def("__str__", &ModelMetadataHolder::print)
+        .def("print", &ModelMetadataHolder::print)
```
Luthaf, line 385: this is registered as `__str__`; so you can do `print(metadata)` from Python and you don't need to do `print(metadata.print())`

frostedoyster: I know, but then the docs are wrong (?)

frostedoyster: I just wanted to print, looked up the docs, did `print()`, didn't work. Should we change the Python docs?

Luthaf: yes!

Luthaf (approved):

PR #692, opened 2024-07-16
Implement Labels selection

### LINKED ISSUE #97: Advanced indexing of tensor blocks
Adding index access functionalities which are inspired by numpy
```python
indices = np.where(X == something)[0]
X[indices]
```
The position function already returns one index
```python
index = tensor_block.properties.position((0,1))
```
But something in a similar spirit as the `blocks_matching` function would be nice
```python
indices = tensor_block.properties.position(Labels(names=["n"], values=[[0]]) # returns a tensor block with all properties that fulfill n=0 in the Labels
```
Already talked with @Luthaf about this. Here details from him which might be important for an implementation

> The function is called from here, with a Labels for selection https://github.com/lab-cosmo/equistore/blob/e65cd07d2f0d5aefa41a7b2beade835b9b1dc03b/python/src/equistore/tensor.py#L266-L271
(i.e. to get all n=0, you would give Labels(names=["n"], values=[[0]])
It ends up here after passing through C: https://github.com/lab-cosmo/equistore/blob/e65cd07d2f0d5aefa41a7b2beade835b9b1dc03b/src/tensor/mod.rs#L190, so this is the function we could extract out of the TensorMap and make available for arbitrary Labels (something like Labels::matching(Labels selection).
We are also doing something similar in rascaline, so standardizing everything in equistore would be nice

### LINKED ISSUE #700: `slice_block` is slow when operating on very large set of Labels
In the example below, using this data file  [iron-snapshot.xyz.txt](https://github.com/user-attachments/files/16415631/iron-snapshot.xyz.txt)

Case (A) takes 6.8s, and case (B) takes 360ms!

```py
import rascaline
import metatensor as mts

import numpy as np
import ase.io

frame_iron = ase.io.read("iron-snapshot.xyz", 0)

# selects a subset of the atoms to speed up and remove memory requirements
crop_dz = 16 
crop_idx = np.where(
    (frame_iron.positions[:, 2] < frame_iron.cell[2, 2] / 2 + crop_dz / 2)
    & (frame_iron.positions[:, 2] > frame_iron.cell[2, 2] / 2 - crop_dz / 2)
)[0]
frame_iron = frame_iron[crop_idx]

core_dz = 4
max_cutoff = 6
core_idx = np.sort(
    np.where(
        (frame_iron.positions[:, 0] > max_cutoff + 1)
        & (frame_iron.positions[:, 0] < 199 - max_cutoff)
        & (frame_iron.positions[:, 1] > max_cutoff + 1)
        & (frame_iron.positions[:, 1] < 199 - max_cutoff)
        & (frame_iron.positions[:, 2] < frame_iron.cell[2, 2] / 2 + core_dz / 2)
        & (frame_iron.positions[:, 2] > frame_iron.cell[2, 2] / 2 - core_dz / 2)
    )[0]
)
frame_core = frame_iron[core_idx]

# compute the full NL
nl_code = rascaline.NeighborList(cutoff=max_cutoff, full_neighbor_list=True)
nl_all = nl_code.compute(frame_iron)

# ==================== (A) select with slice_block
# %%time
nl_selected = mts.slice_block(
    nl_all.block(0),
    axis="samples",
    labels=mts.Labels(names=["first_atom"], values=core_idx[:,np.newaxis])
)

# ==================== (B) manual selection
# %%time
labs = core_idx
samp = np.asarray(nl_all.block(0).samples["first_atom"])
mask = np.zeros(len(samp), dtype=bool)

# sorting order and reverse order
sort_idx = np.argsort(samp)
sort_inv = np.argsort(sort_idx)
sort_samp = samp[sort_idx]

# now blocks with the same atom are contiguous so we can use searchsorted to find the bounds
lower = np.searchsorted(sort_samp, labs)
upper = np.searchsorted(sort_samp, labs + 1)


for l, u in zip(lower, upper):
    mask[l:u] = True
mask = mask[sort_inv]

# actually we better extract a classical NL
neigh_dx = nl_all.block(0).values[sort_idx][mask].squeeze()
neigh_sz = upper - lower
neigh_i = np.cumsum(neigh_sz) - neigh_sz[0]
neigh_dr = np.sqrt((neigh_dx**2).sum(axis=1))
```

#692 might help to make the selection faster, but barring that we could also implement a special case in `slice_block` when there is a single dimension in selected labels.

This was found by @ceriottm while working on https://github.com/ceriottm/aml-intro/.

Luthaf:
This function enables the selection of a subset of entries in Labels using another Labels to define the selection. Mulitple entries in the selection are interpreted as an `or` operation.

Fix #97 

Fix #700 as well, bringing the time for `slice_block` in the issue example to ~380ms, very similar to the manual implementation.

Still TBD: 

- [x] check the operations for opportunities to use the new function
- [ ] check the learn classes for opportunities to use the new function
- [x] check if we can replace the manual implementation in rascaline with this one

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1811714225.zip)

### DIFF
--- a/python/metatensor-core/metatensor/labels.py
+++ b/python/metatensor-core/metatensor/labels.py
@@ -6,7 +6,7 @@ from typing import BinaryIO, List, Optional, Sequence, Tuple, Union, overload
 
 import numpy as np
 
-from ._c_api import mts_labels_t
+from ._c_api import c_uintptr_t, mts_labels_t
 from ._c_lib import _get_library
 from .utils import _ptr_to_const_ndarray
 
@@ -955,6 +955,44 @@ class Labels:
         )
 
         return Labels._from_mts_labels_t(output), first_mapping, second_mapping
+
+    def select(self, selection: "Labels") -> np.ndarray:
+        """
+        Select entries in these :py:class:`Labels` that match the ``selection``.
+
+        The selection's names must be a subset of the names of these labels.
+
+        All entries in these :py:class:`Labels` that match one of the entry in the
+        ``selection`` for all the selection's dimension will be picked. Any entry in the
+        ``selection`` but not in these :py:class:`Labels` will be ignored.
+
+        >>> import numpy as np
+        >>> from metatensor import Labels
+        >>> labels = Labels(
+        ...     names=["a", "b"],
+        ...     values=np.array([[0, 1], [1, 2], [0, 3], [1, 1], [2, 4]]),
+        ... )
+        >>> selection = Labels(names=["a"], values=np.array([[0], [2], [5]]))
+        >>> print(labels.select(selection))
+        [0 2 4]
+
+        :param selection: description of the entries to select
+        :return: 1-dimensional ndarray containing the integer indices of selected
+            entries
+        """
+        selected = np.zeros((len(self)), dtype=np.int64)
+        selected_count = c_uintptr_t(len(self))
+
+        self._lib.mts_labels_select(
+            self._as_mts_labels_t(),
+            selection._as_mts_labels_t(),
+            selected.ctypes.data_as(ctypes.POINTER(ctypes.c_int64)),
+            ctypes.pointer(selected_count),
+        )
+
+        selected.resize(selected_count.value, refcheck=False)
+
+        return selected
 
     def print(self, max_entries: int, indent: int = 0) -> str:
         """print these :py:class:`Labels` to a string
--- a/python/metatensor-torch/metatensor/torch/documentation.py
+++ b/python/metatensor-torch/metatensor/torch/documentation.py
@@ -579,6 +579,31 @@ class Labels:
             used in the output, the mapping for them is set to ``-1``.
         """
 
+    def select(self, selection: "Labels") -> torch.Tensor:
+        """
+        Select entries in these :py:class:`Labels` that match the ``selection``.
+
+        The selection's names must be a subset of the names of these labels.
+
+        All entries in these :py:class:`Labels` that match one of the entry in the
+        ``selection`` for all the selection's dimension will be picked. Any entry in the
+        ``selection`` but not in these :py:class:`Labels` will be ignored.
+
+        >>> import torch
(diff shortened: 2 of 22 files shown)


python/metatensor-core/metatensor/labels.py
```diff
@@ -956,6 +956,34 @@ def intersection_and_mapping(
...
+
+        The selection's names must be a subset of the names of these labels.
+
+        All entries in these :py:class:`Labels` that match one of the entry in the
+        ``selection`` for all the selection's dimension will be picked. Any entry in the
+        ``selection`` but not in these :py:class:`Labels` will be ignored.
+
+        :param selection: description of the entries to select
+        :return: 1-dimensional ndarray containing the integer indices of selected
+            entries
```
PicoCentauri, line 971: Maybe add an example as for the other methods.

python/metatensor-torch/metatensor/torch/documentation.py
```diff
@@ -579,6 +579,21 @@ def intersection_and_mapping(
...
+
+        The selection's names must be a subset of the names of these labels.
+
+        All entries in these :py:class:`Labels` that match one of the entry in the
+        ``selection`` for all the selection's dimension will be picked. Any entry in the
+        ``selection`` but not in these :py:class:`Labels` will be ignored.
+
+        :param selection: description of the entries to select
+        :return: 1-dimensional tensor containing the integer indices of selected
+            entries
```
PicoCentauri, line 594: As above: maybe an example could be useful.

python/metatensor-core/metatensor/labels.py
Luthaf: Done!

python/metatensor-torch/metatensor/torch/documentation.py
Luthaf: Done

PR #695, opened 2024-07-18
Do not use system packages when building Python wheels

Luthaf:
We don't want our wheels to mistakenly find and use a global torch (or any other dependency).

# Reviewer checklist


Luthaf:
Unfortunately this also breaks the code that find `make`, and I can't find a way to only apply it to `find_package`.

PR #703, opened 2024-07-30
Add an atomistic example with neighbor list

### LINKED ISSUE #701: Add metatensor-atomistic example with neighborlist
We have a good example explaining how to create a model and export one. However, we don't show how to request and use a neighborlist (using `requested_neighbor_lists()`) in any of our example. The use is explained in the docs but I think an example with a neighborlist should help.

The example basically boils down to a code block like

```python
class Model
    def __init__(self):
        self.nl = NeighborListOptions(...)

    def requested_neighbor_lists(self):
        return [self.nl]

    def forward(self, systems, ...):
        neighbors = systems[0].get_neighbor_list(self.nl)
```

and maybe running this code somehow.

PicoCentauri:
Fixes #701 

As given in the title this PR adds an example of a metatensor atomistic model using a neighbor list. The model runs a liquid argon system based on Lennard-Jones interactions. In the end of the example, we reproduce the RDF of the system.

To speed up the computation we are not creating the system from scratch but I load an equilibrated system from file.

 - ~[ ] Tests updated (for new features and bugfixes)?~
 - ~[ ] Issue referenced (for PRs that solve an issue)?~

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1815303817.zip)


python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,300 @@
+r"""
+.. _atomistic-tutorial-nl-md:
+
+Running Molecular Dynamics with a Neighbor List (NL)
+====================================================
```
Luthaf, line 5:
```suggestion
Creating models that use neighbor lists
=======================================
```

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,300 @@
...
+
+.. py:currentmodule:: metatensor.torch.atomistic
+
+This tutorial demonstrates how to use metatensor atomistic models to run a Molecular
+Dynamics (MD) simulation using ASE in combination with a neighbor list.
```
Luthaf, line 10:
```suggestion
This tutorial demonstrates how to create an atomistic model that requires
neighbor lists, and use it to run MD simulations. When using such model to run
simulations, the neighbor lists will be provided by the simulation engine, so
the model does not have to compute them.
```

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,300 @@
...
+
+# tools for analysis and visualization
+import ase.geometry.rdf
+
+# tools to run the simulation
+import ase.io
+import ase.md
+import ase.visualize.plot
+import chemiscope
+import matplotlib.pyplot as plt
```
Luthaf, line 34: should these be moved to the analysis/visualization section?

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,300 @@
...
+# between a pair of neutral atoms or molecules. It is given by the equation:
+#
+# .. math:: V(r) = 4\epsilon \left[ \left( \frac{\sigma}{r} \right)^{12} - \left(
+#     \frac{\sigma}{r} \right)^{6} \right]
+#
+# where :math:`\epsilon` is the depth of the potential well, :math:`\sigma` is the
+# finite distance at which the inter-particle potential is zero, and :math:`r` is the
+# distance between the particles. The 12-6 form is chosen because the :math:`r^12` term
+# approximates the Pauli repulsion at short ranges, while the :math:`r^6` term
+# represents the attractive van der Waals forces. The :math:`r^{12}` is choosen becuase
```
Luthaf, line 69:
```suggestion
# represents the attractive van der Waals forces. The :math:`r^{12}` is chosen because
```

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,300 @@
...
+# it is just the square of the :math:`r^6` and therefore allows fast evaluation.
+#
+# A Lennard-Jones potential is well-suited for argon because it accurately represents
+# the van der Waals forces that dominate the interactions between argon atoms especially
+# as for all noble gases. This potential was used in one of the first MD simulations:
```
Luthaf, line 74:
```suggestion
# A Lennard-Jones potential is well-suited for argon because it accurately represents
# the van der Waals forces that dominate the interactions between argon atoms,
# as for all noble gases. This potential was used in one of the first MD simulations:
```

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,300 @@
...
+# The model below is a simplified version of a `more complex Lennard-Jones model
+# <https://github.com/Luthaf/metatensor-lj-test/blob/main/src/metatensor_lj_test/pure.py>`_.
+# The linked version also implements ``per_atom`` energies as well as atom selection
+# using the ``selected_atoms`` parameter of the ``forward()`` method. In this model, we
+# shift the energy by its value at the ``cutoff``. This will break the conservativeness
+# of the potential, which is unproblematic in here because we use a large cutoff and
+# therefore the potential already almost decayed to zero. For more sophisticated methods
+# like a polynomial switching potential like XPLOR, we refer to the literature.
+
+
```
Luthaf, line 88: before showing an implementation of the model, I would try to explain a bit more how the data is stored in neighbors, and how to use it.

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,300 @@
...
+        super().__init__()
+        self._nl_options = NeighborListOptions(cutoff=cutoff, full_list=False)
+
+        self._sigma = sigma
+        self._epsilon = epsilon
+
+        # shift the energy to 0 at the cutoff
+        self._shift = 4 * epsilon * ((sigma / cutoff) ** 12 - (sigma / cutoff) ** 6)
+
+    def requested_neighbor_lists(self) -> List[NeighborListOptions]:
```
Luthaf, line 102: I would add a bunch of docs around this function to explain what it is doing/how it is used

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,300 @@
...
+            distances = neighbors.values.reshape(-1, 3)
+
+            sigma_r_6 = (self._sigma / torch.linalg.vector_norm(distances, dim=1)) ** 6
+            sigma_r_12 = sigma_r_6 * sigma_r_6
+            e = 4.0 * self._epsilon * (sigma_r_12 - sigma_r_6) - self._shift
+
+            # We only compute each pair once (full_list=False in self._nl_options),
+            # and assign half of the energy to each atom
+            energy = energy.index_add(0, all_i, e, alpha=0.5)
+            energy = energy.index_add(0, all_j, e, alpha=0.5)
```
Luthaf, line 147: I don't think we need this if we are not doing `per_atom`. We can directly sum `e` and store that.

Luthaf: Or we keep the `per_atom` option and use it to explain the data in the samples?

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,300 @@
...
+# %%
+#
+# Running the simulation
+# ----------------------
+#
+# The final step to run the simulation is to register our model as the energy calculator
+# for these ``atoms``. This is the job of
+# :py:class:`ase_calculator.MetatensorCalculator`, which takes either an instance of
+# :py:class:`MetatensorAtomisticModel` or the path to a pre-exported model, and allows
+# it to compute the energy, forces, and stress acting on a system.
```
Luthaf, line 216: Do we want to repeat this? Since this was explained in the previous tutorial.

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,300 @@
...
+#
+# We now define and wrap the model, using the initial positions and the Lennard-Jones
+# parameters taken from `Méndez-Bermúdez et.al, Phys. Commun. 2022
+# <https://dx.doi.org/10.1088/2399-6528/ac62a2>`_.
```
Luthaf, line 220: we could add a small sentence on units & making sure to declare the right ones in the capabilities

Luthaf (commented):
Thanks! I think we can add more information about how NL are computed (maybe adding a small diagram showing who is computing what and storing it where) and how to use the data stored inside.

PicoCentauri:
Thanks for the comment.

> maybe adding a small diagram showing who is computing what and storing it where

You mean the engine is computing the nl is giving us the data? Similar to https://docs.metatensor.org/latest/atomistic/overview.html#data-flow-between-the-model-and-engine or what kind of diagram do you have in mind?

python/examples/atomistic/3-atomistic-model-with-nl.py
PicoCentauri: Good point. I will add a section Metatensor Neighbor lists before this one.

python/examples/atomistic/3-atomistic-model-with-nl.py
PicoCentauri: I think we can leave it out. It is not important for ase. But it is for lammps, right?

Luthaf: no, it's fine (the user choses to request per-atom energies or not). What's important for LAMMPS is selected_atoms since we always have to subselect the real atoms from the ghosts.

Luthaf:
> Similar to https://docs.metatensor.org/latest/atomistic/overview.html#data-flow-between-the-model-and-engine or what kind of diagram do you have in mind?

Yes, something similar to this diagram, but focused on only the NL calculation part

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,433 @@
...
+# %%
+#
+# The :py:func:`neighbor_list <ase.neighborlist.neighbor_list>` function returns the
+# neighbor indices: quantities ``"i"`` and ``"j"``, the neighbor shifts: ``"S"``, and
+# the distance vectors: ``"D"``. We now stack these together and convert them into the
+# suitable types.
+
+i = torch.from_numpy(i.astype(int))
+j = torch.from_numpy(j.astype(int))
+neighbor_indices = torch.vstack([i, j])
```
Luthaf, line 120: Can we make this a `n x 2` array instead of `2 x n`? Just to follow metatensor convention from the start

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,433 @@
...
+neighbor_shifts = torch.from_numpy(S.astype(int))
+
+print("neighbor_indices:", neighbor_indices)
+print("neighbor_shifts:", neighbor_shifts)
+
+# %%
+# Create a Neighbor List
+# ^^^^^^^^^^^^^^^^^^^^^^
```
Luthaf, line 128:
```suggestion
# Creating a Neighbor List
# ^^^^^^^^^^^^^^^^^^^^^^^^
```

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,433 @@
...
+print(neighbors)
+
+# %%
+#
+# The data and metadata inside the ``neighbors`` object do not contain information about
+# the ``cutoff`` and whether this is a full or half neighbor list. To account for this,
+# metatensor neighbor lists are always stored together with
+# :py:class:`NeighborListOptions`. For our system, these options can be saved as
+
+options = NeighborListOptions(cutoff=5.0, full_list=True, requestor="ljmodel")
```
Luthaf, line 162: I would not put any `requestor` here, it is an advanced tool that most people don't need to know about.

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,433 @@
...
+
+# %%
+#
+# We set ``full_list=True`` because ASE computes a “full” list (as opposed to a half
+# list) where each `i, j` pair appears twice, stored once with `i` and once with `j`.
```
Luthaf, line 167:
```suggestion
# list) where each `i, j` pair appears twice, stored once as `i, j` and once as `j, i`.
```

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,433 @@
...
+# The ``requestor`` parameter is optional and can be used to further distinguish
+# neighbor lists if several are present.
```
Luthaf, line 169: No, this is just recording who requested a given NL, but two NeighborListOptions with identical cutoff/full_list but different requestor will compare as equal. I would not mention this here

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,433 @@
...
+#
+#   system.add_neighbor_list(options=options, neighbors=neighbors)
+#
+# .. note::
+#
+#   The steps above of creating a neighbor list and wrapping it inside a
+#   :py:class:`TensorBlock <metatensor.torch.TensorBlock>` and attaching it to a system
+#   can be done by the metatensor molecular dynamics interface and are not required by
+#   the model developer. How to request a neighbor list will be presented below when the
+#   actual model is defined.
```
Luthaf, line 183: Should we move this note on top of the section?

```suggestion
# .. note::
#
#   The steps above of creating a neighbor list and wrapping it inside a
#   :py:class:`TensorBlock <metatensor.torch.TensorBlock>` and attaching it to a system
#   will be done by the simulation engine and should not be handled by
#   the model developer. How to request a neighbor list will be presented below when the
#   actual model is defined.
```

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,433 @@
...
+# You can also loop over all attached lists of a :py:class:`System` using
+# :py:meth:`System.known_neighbor_lists` to find a suitable one based on
+# :py:class:`NeighborListOptions` attributes like ``cutoff``, ``full_list``, and
+# ``requestor``.
+#
+# After the neighbor list is attached, we can access the data and metadata. First, we
+# extract the ``distances`` between the neighboring pairs within the cutoff, which we
+# will need for our model below.
+
+distances = neighbors.values
```
Luthaf, line 205: maybe 

```suggestion
distances = neighbors.values

print(distances.shape)
```

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -0,0 +1,433 @@
...
+# use it we can construct our Lennard-Jones model with a fixed cutoff. The Lennard-Jones
+# potential is a mathematical basis to approximate the interaction between a pair of
+# neutral atoms or molecules. It is given by the equation:
+#
+# .. math:: V(r) = 4\epsilon \left[ \left( \frac{\sigma}{r} \right)^{12} - \left(
+#     \frac{\sigma}{r} \right)^{6} \right]
+#
+# where :math:`\epsilon` is the depth of the potential well, :math:`\sigma` is the
+# finite distance at which the inter-particle potential is zero, and :math:`r` is the
+# distance between the particles. The 12-6 form is chosen because the :math:`r^12` term
```
Luthaf, line 237:
```suggestion
# distance between the particles. The 12-6 form is chosen because the :math:`r^{12}` term
```

PicoCentauri:
Thanks for the second round of review @Luthaf. I hope, I implemented all comments.

Luthaf:
I've added a couple more tweaks to the tutorial, let me know if these are fine for you!

PR #705, opened 2024-07-30
Fixes for Torch 2.4

Luthaf:
Fix the build with the new torch release and build metatensor-torch wheels for this version.

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1811697448.zip)

⚙️ [Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1756405490.zip)

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -22,16 +22,16 @@ jobs:
             torch-version: "1.12.*"
           - os: ubuntu-20.04
             python-version: "3.9"
-            torch-version: "2.3.*"
+            torch-version: "2.4.*"
           - os: ubuntu-20.04
             python-version: "3.12"
-            torch-version: "2.3.*"
+            torch-version: "2.4.*"
           - os: macos-14
             python-version: "3.12"
-            torch-version: "2.3.*"
+            torch-version: "2.4.*"
           - os: windows-2019
             python-version: "3.12"
-            torch-version: "2.3.*"
+            torch-version: "2.4.*"
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


.github/workflows/build-wheels.yml
```diff
@@ -120,11 +120,13 @@ jobs:
           - {os: windows-2019, arch: arm64}
           # https://github.com/pytorch/pytorch/issues/125109
           - {os: windows-2019, torch-version: '2.3'}
+          - {os: windows-2019, torch-version: '2.4'}
           # arm64-macos is only supported for torch >= 2.0
           - {os: macos-14, arch: arm64, torch-version: '1.12'}
           - {os: macos-14, arch: arm64, torch-version: '1.13'}
           # x86_64-macos is only supported for torch <2.3
           - {os: macos-13, arch: x86_64, torch-version: '2.3'}
+          - {os: macos-13, arch: x86_64, torch-version: '2.4'}
```
PicoCentauri, line 129: Is the origin of excluding macos the same as the issue for windows which is linked above?

Luthaf: no, it's that PyTorch no longer distribute Intel macOS builds (there is a comment just above these two lines, but I can try to make it clearer!)

PicoCentauri: It is fine but maybe bit clearer could be something like

`# arch x86_64 on macos is only supported for torch <2.3`

PicoCentauri (approved):

PR #708, opened 2024-08-01
Allow for extra metadata in atomistic models

frostedoyster:
As in the title.
Should we print this extra metadata with the other metadata or not?
Should we implement a getter? (The other fields don't have it though)

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1766765529.zip)

### DIFF
--- a/metatensor-torch/include/metatensor/torch/atomistic/model.hpp
+++ b/metatensor-torch/include/metatensor/torch/atomistic/model.hpp
@@ -243,12 +243,14 @@ public:
         std::string name_,
         std::string description_,
         std::vector<std::string> authors_,
-        torch::Dict<std::string, std::vector<std::string>> references_
+        torch::Dict<std::string, std::vector<std::string>> references_,
+        torch::Dict<std::string, std::string> extra_metadata_
     ):
         name(std::move(name_)),
         description(std::move(description_)),
         authors(std::move(authors_)),
-        references(references_)
+        references(references_),
+        extra_metadata(extra_metadata_)
     {
         this->validate();
     }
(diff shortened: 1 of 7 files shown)


Luthaf:
> Should we print this extra metadata with the other metadata or not?

I would not, since we don't know the best way to print this.

> Should we implement a getter? (The other fields don't have it though)

Just having this as a read-write property should be fine!

metatensor-torch/include/metatensor/torch/atomistic/model.hpp
```diff
@@ -274,6 +276,9 @@ class METATENSOR_TORCH_EXPORT ModelMetadataHolder: public torch::CustomClassHold
     /// - "model": for reference specific to this exact model
     torch::Dict<std::string, std::vector<std::string>> references;
 
+    /// Extra metadata about this model. This can be anything.
```
Luthaf, line 279:
```suggestion
    /// Extra metadata about this model. This can be anything, and is intended to be used by 
    /// tools based on metatensor to store data they need.
```

metatensor-torch/include/metatensor/torch/atomistic/model.hpp
```diff
@@ -274,6 +276,9 @@ class METATENSOR_TORCH_EXPORT ModelMetadataHolder: public torch::CustomClassHold
...
+    torch::Dict<std::string, std::string> extra_metadata;
```
Luthaf, line 280: you will also need to add this to `ModelMetadata` serialization

Luthaf: Nitpick: should this be named `extra`? Since it will be accessed as `metadata.extra["something"]`, which feels nicer than `metadata.extra_metadata["something"]`

metatensor-torch/include/metatensor/torch/atomistic/model.hpp
frostedoyster: Perfect, I also updated the Python docs

metatensor-torch/include/metatensor/torch/atomistic/model.hpp
frostedoyster: Whoops, I forgot. And yes, changed to `extra`

Luthaf (approved):

PR #712, opened 2024-08-09
Fix incompatible arguments of torch::throw()

RudolfWeeber:
I had to apply this patch to mcomplie metatensor_torch.
IndexError expects a source location as first argument.

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1795965283.zip)


Luthaf:
Thanks for the PR! This is also what #705 is trying to fix, but there is a build failure on Windows there. In the mean time you should be able to use torch 2.3, since the build failure is specific to torch 2.4.

Luthaf:
This should now be fixed on master!

PR #718, opened 2024-08-26
Features standard output

sofiia-chorna:
Added a new standard output "features" to return the internal or learned features from a model (model's intermediate values before final predictions). The output format checks were taken from the Plumed [doc](https://www.plumed.org/doc-master/user-doc/html/_m_e_t_a_t_e_n_s_o_r.html).

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/lab-cosmo/metatensor/actions/artifacts/1874675914.zip)

### DIFF
--- a/docs/src/atomistic/outputs.rst
+++ b/docs/src/atomistic/outputs.rst
@@ -167,3 +167,65 @@ Energy ensemble gradients
 
 The gradient metadata for energy ensemble is the same as for the ``energy``
 output (see `Energy gradients`_).
+
+.. _features:
+
+Features
+^^^^^^^^
+
+Features are numerical vectors representing a given structure or
+atom/atom-centered environment in an abstract n-dimensional space. They are also
+sometimes called descriptors, representations, embedding, *etc.*
+
+Features can be computed with some analytical expression (for example `SOAP
+power spectrum`_, `atom-centered symmetry functions`_, …), or learned internally
+by a neural-network or a similar architecture.
+
+.. _SOAP power spectrum: https://doi.org/10.1103/PhysRevB.87.184115
+.. _Atom-centered symmetry functions: https://doi.org/10.1063/1.3553717
+
+In metatensor atomistic models, they are associated with the ``"features"`` key
+in the model outputs, and must adhere to the following metadata:
+
+.. list-table:: Metadata for features output
+  :widths: 2 3 7
+  :header-rows: 1
+
+  * - Metadata
+    - Names
+    - Description
+
+  * - keys
+    - ``"_"``
+    - the features keys must have a single dimension named ``"_"``, with a single
+      entry set to ``0``. The feature is always a
+      :py:class:`metatensor.torch.TensorMap` with a single block.
+
+  * - samples
+    - ``["system", "atom"]`` or ``["system"]``
+    - the samples should be named ``["system", "atom"]`` for per-atom outputs;
+      or ``["system"]`` for global outputs.
+
+      The ``"system"`` index should always be 0, and the ``"atom"`` index should
+      be the index of the atom (between 0 and the total number of atoms). If
+      ``selected_atoms`` is provided, then only the selected atoms for each
+      system should be part of the samples.
+
+  * - components
+    -
+    - the features must not have any components.
+
+  * - properties
+    -
+    - the features can have arbitrary properties.
+
+.. note::
+  Features are typically handled without a unit, so the ``"unit"`` field of
+  :py:func:`metatensor.torch.atomistic.ModelOutput` is mainly left empty.
+
+Features gradients
+------------------
+
+As for the :ref:`energy <energy-gradients>`, features are typically used with
+automatic gradient differentiation. Explicit gradients could be allowed if you
+have a use case for them, but are currently not until they are fully specified.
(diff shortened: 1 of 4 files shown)


frostedoyster:
It looks good to me! Is this always meant to be used without units? Perhaps it would be worth to add this to the docs

sofiia-chorna:
> It looks good to me! Is this always meant to be used without units? If so, perhaps it would be worth adding it to the docs

Thanks for checking! I added a note regarding the units.

python/metatensor-torch/metatensor/torch/atomistic/outputs.py
```diff
@@ -202,3 +142,107 @@ def _check_energy_like(
...
+    """
+    Ensure the TensorMap has a single block with the expected key
+    """
+    if value.keys != Labels("_", torch.tensor([[0]])):
+        raise ValueError(
+            f"invalid keys for '{name}' output: expected `Labels('_', [[0]])`"
+        )
+
+
+def _validate_components(name: str, block: TensorBlock):
```
Luthaf, line 241:
```suggestion
def _validate_no_components(name: str, block: TensorBlock):
```

python/metatensor-torch/metatensor/torch/atomistic/outputs.py
```diff
@@ -202,3 +142,107 @@ def _check_energy_like(
...
+    # Should not have any explicit gradients
+    # all gradient calculations are done using autograd
+    if len(features_block.gradients_list()) > 0:
+        raise ValueError(
+            "invalid gradients for 'features' output: it should not have any explicit",
+            "gradients. all gradient calculations should be done using autograd",
+        )
+
+
+def _validate_samples(
```
Luthaf, line 177:
```suggestion
def _validate_atomic_samples(
```

docs/src/atomistic/outputs.rst
```diff
@@ -167,3 +167,43 @@ Energy ensemble gradients
...
+    - same as `Energy`_
+    - same as `Energy`_
+
+  * - samples
+    - same as `Energy`_
+    - same as `Energy`_
+
+  * - components
+    - same as `Energy`_
+    - same as `Energy`_
```
Luthaf, line 198: I think it is worth repeating the information here, we used same as `Energy` for the energy ensemble which is related, but this output is not related to the energy and might end up in a different page.

docs/src/atomistic/outputs.rst
```diff
@@ -167,3 +167,43 @@ Energy ensemble gradients
...
 output (see `Energy gradients`_).
+
+.. _features:
+
+Features
+^^^^^^^^
+
+Last layer features represent the outputs generated by the model before the final predictions,
+such as energy or forces, are made. They are associated with the ``"features"`` key in the model
+outputs, and must adhere to the following metadata:
```
Luthaf, line 178: These are not necessarily "last layer" features, since there might not even be multiple layer, or another prediction made by the model.

docs/src/atomistic/outputs.rst
```diff
@@ -167,3 +167,43 @@ Energy ensemble gradients
...
+    - same as `Energy`_
+    - same as `Energy`_
+
+  * - properties
+    -
+    - the features can have arbitrary properties.
+
+  * - gradients
+    -
+    - the features should not have any explicit gradients, all gradient calculations are done using autograd.
```
Luthaf, line 206: I think we'll want to allow explicit gradients as well, just in case. I'll write this part.

Luthaf (commented):
This looks good overall. The only point is about the general name: I'm not sure we want to use "features" for this. A couple of potential alternatives:

- "fingerprint"
- "representation"
- "descriptor"
- "ml-<any of the above>"
- ...

docs/src/atomistic/outputs.rst
sofiia-chorna: So, you think we need to split the pages with the standard outputs related and not related to the energy?

sofiia-chorna:
> This looks good overall. The only point is about the general name: I'm not sure we want to use "features" for this. A couple of potential alternatives:
> 
> * "fingerprint"
> * "representation"
> * "descriptor"
> * "ml-"
> * ...

@Luthaf @frostedoyster 
Thanks for reviewing!
I’m not sure which name among proposed is better... Unless you have some preferences, we could ask in the metatensor chat, if you wish.

docs/src/atomistic/outputs.rst
ceriottm: technically these are internal/hidden/learned features. all in all, `features` might be the least bad possible name if we want to have a generic way to refer to the intermediate values in the model. maybe 'learned_features' or 'model_features' but I'm not sure there's much gain from all those extra letters.

Luthaf: this output is not only intended for internal/learned features, but any kind of representation (so PLUMED would move to use this output instead of the current custom output used over there for CV)

ceriottm: IMO then `features` is the most generic catch-all term - but I support polling MLDEV.

docs/src/atomistic/outputs.rst
Luthaf: We don't need to do it now, but we will likely do something like this at some point. For now just having the documentation for this output be self contained should be enough!

sofiia-chorna: Ok ok, I added the doc

frostedoyster (approved):

Luthaf:
I'll update the main description quickly and then this should be good to merge!

Issue #646, opened 2024-06-07
Default argument for `cell` in `metatensor.torch.atomistic.System`?


frostedoyster:
When creating a system, it's annoying to have to set a cell even though the system is non-periodic

Luthaf:
I'm not sure about this. I like that users are forced to think about what the `cell` should be, this lowers the risk that they forget about it. Like I've been bitten by ASE not requiring a value for `pbc` (even if `cell` is set) multiple times already.

Luthaf:
This is unlikely to happen

PR #735, opened 2024-09-23
Add PBCs to the `System` class

frostedoyster:
This PR implements PBCs in the `System` class, allowing for structures where only some of the dimensions have PBCs.

A few questions:
- should we have default arguments (most likely `True, True, True` for backward compatibility)?
- is the strategy to ignore the cell if the PBCs are False? Or do we require it to have some all-zero rows if PBCs are False?

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2004641647.zip)

### DIFF
--- a/metatensor-torch/CHANGELOG.md
+++ b/metatensor-torch/CHANGELOG.md
@@ -6,18 +6,19 @@ a changelog](https://keepachangelog.com/en/1.1.0/) format. This project follows
 
 ## [Unreleased](https://github.com/metatensor/metatensor/)
 
-<!-- Possible sections:
-
 ### Added
 
 ### Fixed
 
 ### Changed
 
+- the `System` class now supports boundary conditions along some axes but not others. This is implemented
+  via a new `pbc` attribute. Any non-periodic dimension in a `System` must have the corrresponding cell
+  vector set to zero.
+
 ### Deprecated
 
 ### Removed
--->
 
 ## [Version 0.5.5](https://github.com/metatensor/metatensor/releases/tag/metatensor-torch-v0.5.5) - 2024-09-03
 
--- a/metatensor-torch/CHANGELOG.md
+++ b/metatensor-torch/CHANGELOG.md
@@ -6,19 +6,22 @@ a changelog](https://keepachangelog.com/en/1.1.0/) format. This project follows
 
 ## [Unreleased](https://github.com/metatensor/metatensor/)
 
-### Added
+<!-- Possible sections for each package:
 
-### Fixed
+#### Added
+
+#### Fixed
+
+#### Changed
+
+#### Removed
+-->
 
 ### Changed
 
 - the `System` class now supports boundary conditions along some axes but not others. This is implemented
   via a new `pbc` attribute. Any non-periodic dimension in a `System` must have the corrresponding cell
   vector set to zero.
-
-### Deprecated
-
-### Removed
 
 ## [Version 0.5.5](https://github.com/metatensor/metatensor/releases/tag/metatensor-torch-v0.5.5) - 2024-09-03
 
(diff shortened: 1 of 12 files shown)


Luthaf (commented):
> should we have default arguments (most likely True, True, True for backward compatibility)?

With the same reasoning as in #646, I would not use a default value but rather force the user to think explicitly about this.

> is the strategy to ignore the cell if the PBCs are False? Or do we require it to have some all-zero rows if PBCs are False?

I'm not sure. I'd say it would be good to force the cell row to be 0 is the pbc is disabled along this dimension. If nothing else, this should help catch broken NL implementations downstream.

metatensor-torch/src/atomistic/system.cpp
```diff
@@ -513,6 +546,38 @@ void SystemHolder::set_cell(torch::Tensor cell) {
     this->cell_ = std::move(cell);
 }
 
+void SystemHolder::set_pbc(torch::Tensor pbc) {
+    if (pbc.device() != this->device()) {
```
Luthaf, line 550: should we move the constructor checks in this function & then call this function from constructor to perform the same checks? Right now this is missing the check for cell vectors vs pbc

metatensor-torch/src/atomistic/system.cpp
```diff
@@ -410,6 +411,38 @@ SystemHolder::SystemHolder(torch::Tensor types, torch::Tensor positions, torch::
...
+    if (pbc_.scalar_type() != torch::kBool) {
+        C10_THROW_ERROR(ValueError,
+            "`pbc` must be a tensor of booleans, got " +
+            scalar_type_name(pbc_.scalar_type()) + " instead"
+        );
+    }
+
+    // if the pbcs are False along any directions, we check that the
+    // corresponding cell vectors are zero
+    if (pbc_.device().str() != "meta") {  // skip this check for meta tensors
```
Luthaf, line 438:
```suggestion
    if (pbc_.device().is_meta()) {  // skip this check for meta tensors
```

Luthaf (commented):
Looks good, could you update the CHANGELOG as well?

Also, this will need changes in the LAMMPS and PLUMED interfaces once we release the corresponding code, will you have time to do these changes? 

I think this is different enough from the previous code assumptions that it is worth doing a 0.6 (breaking release).

metatensor-torch/src/atomistic/system.cpp
frostedoyster: Moving all the checks into the setter meant dropping a lot of detail in the error messages. I've re-written the check in the setter so the error messages can be better

frostedoyster:
I'd be happy do adapt the LAMMPS and PLUMED interfaces

metatensor-torch/CHANGELOG.md
```diff
@@ -6,18 +6,19 @@ a changelog](https://keepachangelog.com/en/1.1.0/) format. This project follows
 
 ## [Unreleased](https://github.com/metatensor/metatensor/)
 
-<!-- Possible sections:
```
Luthaf, line 9: please leave this in place & copy the section you need below

metatensor-torch/src/atomistic/system.cpp
Luthaf: still relevant?

metatensor-torch/src/atomistic/system.cpp
```diff
@@ -513,6 +546,49 @@ void SystemHolder::set_cell(torch::Tensor cell) {
...
+    if (pbc.size(0) != 3) {
+        C10_THROW_ERROR(ValueError,
+            "new `pbc` must contain 3 entries, got a tensor with " +
+            std::to_string(pbc.size(0)) + " values"
+        );
+    }
+
+    // if the pbcs are False along any directions, we check that the
+    // corresponding cell vectors are zero
+    if (pbc.device().str() != "meta") {  // skip this check for meta tensors
```
Luthaf, line 580:
```suggestion
    if (pbc.device().is_meta()) {  // skip this check for meta tensors
```

frostedoyster: Doesn't compile :(

Luthaf: ok, I'm trying a comparison with `pbc_.device().type()` instead. I'll revert if it does not work

Luthaf (approved):

PR #738, opened 2024-09-24
Allow to request `additional_outputs` in ASE calculator

Luthaf:
This enable the simultaneous calculation of multiple properties during an MD run. The main caveat is that the `additional_outputs` will be computed at every step, even if they would not be needed. One of the main use case is thus to compute things one would need at every step anyway, such as uncertainty estimate on the energy.

# Reviewer checklist

### DIFF
--- a/python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
+++ b/python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
@@ -66,10 +66,20 @@ class MetatensorCalculator(ase.calculators.calculator.Calculator):
     installed, in which case it will be used instead.
     """
 
+    additional_outputs: Dict[str, TensorMap]
+    """
+    Additional outputs computed by :py:meth:`calculate` are stored in this dictionary.
+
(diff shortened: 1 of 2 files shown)


python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -66,10 +66,20 @@ class MetatensorCalculator(ase.calculators.calculator.Calculator):
     installed, in which case it will be used instead.
     """
 
+    additional_outputs: Dict[str, TensorMap]
```
frostedoyster, line 69:
```suggestion
    additional_outputs: Dict[str, TensorMap] = {}
```
Should this be initialized? Perhaps to `{name: None for name in ...}`

frostedoyster (approved):

python/metatensor-torch/metatensor/torch/atomistic/ase_calculator.py
frostedoyster: I don't see it being initialized in the `__init__`, maybe I'm being dumb

PR #747, opened 2024-10-11
Switch to ruff for linting

ceriottm:
Got sufficiently pestered by flake8 to do this. I think none of the checklist items apply.

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2082859285.zip)

⚙️ [Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/metatensor/metatensor/actions/artifacts/2083079029.zip)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,17 +45,6 @@ packages = []
 
 ### ======================================================================== ###
 
-[tool.isort]
-skip = "__init__.py"
-profile = "black"
-line_length = 88
-indent = 4
-include_trailing_comma = true
-lines_after_imports = 2
-known_first_party = "metatensor"
-
-### ======================================================================== ###
-
 [tool.coverage.paths]
 # path remapping for coverage. `coverage` will map path matching the second
 # string to the first string.
@@ -78,3 +67,13 @@ filterwarnings = [
     "ignore:Attribute s is deprecated and will be removed in Python 3.14:DeprecationWarning",
     "ignore:ast.NameConstant is deprecated and will be removed in Python 3.14:DeprecationWarning",
 ]
+
+### ======================================================================== ###
+
+[tool.ruff.lint]
+select = ["E", "F", "B", "I"]
+ignore = ["B018", "B904"]
+
+[tool.ruff.lint.isort]
+lines-after-imports = 2
+known-first-party = ["metatensor"]
--- a/python/metatensor-operations/metatensor/operations/divide.py
+++ b/python/metatensor-operations/metatensor/operations/divide.py
@@ -85,9 +85,7 @@ def _divide_block_block(block_1: TensorBlock, block_2: TensorBlock) -> TensorBlo
             [-1] + [1] * diff_components + _shape
         ) ** 2 + gradient_1.values / block_2.values[
             _dispatch.to_index_array(gradient_samples_to_values_samples_2)
-        ].reshape(
-            [-1] + [1] * diff_components + _shape
-        )
+        ].reshape([-1] + [1] * diff_components + _shape)
 
         result_block.add_gradient(
             parameter=parameter_1,
--- a/tox.ini
+++ b/tox.ini
@@ -210,34 +210,28 @@ commands =
 
 
 [testenv:lint]
-description =
-    lint the Python code with flake8 (code linter), black (code formatter), and
-    isort (sorting of imports)
+description = Run linters and type checks
 package = skip
 deps =
-    flake8
-    flake8-bugbear
-    black
+    ruff
     blackdoc
-    isort
 
 commands =
-    flake8 --extend-exclude "*/build/*" {[testenv]lint_folders}
-    black --check --diff {[testenv]lint_folders}
+    ruff format --diff {[testenv]lint_folders}
+    ruff check {[testenv]lint_folders}
     blackdoc --check --diff {[testenv]lint_folders}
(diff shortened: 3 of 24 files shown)


python/metatensor-operations/metatensor/operations/divide.py
```diff
@@ -85,9 +85,7 @@ def _divide_block_block(block_1: TensorBlock, block_2: TensorBlock) -> TensorBlo
             [-1] + [1] * diff_components + _shape
         ) ** 2 + gradient_1.values / block_2.values[
             _dispatch.to_index_array(gradient_samples_to_values_samples_2)
-        ].reshape(
-            [-1] + [1] * diff_components + _shape
-        )
+        ].reshape([-1] + [1] * diff_components + _shape)
```
Luthaf, line 88: so this is ruff vs black giving different formatting?

tox.ini
```diff
@@ -210,34 +210,28 @@ commands =
...
-    black
+    ruff
     blackdoc
-    isort
 commands =
-    black {[testenv]lint_folders}
     blackdoc {[testenv]lint_folders}
-    isort {[testenv]lint_folders}
+    ruff format {[testenv]lint_folders}
+    ruff check --fix-only {[testenv]lint_folders}
```
Luthaf, line 234: this auto-fix some of the linting issues?

tox.ini
```diff
@@ -210,34 +210,28 @@ commands =
...
     blackdoc --check --diff {[testenv]lint_folders}
-    isort --check-only --diff --extend-skip-glob "*/build/*" {[testenv]lint_folders}
+
 
 [testenv:format]
 description = Abuse tox to do actual formatting on all files.
 package = skip
 deps =
-    black
+    ruff
```
Luthaf, line 229: should we pin ruff to a fixed version? How often do they update/change rules?

Luthaf (commented):
Could you also remove the remaining flake8 configuration in tox.ini, as well as isort configuration in pyproject.toml?

Also, should we translate the above configurations to ruff, or does it already applies the current settings?

Luthaf:
From reading https://docs.astral.sh/ruff/linter/#rule-selection, it seems that we need to explicitly enable the flake8 rules again, and that the current changes just disable them.

Not sure what other lab projects using ruff are doing, but we might want to enable the lint rules there as well!

python/metatensor-operations/metatensor/operations/divide.py
ceriottm: yes, and I'd say it's usually better.

tox.ini
ceriottm: yes it should fix some of the linting issues - basically if you tox format, it should pass tox lint

tox.ini
ceriottm: IDK, if we start getting too much linter noise we can revisit this

ceriottm:
> From reading https://docs.astral.sh/ruff/linter/#rule-selection, it seems that we need to explicitly enable the flake8 rules again, and that the current changes just disable them.
> 
> Not sure what other lab projects using ruff are doing, but we might want to enable the lint rules there as well!

We just run and let it apply default rules. They make a lot of sense it seems. IDK if we want to add other requirements.

Luthaf:
> By default, Ruff enables Flake8's F rules, along with a subset of the E rules, omitting any stylistic rules that overlap with the use of a formatter, like ruff format or [Black](https://github.com/psf/black).

To re-create the same config as was previously used here, we would need to enable `F, E, B, I` rules. I'd say that most projects should enable `I` as well (`isort` rules), but `B` (`flake8-bugbear`) can be enabled on a project-by-project basis.

ceriottm:
> > By default, Ruff enables Flake8's F rules, along with a subset of the E rules, omitting any stylistic rules that overlap with the use of a formatter, like ruff format or [Black](https://github.com/psf/black).
> 
> To re-create the same config as was previously used here, we would need to enable `F, E, B, I` rules. I'd say that most projects should enable `I` as well (`isort` rules), but `B` (`flake8-bugbear`) can be enabled on a project-by-project basis.

I added those and rerun. There were a couple of bugbear errors that seemed really overkill, and there is quite a bit of linter noise - I think due to the slightly different import sorting. I'll see if it's possible to get closer to the isort config we used

tox.ini
Luthaf: Hmm, should we rename this to `tox -e fix-lint`? Since it does more than formatting? Or is the old name fine?

pyproject.toml
```diff
@@ -78,3 +67,13 @@ filterwarnings = [
     "ignore:Attribute s is deprecated and will be removed in Python 3.14:DeprecationWarning",
     "ignore:ast.NameConstant is deprecated and will be removed in Python 3.14:DeprecationWarning",
 ]
+
+### ======================================================================== ###
+
+[tool.ruff.lint]
+select = ["E", "F", "B", "I"]
+ignore = ["B018", "B904"]
```
Luthaf, line 75: Both of these seems relatively useful, but I can see B904 being quite noisy on the current codebase. I'll re-enable them in a separate PR & clean up the code

Luthaf: B018: https://docs.astral.sh/ruff/rules/useless-expression/

B904: https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/

Luthaf:
I'll update this branch w.r.t. master & then merge

PR #752, opened 2024-10-17
Use larger SmallVec to store Labels entries in HashMap

Luthaf:
The previous default of 4 was a problem when storing neighbors samples in atomistic model, which have 5 dimensions. 

In the torch-pme LAMMPS benchmark, this brings the cost of creating neighbor samples Labels for 167944 pairs from 14ms down to 7.5ms.

Out of the remaining 7.5ms, 2.5ms is copying the entries in the internal array and 5ms is constructing the HashMap. I'll think about ways to mitigate this (e.g. the HashMap could be lazily initialized easily, this would be harder for the internal array, …).

# Reviewer checklist

### DIFF
--- a/metatensor-core/src/labels.rs
+++ b/metatensor-core/src/labels.rs
@@ -96,14 +96,21 @@ impl LabelValue {
     }
 }
 
-type DefaultHasher = std::hash::BuildHasherDefault<ahash::AHasher>;
+// Labels uses `AHash` instead of the default hasher in std since `AHash` is
+// much faster and we don't need the cryptographic strength hash from std.
+type AHashHasher = std::hash::BuildHasherDefault<ahash::AHasher>;
+
+// Use a small vec to store Labels entries in the `positions`. This helps by
+// removing heap allocations in the most common case (fewer than 8 dimensions in
+// the Labels), while still allowing the Labels to contains many dimensions.
+type LabelsEntry = SmallVec<[LabelValue; 8]>;
 
 /// Builder for `Labels`, this should be used to construct `Labels`.
 pub struct LabelsBuilder {
     // cf `Labels` for the documentation of the fields
     names: Vec<ConstCString>,
     values: Vec<LabelValue>,
-    positions: HashMap<SmallVec<[LabelValue; 4]>, usize, DefaultHasher>,
+    positions: HashMap<LabelsEntry, usize, AHashHasher>,
 }
 
 impl LabelsBuilder {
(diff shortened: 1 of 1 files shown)


metatensor-core/src/labels.rs
```diff
@@ -96,14 +96,21 @@ impl LabelValue {
...
 
-type DefaultHasher = std::hash::BuildHasherDefault<ahash::AHasher>;
+// Labels uses `AHash` instead of the default hasher in std since `AHash` is
+// much faster and we don't need the cryptographic strength hash from std.
+type AHashHasher = std::hash::BuildHasherDefault<ahash::AHasher>;
+
+// Use a small vec to store Labels entries in the `positions`. This helps by
+// removing heap allocations in the most common case (fewer than 8 dimensions in
+// the Labels), while still allowing the Labels to contains many dimensions.
+type LabelsEntry = SmallVec<[LabelValue; 8]>;
```
ceriottm, line 106: Let me understand - does this mean the minimum size of a labelsentry  is 8 ints? Does this have memory implications?

ceriottm:
I confirm this speeds up quite a bit the creation of the labels

ceriottm (approved):
Looks good, works and is faster indeed. Only concern I have (which is probably my poor rust knowledge) is whether this implies a floor of 8 ints per label memory footprint. Approved conditional on this not being the case ^_^

metatensor-core/src/labels.rs
Luthaf: Yes, this will make the LabelsEntry at least 33 bytes (8 ints + 1 discriminant to know if the data is on stack or the heap). Keep in mind that LabeslEntry is only used inside the `positions` HashMap, and not the `values` array.

In the worst case (Labels with 1 dimension), this means we'll use `size * 4` (in the values) + `size * (33 + 2)` (in the `positions` HashMap, 33 bytes per labels entry and 2 bytes for the position index) bytes of storage.

So for 100000 entries, this will take 3.7 MiB of RAM, which feels fine to me. When running on GPU, only the `values` are duplicated in GPU memory, so that would be an extra 390kiB of VRAM.

Finally, I'm looking into creating the HashMap only when required (i.e. when the user calls a function that needs it), which would further reduce the memory footprint.

ceriottm: Understood. Maybe one could make the size a #DEFINE so that one can set it, but seems overkill.

PR #753, opened 2024-10-18
Implement EquivariantTransform in metatensor-learn

### LINKED ISSUE #744: Equivariant activation functions in metatensor-learn
As discussed with @jwa7, it would be nice to add equivariant activation functions to metatensor-learn, where nonlinear activations are applied to invariants built from contractions of each covariant block.

ppegolo:
Fixes #744. Add EquivariantTransform to metatensor-learn, where a callable is applied to invariants built from contractions of each covariant block and elementwise multiplied back to the covariant blocks.

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2094604609.zip)


python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,131 @@
...
+        return self.module_map(tensor)
+
+
+class _CovariantTransform(Module):
+    """
+    Custom :py:class:`Module` to apply a generic :py:class:`callable` to invariants
+    built from a covariant block.
+
+    :param in_features: :py:class:`int`
+        Input feature dimension.
```
jwa7, line 102:
```suggestion
    :param in_features: a :py:class:`int`, the input feature dimension. This also corresponds to the output
        feature size as the shape of the tensor passed to :py:meth:`forward` is preserved.
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,131 @@
...
+    :param module: :py:class:`callable`
+        A callable that instantiates a native :py:class:`torch.nn.module`
+        with parameters `in_features`
```
jwa7, line 105:
```suggestion
    :param module: a :py:class:`callable` that when called with parameters ``in_features`` and optionally ``dtype``
        and ``device`` constructs a native :py:class:`torch.nn.Module`.
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,131 @@
...
+        self.in_features = in_features
+        self.module = module(in_features, device=device, dtype=dtype)
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+
+        invariant_components = input.norm(dim=1, keepdim=True)
+        nonlinear_at_invariants = self.module(invariant_components)
+        tensor_out = nonlinear_at_invariants * input
+
+        return tensor_out
```
jwa7, line 131:
```suggestion
        assert len(input.shape) == 3, "``input`` must be a three-dimensional tensor"
        invariant = input.norm(dim=1, keepdim=True)
        invariant_transformed = self.module(invariant_components)
        tensor_out = invariant_transformed * input

        return tensor_out
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,131 @@
...
+        dtype: Optional[torch.dtype] = None,
+    ) -> None:
+
+        super().__init__()
+
+        self.in_features = in_features
+        self.module = module(in_features, device=device, dtype=dtype)
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+
```
jwa7, line 126:
```suggestion
    """
    Creates an invariant block from the ``input`` covariant, and transforms it by applying the torch 
    ``module`` passed to the class constructor. Then uses the transformed invariant as an element
    -wise multiplier for the ``input`` block. 
    
    Transformations are applied consistently to components (axis 1) to preserve equivariance.
    """
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,131 @@
...
+        :param tensor: :py:class:`TensorMap` with the input tensor to be transformed.
+        :return: :py:class:`TensorMap`
+        """
+        return self.module_map(tensor)
+
+
+class _CovariantTransform(Module):
+    """
+    Custom :py:class:`Module` to apply a generic :py:class:`callable` to invariants
+    built from a covariant block.
```
jwa7, line 99:
```suggestion
    Applies an arbitrary shape-preserving transformation defined in ``module`` to a 3-dimensional
    tensor in a way that preserves equivariance.
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,131 @@
...
+            modules.append(module_i)
+
+        self.module_map = ModuleMap(in_keys, modules, out_properties)
+
+    def forward(self, tensor: TensorMap) -> TensorMap:
+        """
+        Apply the transformation to the input tensor map `tensor`.
+
+        :param tensor: :py:class:`TensorMap` with the input tensor to be transformed.
+        :return: :py:class:`TensorMap`
```
jwa7, line 91:
```suggestion
        :return: :py:class:`TensorMap` corresponding to the transformed input ``tensor``.
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,131 @@
...
+    :param dtype: :py:class:`torch.dtype` passed to the module instance.
+    """
+
+    def __init__(
+        self,
+        module: callable,
+        in_keys: Labels,
+        invariant_key_idxs: List[int],
+        in_features: Union[int, List[int]],
+        out_features: Optional[Union[int, List[int]]] = None,
```
jwa7, line 47: `out_features` are not needed here as they are the same as `in_features`. The  user can still pass `out_properties` if they want the output TensorMap to have the correct properties metadata, i.e. passing:
```py
EquivariantTransform(
    ...,
    out_properties=[tensor[key].block.properties for key in in_keys],
    ...,
)
```

```suggestion
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,131 @@
...
+    ) -> None:
+        super().__init__()
+        # Infer `out_features` if not provided
+        if out_features is None:
+            if out_properties is None:
+                raise ValueError(
+                    "If `out_features` is not provided,"
+                    " `out_properties` must be provided."
+                )
+            out_features = [len(p) for p in out_properties]
```
jwa7, line 61:
```suggestion
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,131 @@
...
+
+        # Check input parameters, convert to lists (for each key) if necessary
+        in_features = _check_module_map_parameter(
+            in_features, "in_features", int, len(in_keys), "in_keys"
+        )
+        out_features = _check_module_map_parameter(
+            out_features, "out_features", int, len(in_keys), "in_keys"
+        )
```
jwa7, line 69:
```suggestion
```

python/metatensor-learn/tests/equivariant_transform.py
```diff
@@ -0,0 +1,79 @@
...
+        torch.nn.Linear(
+            in_features=5,
+            out_features=in_features,
+            device=device,
+            dtype=dtype,
+        ),
+    )
+
+
+def test_equivariance(tensor, wigner_d_real):
```
jwa7, line 55: Looks good! Maybe we could also have a test with a module that uses the identity operation (`torch.nn.Identity`) and check input equals output?

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,131 @@
...
+from .module_map import ModuleMap
+from ._utils import _check_module_map_parameter
+
+
+class EquivariantTransform(Module):
+    """
+    Applies a module to invariant blocks and to the invariants commputed from each
+    covariant block of a :py:class:`TensorMap` passed to its forward method, indexed by
+    :param in_keys:. For covariant blocks, the transformed values are then elementwise
+    multiplied to the original block, without affecting its `component` dimension.
```
jwa7, line 16:
```suggestion
    A custom :py:class:`torch.nn.Module` that applies an arbitrary shape- and equivariance-preserving
    transformation to an input :py:class:`TensorMap`.
    
    ``module`` is passed as a callable with parameters ``in_features`` and optionally ``dtype`` and 
    ``device``. This callable constructs an arbitrary shape-preserving :py:class:`torch.nn.Module` 
    transformation (i.e. :py:class:`torch.nn.Tanh`). Separate instantiations are created for each block
    using the metadata information passed in ``in_keys`` and ``in_features``. 
    
    For invariant blocks in `in_keys` and indexed by `invariant_key_idxs`, the transformation is applied
    as is. For covariant blocks, an invariant multiplier that preserves covariance is created from the 
    transformation.
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,131 @@
...
+        bias applied.
+    :param in_features: list of int, the number of features in the input tensor for each
+        block indexed by the keys in :param in_keys:. If passed as a single value, the
+        same number of features is assumed for all blocks.
+    :param out_properties: list of :py:class`Labels` (optional), the properties labels
+        of the output. By default the output properties are relabeled using
+        Labels.range. If provided, :param out_features: can be inferred and need not be
+        provided.
+    :param device: :py:class:`torch.device` Device to instantiate the module in.
+    :param dtype: :py:class:`torch.dtype` passed to the module instance.
```
jwa7, line 38:
```suggestion
    :param device: :py:class:`str` or :py:class:`torch.device`, the computational device
        to use for calculations.
    :param dtype: :py:class:`torch.dtype` , the scalar type to use to store parameters.
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,131 @@
...
+        `in_keys` in the input :py:class:`TensorMap`. Only blocks for these keys will
+        have bias applied according to the user choice. Covariant blocks will not have
+        bias applied.
+    :param in_features: list of int, the number of features in the input tensor for each
+        block indexed by the keys in :param in_keys:. If passed as a single value, the
+        same number of features is assumed for all blocks.
+    :param out_properties: list of :py:class`Labels` (optional), the properties labels
+        of the output. By default the output properties are relabeled using
+        Labels.range. If provided, :param out_features: can be inferred and need not be
+        provided.
```
jwa7, line 36:
```suggestion
    :param out_properties: list of :py:class`Labels` (optional), the properties labels
        of the output. By default the output properties are relabeled using
        Labels.range.
```

jwa7 (changes requested):
Looks good so far! :)

A few comments, see below. Still to come are torch tests and documentation, but we can deal with this once we've finalised the design.

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,145 @@
...
+        of the output. By default the output properties are relabeled using
+        Labels.range.
+    :param device: :py:class:`str` or :py:class:`torch.device`, the computational device
+        to use for calculations.
+    :param dtype: :py:class:`torch.dtype` , the scalar type to use to store parameters.
+    """
+
+    def __init__(
+        self,
+        module: callable,
```
Luthaf, line 51: Should this be

```suggestion
        module: torch.nn.Module,
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,145 @@
...
+        :param tensor: :py:class:`TensorMap` with the input tensor to be transformed.
+        :return: :py:class:`TensorMap`
+        """
+        return self.module_map(tensor)
+
+
+class _CovariantTransform(Module):
+    """
+    Applies an arbitrary shape-preserving transformation defined in ``module`` to a
+    3-dimensional tensor in a way that preserves equivariance.
```
Luthaf, line 104: this needs to say it is using the norm to create invariants

Luthaf (commented):
just small comments, I did not look deep into this

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
jwa7: So the idea was to not have it yet initialised, so that different instances for each block can be created on runtime according to feature size.

For instance, `module` could just be i.e. a `Tanh` activation, but also any shape-preserving transformation like:
```py
def module(in_features: int, dtype, device):
    return torch.nn.Sequential(
        torch.nn.Linear(in_features, 10),
        torch.nn.Tanh(),
        torch.nn.Linear(10, in_features),
    )
```

with the only requirement being that it is parametrized by `in_features`. Do you see a better way of doing it?

Luthaf: This is deviating a bit from usual torch API, so I would rather not do something like this. How about taking either a single module (then used by all blocks, with weight sharing) or a list of modules (one per block)?

Also, as a side comment, why does this need to preserve the number of features?

jwa7: So the shape needs to be preserved so that it can be used as an element wise multiplier for covariants, and for that reason we can't weight share as the dimensions for all blocks are in general different.

And yes I suppose we could pass a list

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,140 @@
...
+    as the parameter for all blocks. Alternatively, they can be passed as a list to
+    control the parameters applied to each block indexed by the keys in :param in_keys:.
+
+    :param module: a :py:class:`list` of :py:class:`torch.nn.Module` containing the
+        transformations to be applied to from each block indexed by
+        :param:`in_keys`. Transformations for invariant and covariant blocks differ. See
+        above.
+    :param in_keys: :py:class:`Labels`, the keys that are assumed to be in the input
+        tensor map in the :py:meth:`forward` method.
+    :param in_features: list of int, the number of features in the input tensor for each
```
jwa7, line 37:
```suggestion
    :param in_features: :py:class:`list` of :py:class:`int`, the number of features in the input tensor for each
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,140 @@
...
+        block indexed by the keys in :param in_keys:. If passed as a single value, the
+        same number of features is assumed for all blocks.
+    :param out_properties: list of :py:class`Labels` (optional), the properties labels
```
jwa7, line 40:
```suggestion
    :param out_properties: :py:class:`list` of :py:class`Labels` (optional), the properties labels
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
jwa7: @ppegolo just a reminder on this one :)

jwa7 (changes requested):
Looks great! Almost there - a few small comments below to address :)

jwa7:
Also @ppegolo I think a test needs updating. There's the following error in the CI:
```
self = <[AttributeError("'Linear' object has no attribute 'bias'") raised in repr()] Linear object at 0x7f27c73b8bf0>
in_features = 2, out_features = 5, bias = True, device = device(type='cpu')
dtype = 7
```
as in the test `test_equivariant_transform` in torchscript.py, the dtype is being picked up as an `int`. Might be best to pass `dtype=tensor.block(0).values.dtype` instead?

jwa7 (approved):
LGTM!

python/metatensor-learn/CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased](https://github.com/metatensor/metatensor/)
 
+### Added
+
+- Added `metatensor.learn.nn` module `EquivariantTransform` to apply any
```
Luthaf, line 11:
```suggestion
- Added `metatensor.learn.nn.EquivariantTransform` to apply any 
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,142 @@
...
+import torch
+from torch.nn import Module
+
+from .._backend import Labels, TensorMap
+from .._dispatch import int_array_like
+from ._utils import _check_module_map_parameter
+from .module_map import ModuleMap
+
+
+class EquivariantTransform(Module):
```
Luthaf, line 12:
```suggestion


from .._backend import Labels, TensorMap
from .._dispatch import int_array_like
from ._utils import _check_module_map_parameter
from .module_map import ModuleMap


class EquivariantTransform(torch.nn.Module):
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    A custom :py:class:`torch.nn.Module` that applies an arbitrary shape- and
+    equivariance-preserving transformation to an input :py:class:`TensorMap`.
+
+    ``module`` is passed as a callable with parameters ``in_features`` and optionally
+    ``dtype`` and ``device``. This callable constructs an arbitrary shape-preserving
+    :py:class:`torch.nn.Module` transformation (i.e. :py:class:`torch.nn.Tanh`).
+    Separate instances are created for each block using the metadata information
+    passed in ``in_keys`` and ``in_features``.
```
Luthaf, line 21: I think this is outdated

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,142 @@
...
+
+    For invariant blocks in `in_keys` and indexed by `invariant_keys`, the
+    transformation is applied as is. For covariant blocks, an invariant multiplier is
+    created applying the transformation to the norm of the block over the component
+    dimension.
```
Luthaf, line 26:
```suggestion
    For invariant blocks (specified with ``invariant_keys``), the transformation is 
    applied as is. For covariant blocks, an invariant multiplier is created, applying 
    the transformation to the norm of the block over the component dimensions.
```

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,142 @@
...
+        passed as a single value, the same number of features is assumed for all blocks.
+    :param out_properties: :py:class:`list` of :py:class`Labels` (optional), the
+        properties labels
+        of the output. By default the output properties are relabeled using
+        Labels.range.
+    :param invariant_keys: a :py:class:`Labels` object that is used to select the
+        invariant keys from ``in_keys``. If not provided, the invariant keys are assumed
+        to be those where key dimensions ``["o3_lambda", "o3_sigma"]`` are equal to
+        ``[0, 1]``.
+    """
```
Luthaf, line 45: It would be nice to have some usage example for this class, since I doubt someone will be able to use it just by reading the docstring.

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
```diff
@@ -0,0 +1,142 @@
...
+    :py:class:`torch.nn.Module` transformation (i.e. :py:class:`torch.nn.Tanh`).
+    Separate instances are created for each block using the metadata information
+    passed in ``in_keys`` and ``in_features``.
+
+    For invariant blocks in `in_keys` and indexed by `invariant_keys`, the
+    transformation is applied as is. For covariant blocks, an invariant multiplier is
+    created applying the transformation to the norm of the block over the component
+    dimension.
+
+    :param module: a :py:class:`list` of :py:class:`torch.nn.Module` containing the
```
Luthaf, line 28:
```suggestion
    :param modules: a :py:class:`list` of :py:class:`torch.nn.Module` containing the
```

Luthaf (commented):
I have a couple of improvements to suggest to the documentation. One other thing is thr name of this class. I'm not 100% sold on `EquivariantTransform`, since it is a bit close to _transformer_. `EquivariantNonLinear` would maybe be a better name, but is quite a mouthful. `EquivariantTransformation`, `EquivariantApply` are other alternative, but I don't really love them either. Thoughts?

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
jwa7: In the docstring or elsewhere? i.e. could we make a tutorial of using `learn.nn` modules together in defining custom architectures separately?

jwa7:
Thanks for the comments @Luthaf . I think `EquivariantTransformation` makes most sense here. I can't really think of a better one for now

ppegolo:
Thanks @Luthaf and @jwa7. Either `EquivariantTransformation` or `EquivariantNonLinear` are fine by me. The latter is the way I called it in the version it's implemented in `mlelec`, but I think I prefer `EquivariantTransformation` now

Luthaf:
From #744 title, `EquivariantActivation` would also be a good name for this IMO

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
ppegolo: Removed, thanks for spotting it

ppegolo:
At the beginning we thought about that name, but we later realized the scope of the module is slightly broader, as it's not limited to applying activation functions, but any module. So maybe `EquivariantTransformation` is a better name

Luthaf:
ok, let's go with `EquivariantTransformation` for now!

ppegolo:
Updated the name to `EquivariantTransformation`

python/metatensor-learn/metatensor/learn/nn/equivariant_transform.py
jwa7: Ok we've added a docstring example, tutorial example coming later :)

python/metatensor-learn/metatensor/learn/nn/equivariant_transformation.py
```diff
@@ -35,6 +35,92 @@ class EquivariantTransformation(torch.nn.Module):
...
+        ...     return torch.nn.Tanh()
+        ...
+
+        Instantiate the modules containing the transformation
+
+        >>> in_features = [len(tensor.block(key).properties) for key in tensor.keys]
+        >>> modules = [
+        ...     module_wrapper(in_feat, device=tensor.device, dtype=tensor.dtype)
+        ...     for in_feat in in_features
+        ... ]
```
Luthaf, line 96: this seems not relevant for this function? 
```suggestion
        Define the transformation to apply to the TensorMap
        >>> modules = [torch.nn.Tanh(), torch.nn.Tanh()]
```

Luthaf (approved):
looks good overall, thanks both!

python/metatensor-learn/metatensor/learn/nn/equivariant_transformation.py
jwa7: Yes this could be useful, especially when building more complicated sequential objects and including and sandwiching with i.e. rascaline / CG layers. Something to come back to for sure :)

ppegolo: You're right, updated

PR #761, opened 2024-10-24
add unit check and warning when saving

### LINKED ISSUE #630: Warn when creating/exporting a `MetatensorAtomisticModel` without units
When creating an MetatensorAtomisticModel instance it is allowed to keep the units as empty strings. This is allowed and  values will be passed to MD engines without conversion. However, this could be problematic when switching engines. In metatensor-models we [warn](https://github.com/lab-cosmo/metatensor-models/blob/2bc152fab8ae7eb38893ab70d56710d266180c95/src/metatensor/models/utils/export.py#L25-L38) users about these possible issues and it might be useful to port these warnings directly into `metatensor`.

See also https://github.com/lab-cosmo/metatensor-models/pull/185#discussion_r1613360706

SanggyuChong:
This fixes #630.

Added simple checking routines within the save function to look for the units, and to raise a warning when the string is empty.
EDIT: took inspiration from the export function of metatrain as suggested by @Luthaf.

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2098928048.zip)

### DIFF
--- a/python/metatensor-torch/metatensor/torch/atomistic/model.py
+++ b/python/metatensor-torch/metatensor/torch/atomistic/model.py
@@ -483,6 +483,19 @@ class MetatensorAtomisticModel(torch.nn.Module):
         except RuntimeError as e:
             raise RuntimeError("could not convert the module to TorchScript") from e
 
+        if self._capabilities.length_unit == "":
+            warnings.warn(
+                "No `length_unit` was provided for the model.",
+                stacklevel=1,
+            )
+
+        for name, output in self._capabilities.outputs.items():
+            if output.unit == "":
+                warnings.warn(
+                    f"No units were provided for output {name!r}.",
+                    stacklevel=1,
+                )
+
         # TODO: can we freeze these?
         # module = torch.jit.freeze(module)
 
--- a/python/metatensor-torch/metatensor/torch/atomistic/model.py
+++ b/python/metatensor-torch/metatensor/torch/atomistic/model.py
@@ -490,11 +490,14 @@ class MetatensorAtomisticModel(torch.nn.Module):
             )
 
         for name, output in self._capabilities.outputs.items():
-            if output.unit == "":
-                warnings.warn(
-                    f"No units were provided for output {name}.",
-                    stacklevel=1,
-                )
+            # TODO: coordinate a list of standard outputs needing
+            # unit checks, should also be consistent with `outputs.py`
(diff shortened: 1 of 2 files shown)


python/metatensor-torch/metatensor/torch/atomistic/model.py
```diff
@@ -483,6 +483,19 @@ def save(self, file: Union[str, Path], collect_extensions: Optional[str] = None)
...
+                "No length unit was provided for the model.",
+                stacklevel=1,
+            )
+
+        for name, output in self._capabilities.outputs.items():
+            if output.unit == "":
+                warnings.warn(
+                    f"No units were provided for output {name}.",
+                    stacklevel=1,
+                )
```
Luthaf, line 497: There are some outputs where no units is expected ("features" output), and then we should not warn. 

At the same time, non-standard outputs (anything with `::` inside the name) might also not require units

SanggyuChong: fair, do we want to check only for a certain set of outputs then? (energy, forces, dipoles, ...)

Luthaf: yes, I would only check for "standard outputs" (like we already do with `check_consistency=True`)

SanggyuChong: addressed with a custom list of standard outputs for now, and left a TODO in case we want to standardize this across the code system

PicoCentauri:
This is very nice! If this is merged I can finally remove our custom export implementation in metatrain which just performs these checks xD

https://github.com/metatensor/metatrain/blob/6af1bed826252f35e3be4c7242f6ffba9d2b89d4/src/metatrain/utils/export.py#L16-L48

python/metatensor-torch/tests/atomistic/model.py
```diff
@@ -60,7 +60,7 @@ def model():
         atomic_types=[1, 2, 3],
         interaction_range=4.3,
         outputs={
-            "tests::dummy::long": ModelOutput(
+            "energy": ModelOutput(
```
Luthaf, line 63: this was intentionally using a non-standard output, to test these =). Could you create a separate model with different metadata instead?

Luthaf (approved):
Thanks a lot @SanggyuChong!

PR #765, opened 2024-10-24
add better error message for assign

### LINKED ISSUE #40: Better error message when trying to assign to block.values directly

DavideTisi:
fixes #40 

So far i just want to trigger CI that on my PC do not work

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2101037761.zip)


metatensor-torch/src/block.cpp
```diff
@@ -158,6 +158,13 @@ torch::Tensor TensorBlockHolder::values() const {
     return wrapper->tensor();
 }
 
+void TensorBlockHolder::set_values(const torch::Tensor& new_values) {
+        throw std::runtime_error(
+            "Direct assignment to `values` is not allowed. "
+            "Please use block.values().copy_(new_values) instead."
```
Luthaf, line 164: why not `block.values[:] = new_values`? Since this error will be seen from Python, it might make sense to use the Python syntax as well!

python/metatensor-core/tests/blocks.py
```diff
@@ -469,6 +470,28 @@ def test_to():
...
+        match=re.escape(
+            "Direct assignment to `values` is not allowed. "
+            "Please use block.values[:] = new_values instead."
+        ),
+    ):
+        block.values = np.array([[4, 5, 6]])
+
+    # Check that setting with slice assignment works correctly
+    block.values[:] = np.array([[4, 5, 6]])
+    assert np.allclose(block.values, np.array([[4, 5, 6]]))
```
Luthaf, line 492: could you add the same test in metatensor-torch?

metatensor-torch/src/block.cpp
```diff
@@ -158,6 +158,13 @@ torch::Tensor TensorBlockHolder::values() const {
     return wrapper->tensor();
 }
 
+void TensorBlockHolder::set_values(const torch::Tensor& new_values) {
+        throw std::runtime_error(
+            "Direct assignment to `values` is not allowed. "
```
Luthaf, line 163:
```suggestion
            "Direct assignment to `values` is not possible. "
```

python/metatensor-core/metatensor/block.py
```diff
@@ -264,6 +264,13 @@ def values(self) -> Array:
 
         return mts_array_to_python_array(self._raw_values, parent=self)
 
+    @values.setter
+    def values(self, new_values):
+        raise AttributeError(
+            "Direct assignment to `values` is not allowed. "
```
Luthaf, line 270:
```suggestion
            "Direct assignment to `values` is not possible. "
```

python/metatensor-core/tests/blocks.py
DavideTisi: I try to do it in torch but for torch I get this error that i do not know how to override:

```
In [4]: block.values=torch.tensor([[4., 5., 6.]])
RuntimeError                              Traceback (most recent call last)
Cell In[4], line 1
----> 1 block.values=torch.tensor([[4., 5., 6.]])

RuntimeError: can't set attribute
```

DavideTisi: chatty fixed the problem after a bit of inputs

Luthaf (approved):

PR #766, opened 2024-10-24
Add `is_atomistic_model` function

### LINKED ISSUE #697: Check if model is metatensor atomistic model?
I think it would be good to include a function to check if a torch model is a metatensor atomistic model. What makes it difficult is when the model is torchscripted. For now, I'm using this:
```
def is_exported(model: Any) -> bool:
    """Check if a model has been exported to a MetatensorAtomisticModel.

    :param model: The model to check
    :return: :py:obj:`True` if the ``model`` has been exported, :py:obj:`False`
        otherwise.
    """
    # If the model is saved and loaded again, its type is RecursiveScriptModule
    if isinstance(model, MetatensorAtomisticModel):
        return True
    try:
        model._module
        model._capabilities
        model._metadata
        return True
    except:
        return False
```
(note that `getattr` doesn't seem to work on torchscripted models)
@Luthaf

PicoCentauri:
Fixes #697

As stated in the title this adds a simple function to check if a loaded module is a good or a bad `torch.nn.Module`.

I need this for https://github.com/metatensor/metatrain/pull/352

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2099655165.zip)

### DIFF
--- a/python/metatensor-torch/metatensor/torch/atomistic/model.py
+++ b/python/metatensor-torch/metatensor/torch/atomistic/model.py
@@ -45,6 +45,30 @@ def load_atomistic_model(path, extensions_directory=None) -> "MetatensorAtomisti
     return torch.jit.load(path)
 
 
+def is_atomistic_model(module: torch.nn.Module) -> bool:
+    """
+    Check if a loaded model is a metatensor atomistic model.
+
+    :param model: model to check
+    :raises TypeError: if the model is not a :py:class:`torch.nn.Module`
+    :return: ``True`` if the ``model`` has been exported, ``False`` otherwise.
+    """
+    if not isinstance(module, torch.nn.Module):
+        raise TypeError(
+            f"`module` should be a torch.nn.Module, not {type(module).__name__}"
+        )
+
+    if isinstance(module, MetatensorAtomisticModel):
+        return True
+    elif (
+        isinstance(module, torch.jit._script.RecursiveScriptModule)
+        and module.original_name == "MetatensorAtomisticModel"
+    ):
+        return True
+    else:
+        return False
+
+
 class ModelInterface(torch.nn.Module):
     """
     Interface for models that can be used with :py:class:`MetatensorAtomisticModel`.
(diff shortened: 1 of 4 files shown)


python/metatensor-torch/metatensor/torch/atomistic/model.py
```diff
@@ -45,6 +45,30 @@ def load_atomistic_model(path, extensions_directory=None) -> "MetatensorAtomisti
     return torch.jit.load(path)
 
 
+def is_atomistic_model(module: torch.nn.Module) -> bool:
+    """
+    Check if a loaded model is an metatensor atomistic model.
```
frostedoyster, line 50:
```suggestion
    Check if a loaded model is a metatensor atomistic model.
```

python/metatensor-torch/metatensor/torch/atomistic/model.py
```diff
@@ -45,6 +45,30 @@ def load_atomistic_model(path, extensions_directory=None) -> "MetatensorAtomisti
...
+
+    :param model: model to check
+    :raises TypeError: if the model is not a :py:class:`torch.nn.Module`
+    :return: ``True`` if the ``model`` has been exported, ``False`` otherwise.
+    """
+    if not isinstance(module, torch.nn.Module):
+        raise TypeError(
+            f"`module` should be a torch.nn.Module, not {type(module).__name__}"
+        )
```
frostedoyster, line 59: By making this check, you're assuming that the model is not "exported" (torchscripted), unless something has changed in torch.

IIRC, after torchscripting, a model becomes a `ScriptObject`

python/metatensor-torch/tests/atomistic/model.py
```diff
@@ -270,6 +271,26 @@ def test_access_module(tmpdir):
...
+    model = FullModel()
+    model.train(False)
+
+    capabilities = ModelCapabilities(
+        interaction_range=0.0,
+        supported_devices=["cpu"],
+        dtype="float64",
+    )
+    atomistic = MetatensorAtomisticModel(model, ModelMetadata(), capabilities)
+    scripted_atomistic = torch.jit.script(atomistic)
```
frostedoyster, line 284: Here I think we should also test that it works with `.save()` and then loading it again with `metatensor.torch.load_atomistic_model` (which is the way we're supposed to handle these objects)

frostedoyster (approved):

PR #770, opened 2024-10-24
Added an option for strict neighbor list

ceriottm:
Adds a flag to the neighbor list classes to signal whether the neighbor list is "strict" - that is if it guarantees not to pass atoms beyond the cutoff. 

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2116989539.zip)

### DIFF
--- a/python/examples/atomistic/3-atomistic-model-with-nl.py
+++ b/python/examples/atomistic/3-atomistic-model-with-nl.py
@@ -163,22 +163,47 @@ print(neighbors)
 # %%
 #
 # The data and metadata inside the ``neighbors`` object do not contain information about
-# the ``cutoff`` and whether this is a full or half neighbor list. To account for this,
-# metatensor neighbor lists are always stored together with
-# :py:class:`NeighborListOptions`. For our system, these options can be saved as
+# the ``cutoff``, whether this is a full or half neighbor list, and whether it is
+# restricted to distances strictly below the cutoff. To account for this, metatensor
+# neighbor lists are always stored together with :py:class:`NeighborListOptions`. For
+# example, these options can be defined with
 
-options = NeighborListOptions(cutoff=5.0, full_list=True)
+options = NeighborListOptions(cutoff=5.0, full_list=True, strict=True)
 
 # %%
 #
-# We set ``full_list=True`` because ASE computes a “full” list (as opposed to a half
-# list) where each `i, j` pair appears twice, stored once as `i, j` and once as `j, i`.
+# ``full_list=True`` will give us a list where each `i, j` pair appears twice, stored
+# once as `i, j` and once as `j, i`. ``full_list=False`` would give us a half list, with
+# each pair only stored once. Depending on your model, either of these can be more
+# convenient or faster.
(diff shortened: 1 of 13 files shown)


python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -163,11 +163,12 @@
 # %%
 #
 # The data and metadata inside the ``neighbors`` object do not contain information about
-# the ``cutoff`` and whether this is a full or half neighbor list. To account for this,
+# the ``cutoff``, whether this is a full or half neighbor list, and whether it is
+# restricted to distances strictly below the cutoff (strict). To account for this,
```
Luthaf, line 167:
```suggestion
# restricted to distances strictly below the cutoff (``strict=True``). To account for this,
```

metatensor-torch/tests/neighbor-checks/slanted-loose.json
```diff
@@ -0,0 +1,207 @@
...
+        "cell": [
+            [4.26, -2.45951215, 0.0],
+            [2.13, 1.22975607, 0.0],
+            [0.0, 0.0, 50.0]
+        ]
+    },
+    "options": {
+        "cutoff": 4.1,
+        "full_list": false,
+        "strict": false
```
Luthaf, line 19: I'm not sure what this is intended to test? The set of pairs included in a non-strict NL will depend a lot on the actual implementation. I think these files are more about checking the strict case, and in the non-strict case the set of `expected-neighbors` is the minimal set of pairs that should be included.

python/examples/atomistic/3-atomistic-model-with-nl.py
```diff
@@ -163,11 +163,12 @@
 # %%
 #
 # The data and metadata inside the ``neighbors`` object do not contain information about
-# the ``cutoff`` and whether this is a full or half neighbor list. To account for this,
-# metatensor neighbor lists are always stored together with
+# the ``cutoff``, whether this is a full or half neighbor list, and whether it is
+# restricted to distances strictly below the cutoff (``strict=True``). To account
+# for this, metatensor neighbor lists are always stored together with
```
Luthaf, line 168: We should have a small explanation about when someone would want to use non-strict list & how to properly hand them.

.github/workflows/python-tests.yml
```diff
@@ -22,7 +22,7 @@ jobs:
             torch-version: "1.12.*"
             numpy-version-pin: "<2.0"
           - os: ubuntu-20.04
-            python-version: "3.9"
+            python-version: "3.10"
```
Luthaf, line 25: @ceriottm could you please not force push back changes I undid?

Luthaf (approved):
This should be good to go!

.github/workflows/python-tests.yml
ceriottm: Don't know how this happened TBH. ohshitgit I guess

PR #784, opened 2024-11-05
Make sure to serialize NeighborListOptions.requestors

Luthaf:
This is used among other by rascaline to know which NL it requested and convert them to the internal format. It looks like I forgot to add serialization for these, meaning that the models using rascaline would only re-use NL when running in pure Python mode, and not through TorchScript. In the other cases, rascaline would recompute the NL, hiding the issue, until someone tried to run the model in LAMMPS where we can not recompute the NL.

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2147405572.zip)


abmazitov (approved):
Yes, it worked for me, and the exported GAP model in the end produces correct energies. So @Luthaf as you fell it is ready to be merged, do it

Issue #548, opened 2024-03-12
`save` does not support pathlib objects


PicoCentauri:
When trying to provide a `Path` to the save function as in this example
```python
import metatensor
from pathlib import Path
import numpy as np

label = metatensor.Labels("_", np.array([[0]]))
block = metatensor.TensorBlock(samples=label, components=[], properties=label, values=np.array([[0]]))
tensor = metatensor.TensorMap(label, [block])

file = Path("foo.bar")
metatensor.io.save(file, tensor)
```

I get an `AttributeError: 'PosixPath' object has no attribute 'endswith'`. Which basically due to these lines

https://github.com/lab-cosmo/metatensor/blob/439548812ac386a593f2b071c08874d81579f39a/python/metatensor-core/metatensor/io.py#L406-L407

Neither has a Path object an attribute `.endswith` nor will the `+=` operator work.

To solve this we should maybe first convert the file into a string if it is a `Path` object or create a different logic branch for the Path object.

Luthaf:
Steps to solve this:

- Add a new test trying to use a `Path` object in with the different `save` functions in metatensor (both `metatensor.io.save` and atomistic model's save); and check that they fail with the above error
- change the code, likely converting from `Path` to `str` as needed early in the function
- check that the tests now pass
- update the type annotations and documentation to show that we now support `Path` objects

tulga-rdn:
I'm on it

tulga-rdn:
I'm on it

PicoCentauri:
Closed via #789

tulga-rdn:
@PicoCentauri can I have a sticker now? 😅

PicoCentauri:
Sure, tomorrow in my office 😁

PR #789, opened 2024-11-10
Fix io

tulga-rdn:
Fixes an IO problem that metatensor.io.save does not support pathlib objects, as stated in #548 (currently a draft PR because I couldn't write the tests)

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2213524310.zip)

### DIFF
--- a/python/metatensor-core/metatensor/io/_block.py
+++ b/python/metatensor-core/metatensor/io/_block.py
@@ -189,6 +189,7 @@ def _save_block(
     assert isinstance(block, TensorBlock)
 
     if isinstance(file, (str, pathlib.Path)):
+        file = str(file)
         if not file.endswith(".npz"):
             file += ".npz"
             warnings.warn(
@@ -201,12 +202,8 @@ def _save_block(
         np.savez(file, **all_entries)
     else:
         lib = _get_library()
-        if isinstance(file, (str, pathlib.Path)):
-            if isinstance(file, str):
-                path = file.encode("utf8")
-            elif isinstance(file, pathlib.Path):
-                path = bytes(file)
-
+        if isinstance(file, str):
+            path = file.encode("utf8")
             lib.mts_block_save(path, block._ptr)
         else:
             # assume we have a file-like object
--- a/python/metatensor-core/metatensor/io/_block.py
+++ b/python/metatensor-core/metatensor/io/_block.py
@@ -188,22 +188,30 @@ def _save_block(
 ):
     assert isinstance(block, TensorBlock)
 
-    if isinstance(file, (str, pathlib.Path)):
-        file = str(file)
-        if not file.endswith(".npz"):
-            file += ".npz"
-            warnings.warn(
-                message=f"adding '.npz' extension, the file will be saved at '{file}'",
-                stacklevel=1,
-            )
-
     if use_numpy:
         all_entries = _block_to_dict(block, prefix="", is_gradient=False)
         np.savez(file, **all_entries)
     else:
         lib = _get_library()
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -74,6 +74,7 @@ deps =
 
 changedir = python/metatensor-core
 commands =
+    # install metatensor-operations
```
Luthaf, line 77: We do not install `metatensor-operations` in these tests

python/metatensor-core/metatensor/io/_block.py
```diff
@@ -201,12 +202,8 @@ def _save_block(
...
     else:
         lib = _get_library()
-        if isinstance(file, (str, pathlib.Path)):
-            if isinstance(file, str):
-                path = file.encode("utf8")
-            elif isinstance(file, pathlib.Path):
-                path = bytes(file)
-
+        if isinstance(file, str):
+            path = file.encode("utf8")
```
Luthaf, line 206: So, I would rather keep the logic converting both `str` and Path to bytes, since on some platforms (linux mainly), a file name does not have to be utf-8. 

Then the code handling `.npz` extension can work on `bytes` instead of `str`

tox.ini
tulga-rdn: Oops sorry, I was messing around with some stuff and forgot to remove it

tulga-rdn: I think I removed it

python/metatensor-core/metatensor/io/_block.py
tulga-rdn: Can you check what I did and let me know if you are OK with it? Thanks!

tulga-rdn:
Also, I think the serialization test might be enough to check what I did (pls let me know if it's not)

Luthaf:
The code was broken before, which means that existing tests did not catch the issue, so yes, an additional test would be good!

python/metatensor-core/metatensor/io/_block.py
Luthaf: There is still the same issue. The core of the problem is that there might be some `Path` objects which can NOT be transformed to `str` and/or can NOT have the resulting str utf8 encoded. So we need to work with `bytes` instead.

tulga-rdn:
I think all is OK now (hopefully 😅), can you check @Luthaf ? Thanks and sorry for bothering you with small mistakes 😅

python/metatensor-core/tests/serialization.py
```diff
@@ -193,6 +194,28 @@ def test_save_warning_errors(tmpdir, tensor):
...
+def test_save_pathlib(tmpdir, tensor):
+    # does not have .npz ending and causes warning
+    tmpfile = Path("serialize-test")
+
+    with pytest.warns() as record:
+        with tmpdir.as_cwd():
+            metatensor.save(tmpfile, tensor)
+
+    expected = f"adding '.npz' extension, the file will be saved at '{tmpfile}.npz'"
+    assert str(record[0].message) == expected
```
Luthaf, line 206: You can use the same syntax for `pytest.raises` and `pytest.warns`

```suggestion
    expected = f"adding '.npz' extension, the file will be saved at '{tmpfile}.npz'"
    with tmpdir.as_cwd():
        with pytest.warns(UserWarning, match=expected):
            metatensor.save(tmpfile, tensor)
```

Luthaf (approved):
Thanks @tulga-rdn, I'll merge this once CI is happy

PR #790, opened 2024-11-11
Make dataset indexing O(1)

### LINKED ISSUE #788: Weird timing of `metatensor.learn.data.Dataset`
I've had some problems with this class on large dataset. This script shows the same behavior
```
from metatensor.learn import Dataset
import time
from metatensor.torch.atomistic import System
import torch
import tqdm

systems = [
    System(
        positions=torch.tensor([[0.0, 0.0, 0.0]]),
        types=torch.tensor([1]),
        cell=10.0*torch.eye(3),
        pbc=torch.tensor([True, True, True])
    )
    for _ in range(1000000)
]

dataset = Dataset(system=systems, energy=list(range(1000000)))
start = time.time()

for sample in tqdm.tqdm(dataset):
    sample["system"]
    sample["energy"]

end = time.time()
print("Time taken: ", end - start)
```

Essentially, the iterator starts fast (more than 10000 iterations per second), but, by the end, it's slower by a factor of almost 1000. I wonder if this can be fixed, as it makes it difficult to use on large datasets

Luthaf:
It was accidentally O(n), since it was checking whether the index was in bounds by checking if it was in a list (with `idx not in list`), which uses a linear scan. Since all indexes are contiguous anyway, this is replaced by `idx < 0 and idx >= len(self)`.

Fix #788, thanks @frostedoyster for the detailed reproduction, it really helped identifying the issue!

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2170635027.zip)

### DIFF
--- a/python/metatensor-learn/metatensor/learn/data/dataset.py
+++ b/python/metatensor-learn/metatensor/learn/data/dataset.py
@@ -62,8 +62,6 @@ class _BaseDataset(torch.utils.data.Dataset):
                         f"'{size_arg_name}' ({size}) and data fields: ({field_sizes})"
                     )
 
-        # Define an internal numeric index list as a range from 0 to size
-        self._indices = list(range(size))
         self._field_names = field_names
         self._data = data
         self._size = size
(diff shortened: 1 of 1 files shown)


frostedoyster:
Thank you for the fix!

Issue #760, opened 2024-10-24
Rename neighbor list and options


ceriottm:
With the introduction of a flag for "strict" neighbor lists, it makes sense to reconsider a bit how flags are set and interpreted. It would make sense to rename:
* Neighbor list -> Pair list (since now this is not supposed to be strict "neighbors"
* full -> half (the natural behavior is that all pairs are listed, and "half" is opt-in
* maybe change the name of "strict" if we can find something more telling

PR #797, opened 2024-11-29
Implement serialization for `System`

frostedoyster:
Replaces #673, uses a new `mta` format

----
Edit by @Luthaf:

The idea here is to define the file format for System serialization, but only implement it in Python for now. A lower level implementation will be added later.

---

# Reviewer checklist

----
 📚 [Download documentation preview for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2343299320.zip)

⚙️ [Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/metatensor/metatensor/actions/artifacts/2343466943.zip)

### DIFF
--- a/python/metatensor_torch/metatensor/torch/atomistic/io.py
+++ b/python/metatensor_torch/metatensor/torch/atomistic/io.py
@@ -1,0 +1,165 @@
+import io
+import os
+import warnings
+import zipfile
+from pathlib import Path
+from typing import Union
+
+import numpy as np
+import torch
+
+import metatensor.torch
+
+from . import NeighborListOptions, System
+
+
+def save(file: Union[str, Path, io.BytesIO], system: System) -> None:
+    """Save a System object to a file.
+
+    The provided System must contain float64 data and be on the CPU device.
+
+    The saved file will be a zip archive containing the following files:
+
+    - ``types.npy``, containing the atomic types in numpy's NPY format;
+    - ``positions.npy``, containing the systems' positions in numpy's NPY format;
+    - ``cell.npy``, containing the systems' cell in numpy's NPY format;
+    - ``pbc.npy``, containing the periodic boundary conditions in numpy's NPY format;
+
+    For each neighbor list in the System object, the following files will be saved
+    (where ``{nl_idx}`` is the index of the neighbor list):
+
+    - ``pairs/{nl_idx}/options.json``: the ``NeighborListOptions`` object
+      converted to a JSON string.
+    - ``pairs/{nl_idx}/data.mts``: the neighbor list ``TensorBlock`` object
+
+    For each extra data in the System object, the following file will be saved (where
+    ``{name}`` is the name of the extra data):
+
+    - ``data/{name}.mts``: The extra data ``TensorMap``
+
+    :param file: The path (or file-like object) to save the System to.
+    :param system: The System object to save.
+    """
+
+    if isinstance(file, (str, Path)):
+        if not file.endswith(".mta"):
+            raise ValueError("The provided path must have the `.mta` extension.")
+
+    if _is_system(system):
+        _save_system(file, system)
+    else:
+        raise ValueError("`system` must be a System object.")
+
+
+def load_system(file: Union[str, Path, io.BytesIO]) -> System:
+    """Load a System object from a file.
+
+    The loaded System object will be on the CPU device and contain float64 data.
+
+    :param file: The path (or file-like object) to load the System object from.
+    :return: The System object.
+    """
+    if isinstance(file, (str, Path)):
+        if not os.path.exists(file):
+            raise FileNotFoundError(f"File not found: {file}")
+        if not file.endswith(".mta"):
+            raise ValueError("The provided path must have the `.mta` extension.")
+
+    if _is_system_mta(file):
+        return _load_system(file)
+    else:
+        raise ValueError(f"File does not contain a valid System object: {file}")
+
+
+def _is_system_mta(path: Union[str, Path, io.BytesIO]) -> bool:
+    zipf = zipfile.ZipFile(path, "r")
+    all_zip_files = zipf.namelist()
+    required_files = ["positions.npy", "cell.npy", "types.npy", "pbc.npy"]
+    for file in required_files:
+        if file not in all_zip_files:
+            return False
+    return True
+
+
+def _load_system(path: Union[str, Path, io.BytesIO]) -> System:
+    # we filter a warning related to the fact that numpy arrays from buffers
+    # are not writable, while torch would like arrays to be writable when
+    # converting them to tensors; this is ok because we then clone the tensor
+    warnings.filterwarnings(
+        "ignore", category=UserWarning, message="The given NumPy array is not writable"
+    )
+
+    with zipfile.ZipFile(path, "r") as zipf:
+        positions = torch.from_numpy(np.load(zipf.open("positions.npy")))
+        cell = torch.from_numpy(np.load(zipf.open("cell.npy")))
+        types = torch.from_numpy(np.load(zipf.open("types.npy")))
+        pbc = torch.from_numpy(np.load(zipf.open("pbc.npy")))
+
+        neighbor_list_options_list = []
+        neighbor_lists = []
+        for path in zipf.namelist():
+            if path.startswith("pairs/") and path.endswith("/options.json"):
+                nl_options = NeighborListOptions(0.0, False, False)
+                nl_options._get_method("__setstate__")(zipf.read(path))
+                neighbor_list_options_list.append(nl_options)
+
+                data_path = path[:-12] + "data.mts"
+                numpy_buffer = np.frombuffer(zipf.read(data_path), dtype=np.uint8)
+                tensor_buffer = torch.from_numpy(numpy_buffer)
+                neighbor_lists.append(metatensor.torch.load_block_buffer(tensor_buffer))
+
+        extra_data_dict = {}
+        for path in zipf.namelist():
+            if path.startswith("data/"):
+                name = os.path.basename(path).replace(".mts", "")
+                numpy_buffer = np.frombuffer(zipf.read(path), dtype=np.uint8)
+                tensor_buffer = torch.from_numpy(numpy_buffer)
+                extra_data_dict[name] = metatensor.torch.load_buffer(tensor_buffer)
+
+    system = System(
+        positions=positions,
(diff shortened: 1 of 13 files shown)


frostedoyster:
@Luthaf I'm having trouble with the tests because the IO tests are executed before the system tests, so they steal the custom data warning from the system tests

python/metatensor-torch/tests/atomistic/io.py
```diff
@@ -0,0 +1,73 @@
...
+
+@pytest.mark.parametrize(
+    "pbc", [[True, True, True], [True, False, True], [False, False, False]]
+)
+@pytest.mark.parametrize("nl_size", [10, 100, 1000])
+@pytest.mark.parametrize("cutoff", [3.5, 5.0])
+@pytest.mark.parametrize("full_list", [True, False])
+@pytest.mark.parametrize("strict", [True, False])
+@pytest.mark.parametrize("extra_data", [False, True])
+@pytest.mark.filterwarnings("ignore:custom data")
```
Luthaf, line 19: I don't think it makes sense to parametrize on everything here, it will just make the tests slower without bringing extra checks. I would rather do one test saving everything possible at the same time.

python/metatensor-torch/metatensor/torch/atomistic/io.py
```diff
@@ -0,0 +1,148 @@
...
+    NeighborListOptions,
+    System,
+)
+
+
+def save(path: Union[str, Path], data: System) -> None:
+    """Save a System object to a file.
+
+    The provided System must contain float64 data and be on the CPU device.
+    Neighbor list requestors will not be saved.
```
Luthaf, line 22: Why not? You should be able to use the serialization of NL options and automatically save everything.

python/metatensor-torch/metatensor/torch/atomistic/io.py
```diff
@@ -0,0 +1,148 @@
...
+
```
Luthaf, line 23: This should contain a definition of the file format used.

python/metatensor-torch/metatensor/torch/atomistic/io.py
```diff
@@ -0,0 +1,148 @@
...
+    :param path: The path to save the System object to.
+    :param data: The System object to save.
+    """
+
+    if not isinstance(data, torch.ScriptObject):
+        raise ValueError("`data` must be a scripted object.")
```
Luthaf, line 29: Could this check be made in `_is_system`?

python/metatensor-torch/metatensor/torch/atomistic/io.py
```diff
@@ -0,0 +1,148 @@
...
+from metatensor.torch import load_block_buffer
+from metatensor.torch import save_buffer as metatensor_torch_save_buffer
+
+from . import (
+    NeighborListOptions,
+    System,
+)
+
+
+def save(path: Union[str, Path], data: System) -> None:
```
Luthaf, line 18:
```suggestion
def save(path: Union[str, Path], system: System) -> None:
```

python/metatensor-torch/metatensor/torch/atomistic/io.py
```diff
@@ -0,0 +1,148 @@
...
+    if not isinstance(data, torch.ScriptObject):
+        raise ValueError("`data` must be a scripted object.")
+
+    if _is_system(data):
+        _save_system(path, data)
+    else:
+        raise ValueError("`data` must be a System object.")
+
+
+def load(path: Union[str, Path]) -> System:
```
Luthaf, line 37:
```suggestion
def load_system(path: Union[str, Path]) -> System:
```

python/metatensor-torch/metatensor/torch/atomistic/io.py
```diff
@@ -0,0 +1,148 @@
...
+
+    neighbor_list_dict = np.load(io.BytesIO(data["neighbor_lists"]), allow_pickle=True)
+    neighbor_list_dict = {
+        key: torch.from_numpy(value) for key, value in neighbor_list_dict.items()
+    }
+
+    extra_data_dict = np.load(io.BytesIO(data["extra_data"]), allow_pickle=True)
+    extra_data_dict = {
+        key: torch.from_numpy(value) for key, value in extra_data_dict.items()
+    }
```
Luthaf, line 81: Pickled dict is not something we will be able to load easily from C++ down the line. I would rather do something using file path, like 

```
neighbors/
                      0/
                               options.json
                               values.npz
                      1/
                               options.json
                               values.npz
                      2/
                               options.json
                               values.npz
```

Where `values.npz` is the TensorBlock serialization

For data we should be able to do 


```
data/
        <name-1>.npz
        <name-2>.npz
        <name-3>.npz
```

This might mean we will have to use the `zipfile` module directly, instead of relying on `np.load`

python/metatensor-torch/metatensor/torch/atomistic/io.py
```diff
@@ -0,0 +1,145 @@
...
+    with zipfile.ZipFile(path, "w") as zipf:
+        for nl_idx, nl_options in enumerate(system.known_neighbor_lists()):
+            zipf.writestr(f"neighbor_lists/{nl_idx}/options.json", nl_options.to_json())
+            nl = system.get_neighbor_list(nl_options)
+            tensor_buffer = metatensor_torch_save_buffer(nl)
+            numpy_buffer = tensor_buffer.numpy()
+            nl_buffer = io.BytesIO()
+            np.save(nl_buffer, numpy_buffer)
+            nl_buffer.seek(0)
+            zipf.writestr(f"neighbor_lists/{nl_idx}/data.npy", nl_buffer.read())
```
Luthaf, line 128: This creates a npy file containing the bytes of the metatensor npz:

```
$ file neighbor_lists/0/data.npy
neighbor_lists/0/data.npy: NumPy data file, version 1.0, description {'descr': '|u1', 'fortran_order': False, 'shape': (5500,), }

$ unzip neighbor_lists/0/data.npy
Archive:  neighbor_lists/0/data.npy
warning [neighbor_lists/0/data.npy]:  128 extra bytes at beginning or within zipfile
```

Was this intentional? I guess it would make sure the file is a valid npz, at the cost of making reading it harder.

I think I would rather have a format which is not quite NPZ but easier to handle, so using something like this instead:

```suggestion
            tensor_buffer = metatensor_torch_save_buffer(nl)
            zipf.writestr(f"neighbor_lists/{nl_idx}/data.npz", tensor_buffer.numpy().tobytes())
```

We are already saving JSON in there anyway, so it is not going to be a valid npz. But we could maybe change the extension used by all of these, to communicate it more clearly to the users. `.mta` maybe, if it is not already taken? Or `.mts` and also change the default on metatensor -non-atomistic save/load?

python/metatensor-torch/metatensor/torch/atomistic/io.py
```diff
@@ -0,0 +1,145 @@
...
+
+        for key in system.known_data():
+            data = system.get_data(key)
+            tensor_buffer = metatensor_torch_save_buffer(data)
+            numpy_buffer = tensor_buffer.numpy()
+            data_buffer = io.BytesIO()
+            np.save(data_buffer, numpy_buffer)
+            data_buffer.seek(0)
+            zipf.writestr(f"extra_data/{key}.npy", data_buffer.read())
```
Luthaf, line 137: Same as above

Luthaf:
> I'm having trouble with the tests because the IO tests are executed before the system tests, so they steal the custom data warning from the system tests

I'm not sure what's happening, the code should warn once per custom data name, so if you use a different name from the other tests, both should warn.

metatensor-torch/src/register.cpp
```diff
@@ -353,7 +353,9 @@ TORCH_LIBRARY(metatensor, m) {
             [](const std::string& data) -> NeighborListOptions {
                 return NeighborListOptionsHolder::from_json(data);
             }
-        );
+        )
+        .def("to_json", &NeighborListOptionsHolder::to_json)
+        .def_static("from_json", &NeighborListOptionsHolder::from_json);
```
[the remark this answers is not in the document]
frostedoyster: I tried quite a bit to make `__getstate__` and `__setstate__` work but I couldn't. In particlular, `__setstate__` didn't work at all

python/metatensor-torch/metatensor/torch/atomistic/io.py
```diff
@@ -0,0 +1,145 @@
...
+def _load_system(path: Union[str, Path]) -> System:
+    with zipfile.ZipFile(path, "r") as zipf:
+        positions = torch.from_numpy(np.load(zipf.open("positions.npy")))
+        cell = torch.from_numpy(np.load(zipf.open("cell.npy")))
+        types = torch.from_numpy(np.load(zipf.open("types.npy")))
+        pbc = torch.from_numpy(np.load(zipf.open("pbc.npy")))
+
+        neighbor_list_options_list = []
+        neighbor_lists = []
+        for nl_idx in range(len(zipf.namelist())):
```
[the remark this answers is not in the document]
frostedoyster: I searched a bit but apparently you can't!

frostedoyster:
The names for the extra system data in the two tests are already different, I don't know what the issue is

metatensor-torch/src/register.cpp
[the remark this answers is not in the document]
Luthaf: I managed to make it work, by using `options._get_method("__setstate__")` to access the underlying function.

python/metatensor_torch/metatensor/torch/atomistic/io.py
```diff
@@ -0,0 +1,171 @@
...
+    - ``positions.npy``, containing the systems' positions in numpy's NPY format;
+    - ``cell.npy``, containing the systems' cell in numpy's NPY format;
+    - ``pbc.npy``, containing the periodic boundary conditions in numpy's NPY format;
+
+    For each neighbor list in the System object, the following files will be saved
+    (where ``{nl_idx}`` is the index of the neighbor list):
+
+    - ``pairs/{nl_idx}/options.json``: the ``NeighborListOptions`` object
+      converted to a JSON string.
+    - ``pairs/{nl_idx}/data.npz``: the neighbor list ``TensorBlock`` object
```
Luthaf, line 34: @frostedoyster I changed this to reflect the upcoming renaming in #760

Luthaf (approved):

PR #833, opened 2025-01-30
Try to avoid checks for empty Labels

### LINKED ISSUE #832: Different behavior of `metatensor.Labels` and `metatensor.torch.Labels` for empty `values`
I encountered this issue when trying to use `operations.reduce_over_samples` for empty `TensorBlock`s.

The (edge case, I guess) issue is that this code runs fine

```
import metatensor
import numpy as np
metatensor.Labels(["a"], values=np.empty(0))

>>> Labels(
        a
    )
```

but this one raises an error

```
import metatensor.torch
import torch
metatensor.torch.Labels(["a"], values=torch.empty(0))

ValueError                                Traceback (most recent call last)
Cell In[53], line 3
      1 import metatensor.torch as mts_torch
----> 3 mts_torch.Labels(["a"], values=torch.empty(0))

ValueError: Labels values must be an Tensor of 32-bit integers
```

The error is not raised when explicitly casting to `int` and reshaping:

```
import metatensor.torch
import torch
metatensor.torch.Labels(["a"], values=torch.empty(0).to(torch.int32).reshape(-1, 1))

>>> Labels(
        a
    )
```

It can be fixed, e.g., defining a `_dispatch.zeros_int_like` function in `operations`, but I wonder what @Luthaf thinks about it (maybe it's better to change `metatensor.torch.Labels` instead).

ppegolo:
Fixes #832 

# Reviewer checklist

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2528575853.zip)

### DIFF
--- a/metatensor-torch/src/labels.cpp
+++ b/metatensor-torch/src/labels.cpp
@@ -18,7 +18,7 @@ static torch::Tensor normalize_int32_tensor(torch::Tensor values, size_t shape_l
             C10_THROW_ERROR(ValueError,
                 context + " must be a Tensor of 32-bit integers"
         );
-        }
+    }
 
     if (values.sizes().size() != shape_length) {
         C10_THROW_ERROR(ValueError,
@@ -121,12 +121,10 @@ LabelsHolder::LabelsHolder(torch::IValue names, torch::Tensor values):
     values_(normalize_int32_tensor(values, 2, "Labels values")),
     labels_(torch::nullopt)
 {
-    if (values_.numel() > 0) {
-        if (values_.sizes()[1] != names_.size()) {
-            C10_THROW_ERROR(ValueError,
-                "invalid Labels: the names must have an entry for each column of the array"
-            );
-        }
+    if (values_.sizes()[1] != names_.size()) {
+        C10_THROW_ERROR(ValueError,
+            "invalid Labels: the names must have an entry for each column of the array"
+        );
     }
 
     labels_ = metatensor::Labels(
--- a/metatensor-torch/src/labels.cpp
+++ b/metatensor-torch/src/labels.cpp
@@ -13,16 +13,18 @@ using namespace metatensor_torch;
 /// Check that `values` is a `shape_length`-dimensional array of 32-bit
 /// integers, or convert it to be so.
 static torch::Tensor normalize_int32_tensor(torch::Tensor values, size_t shape_length, const std::string& context) {
-    if (!torch::can_cast(values.scalar_type(), torch::kI32)) {
-        C10_THROW_ERROR(ValueError,
-            context + " must be an Tensor of 32-bit integers"
-       );
-    }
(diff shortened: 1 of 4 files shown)


ppegolo:
Not sure what I broke in the tests

Luthaf:
It looks like we need to update the definition of `Labels.empty` in metatensor-core. You can run the tests on your machine to check that there is nothing else that needs updating!

metatensor-torch/src/labels.cpp
```diff
@@ -13,17 +13,19 @@ using namespace metatensor_torch;
 /// Check that `values` is a `shape_length`-dimensional array of 32-bit
 /// integers, or convert it to be so.
 static torch::Tensor normalize_int32_tensor(torch::Tensor values, size_t shape_length, const std::string& context) {
-    if (!torch::can_cast(values.scalar_type(), torch::kI32)) {
-        C10_THROW_ERROR(ValueError,
-            context + " must be an Tensor of 32-bit integers"
-       );
+    if (values.numel() > 0) {
```
Luthaf, line 16: I think this should also be changed to do something like `values = np.empty(values.shape, dtype=np.int32)`. This is a bit painful to do with the C++ torch API, so I'm happy to help!

ppegolo: I think I've done something like that

metatensor-torch/src/labels.cpp
```diff
@@ -13,10 +13,15 @@ using namespace metatensor_torch;
 /// Check that `values` is a `shape_length`-dimensional array of 32-bit
 /// integers, or convert it to be so.
 static torch::Tensor normalize_int32_tensor(torch::Tensor values, size_t shape_length, const std::string& context) {
+    // if the tensor is empty, make sure the shape is (0, len(sample_names))
+    if (!values.numel()) {
+        values = torch::empty(values.sizes(), torch::TensorOptions().dtype(torch::kInt32));
```
Luthaf, line 18:
```suggestion
        values = torch::empty(values.sizes(), torch::TensorOptions().dtype(torch::kInt32).device(values.device()));
```

we also need the same device

metatensor-torch/tests/labels.cpp
```diff
@@ -38,7 +38,8 @@ TEST_CASE("Labels") {
         CHECK(empty->size() == 0);
         CHECK(empty->count() == 0);
 
-        CHECK_THROWS_WITH(LabelsHolder::create({}, {{}, {}, {}}), "invalid parameter: can not have labels.count > 0 if labels.size is 0");
+        // TODO
```
Luthaf, line 41: needs to be updated =)

ppegolo: Oops, forgot to uncomment it

Luthaf (approved):

Issue #848, opened 2025-02-06
HTML representation of `TensorMap` and `Tensorblock`


pfebrer:
Related to #847 

For environments that support html representations of objects (e.g. jupyter notebooks), implement an html representation, which allows to make the structure clearer than with a simple string representation. This is what `xarray` does, e.g. in https://docs.xarray.dev/en/stable/examples/weather-data.html

Issue #847, opened 2025-02-06
Improve string representation of `TensorMap` and `TensorBlock`


pfebrer:
The idea would be to change the string representations so that the structure of the data is easier to understand.

PR #856, opened 2025-02-13
Improve string and HTML representations.

pfebrer:
Addresses #847 and #848 by modifying slightly the string representation of `TensorMap` and `TensorBlock` and introducing a new HTML representation.

If you agree with the way it looks I'll finish some details and then we would need to port the string representations to the C++ api as well (which I don't know exactly how to test yet :sweat_smile: )

This is how it looks with the tensormap from the tutorials:

```python
import metatensor

tensor = metatensor.load("spherical-expansion.mts")
``` 

TensorBlock html:

Screenshot from 2025-02-13 14-29-48

TensorBlock str:

Screenshot from 2025-02-13 14-37-18

TensorMap html (blocks expand by clicking):

Screenshot from 2025-02-13 14-30-06

TensorMap str:

Screenshot from 2025-02-13 14-37-21


python/metatensor_core/metatensor/_html_stylesheet.py
```diff
@@ -0,0 +1,167 @@
+# CSS styles for the HTML representation
```
Luthaf, line 1: This should not be in a separate file if we want to make it easy to use from C++ as well

Luthaf (commented):
This should also handle Labels in a similar ways. Here are a couple of tweak suggestions:

- TensorBlock str should remove the `<>` and maybe be simplified like this:

```
metatensor.TensorBlock:
    values shape: (...)
    samples: [system, atom]
    components: [xyz, o3_mu]
    properties: [n1, n2, ...]
```


- TensorMap str: I would also remove `<>`, and remove the `4` in `4 keys: (names)`. Just having `keys: (names)` like we do for samples/components/properties would be enough


Finally, all of them needs to check what happen when you have 100+ entries in Labels (key/samples/properties). We will likely want to truncate the representation at some point and/or add scroll bars.

HaoZeke:
@pfebrer this is getting a bit stale ^_^

Luthaf:
Now in #1170

PR #858, opened 2025-02-16
Update setuptools build dependency

frostedoyster:
Setuptools 68, 69 and 70 fail to build the library

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ changelog = "https://docs.metatensor.org/latest/core/CHANGELOG.html"
 
 [build-system]
 requires = [
-    "setuptools >=68",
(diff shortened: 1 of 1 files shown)


PicoCentauri (commented):
I think we use setuptools in various `toml` files:

```
./python/metatensor_core/pyproject.toml:    "setuptools >=75",
./python/metatensor_operations/pyproject.toml:    "setuptools >=68",
./python/metatensor_learn/pyproject.toml:    "setuptools >=68",
./python/metatensor_torch/pyproject.toml:    "setuptools >=75",
```

Maybe we should use a consistent version of 75 in all of these files?

PicoCentauri (approved):

PR #863, opened 2025-02-17
Fix bug

frostedoyster:
Avoids failure if no `featomic_torch.libs/` can be found

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2607795409.zip)

### DIFF
--- a/python/metatensor_torch/metatensor/torch/atomistic/_extensions.py
+++ b/python/metatensor_torch/metatensor/torch/atomistic/_extensions.py
@@ -32,8 +32,10 @@ def _featomic_deps_path():
         for prefix in site_packages:
             # find where featomic is and look inside the `featomic_torch.libs` directory
             if os.path.exists(os.path.join(prefix, "featomic")):
-                correct_prefix = prefix
-                libs_list += os.listdir(os.path.join(prefix, "featomic_torch.libs"))
+                libs_dir = os.path.join(prefix, "featomic_torch.libs")
+                if os.path.exists(libs_dir):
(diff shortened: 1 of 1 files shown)


python/metatensor_torch/metatensor/torch/atomistic/_extensions.py
```diff
@@ -32,8 +32,10 @@ def _featomic_deps_path():
         for prefix in site_packages:
             # find where featomic is and look inside the `featomic_torch.libs` directory
             if os.path.exists(os.path.join(prefix, "featomic")):
-                correct_prefix = prefix
-                libs_list += os.listdir(os.path.join(prefix, "featomic_torch.libs"))
+                libs_dir = os.path.join(prefix, "featomic_torch.libs")
+                if os.path.exists(libs_dir):
+                    correct_prefix = prefix
```
PicoCentauri, line 37: Are you sure that this will not break further below 

```python
        gomp_dependency = [
            os.path.join(correct_prefix, "featomic_torch.libs", libgomp_path)
        ]
```
here you use `correct_prefix` but it might not be defined because you only define this within this condition

frostedoyster: Good catch, thanks

PicoCentauri:
I did a little cleanup of the function. @frostedoyster please check the lines carefully. If you are happy, I merge and will do another release of metatensor-torch... *sigh*.

PR #867, opened 2025-02-19
Only calculate stress if PBCs are there

frostedoyster:
As in the title, otherwise the stress is +inf or -inf, and warnings are displayed when we convert those to Voigt stresses

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2615423377.zip)

### DIFF
--- a/python/metatensor_torch/metatensor/torch/atomistic/ase_calculator.py
+++ b/python/metatensor_torch/metatensor/torch/atomistic/ase_calculator.py
@@ -295,7 +295,9 @@ class MetatensorCalculator(ase.calculators.calculator.Calculator):
         )
         calculate_energies = "energies" in properties
         calculate_forces = "forces" in properties or "stress" in properties
-        calculate_stress = "stress" in properties or "forces" in properties
+        calculate_stress = (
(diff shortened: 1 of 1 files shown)


python/metatensor_torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -295,7 +295,9 @@ def calculate(
         )
         calculate_energies = "energies" in properties
         calculate_forces = "forces" in properties or "stress" in properties
-        calculate_stress = "stress" in properties or "forces" in properties
+        calculate_stress = (
+            "stress" in properties or "forces" in properties
+        ) and atoms.pbc.all()
```
Luthaf, line 300:
```suggestion
        calculate_stress = "stress" in properties
        if calculate_stress and not atoms.pbc.all():
            warning("stress requested but likely to be wrong ………")
        
        if "forces" in properties and atoms.pbc.all():
            # we have PBC, and since the user requested forces we will run backward, so let's do stress as well
            calculate_stress = True
```

Luthaf (approved):

PR #874, opened 2025-03-14
Fix extensions for sphericart torch

ppegolo:
Similar to #863, but for `sphericart`

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/2754369343.zip)

### DIFF
--- a/python/metatensor_torch/metatensor/torch/atomistic/_extensions.py
+++ b/python/metatensor_torch/metatensor/torch/atomistic/_extensions.py
@@ -58,11 +58,49 @@ def _featomic_deps_path():
     return deps_path
 
 
+def _sphericart_deps_path():
+    import sphericart
+
+    deps_path = [sphericart._c_lib._lib_path()]
+
+    # add libgomp shared library, which is added by cibuildwheel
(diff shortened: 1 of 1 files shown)


ppegolo:
Can't test it with the current version of metatensor-torch as it's incompatible with sphericart-torch, but implementing the fix directly in the site-packages allows to load models with sphericart-torch extensions

frostedoyster (approved):
This code is very similar to the one for rascaline, so it might be worth to extract a common function. However, since we might have to do this for vesin as well, we'll extract that function later

python/metatensor_torch/metatensor/torch/atomistic/_extensions.py
```diff
@@ -58,11 +58,49 @@ def _featomic_deps_path():
     return deps_path
 
 
+def _sphericart_deps_path():
+    import sphericart
+
+    deps_path = [sphericart._c_lib._lib_path()]
```
Luthaf, line 64: This is not needed for sphericart, since the same implementation is duplicated in `libsphericart` and `libsphericart_torch`.

PR #883, opened 2025-03-29
Make the ASE calculator warn for high uncertainties

frostedoyster:
Adds uncertainty warnings to the ASE calculator. The threshold is set to 0.1 eV/atom, which is "chemical accuracy", according to some. The tests need https://github.com/metatensor/lj-test/pull/5 to work correctly

 - ~Issue referenced (for PRs that solve an issue)?~

# Reviewer checklist

### DIFF
--- a/python/metatensor_torch/metatensor/torch/atomistic/ase_calculator.py
+++ b/python/metatensor_torch/metatensor/torch/atomistic/ase_calculator.py
@@ -186,6 +186,10 @@ class MetatensorCalculator(ase.calculators.calculator.Calculator):
         # so we pretend to be able to compute all properties ASE knows about.
         self.implemented_properties = ALL_ASE_PROPERTIES
 
+        self._model_has_energy_uncertainty = (
+            "energy_uncertainty" in self._model.capabilities().outputs
+        )
+
     def todict(self):
         if "model_path" not in self.parameters:
             raise RuntimeError(
@@ -321,6 +325,8 @@ class MetatensorCalculator(ase.calculators.calculator.Calculator):
         with record_function("ASECalculator::prepare_inputs"):
             outputs = _ase_properties_to_metatensor_outputs(properties)
             outputs.update(self._additional_output_requests)
+            if calculate_energy and self._model_has_energy_uncertainty:
+                outputs["energy_uncertainty"] = _get_energy_uncertainty_output()
 
             capabilities = self._model.capabilities()
             for name in outputs.keys():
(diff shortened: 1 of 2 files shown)


python/metatensor_torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -321,6 +325,8 @@ def calculate(
         with record_function("ASECalculator::prepare_inputs"):
             outputs = _ase_properties_to_metatensor_outputs(properties)
             outputs.update(self._additional_output_requests)
+            if calculate_energy and self._model_has_energy_uncertainty:
+                outputs["energy_uncertainty"] = _get_energy_uncertainty_output()
```
Luthaf, line 329: Is this something we always want to do? I see a point for having it on by default, but I feel like users should be able to turn this off

python/metatensor_torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -709,3 +729,12 @@ def _full_3x3_to_voigt_6_stress(stress):
             (stress[0, 1] + stress[1, 0]) / 2.0,
         ]
     )
+
+
+def _get_energy_uncertainty_output():
+    return ModelOutput(
+        quantity="energy",
```
Luthaf, line 736:
```suggestion
        quantity="energy_uncertainty",
```

python/metatensor_torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -397,6 +403,20 @@ def calculate(
...
+            if calculate_energy and self._model_has_energy_uncertainty:
+                uncertainty = outputs["energy_uncertainty"].block().values
+                assert uncertainty.shape == (1, 1)
+                uncertainty = uncertainty.detach().cpu().numpy()[0, 0]
+                if uncertainty / len(atoms) > 0.1:
+                    # 100 meV/atom, which is more than 2 kcal/mol, means that this
+                    # prediction is not chemically accurate
+                    warnings.warn(
+                        "the model's energy uncertainty is larger than 100 meV/atom; "
+                        "this prediction is not chemically accurate",
```
Luthaf, line 416:
```suggestion
                        "the model's energy uncertainty is larger than 100 meV/atom ({uncertainty / len(atoms):.4} meV/atom), "
                        "you might be using it on a structure it can not handle",
```

python/metatensor_torch/metatensor/torch/atomistic/ase_calculator.py
ceriottm: I tend to agree, it's good to have it on-by-default, but it will have a non-zero overhead so maybe one should be able to switch it off

python/metatensor_torch/metatensor/torch/atomistic/ase_calculator.py
ceriottm: Should we make this threshold an option, and say that setting it to zero disables the check?
Also, IDK how hard it would be to get per-atom energies and uncertainties, but I think in many cases it would be a better check than looking at the mean uncertainty.

Luthaf:
Can you re-open this in https://github.com/metatensor/metatomic?

PR #897, opened 2025-04-11
Small fixes to the documentation

RMeli:
Small fixes for the documentation, which I spotted while going through (parts of) it.


Luthaf (approved):
Thanks a lot for this! Just one small change =)

PR #908, opened 2025-04-24
Use an implementation less sensitive to numerical noise for std/var over samples

### LINKED ISSUE #902: NaN in std_over_samples for float32
I encountered a NaN when using `metatensor.torch.std_over_samples` with float32 data. As far as I investigated, the reason is numerical instability in calculation of variance : in this case we have a negative variance => sqrt(-value) returning NaN

Hovewer, with float64 there is no issue. 

I attach a tensormap & notebook to reproduce the issue:

[bug.zip](https://github.com/user-attachments/files/19852861/bug.zip)

Here are the logs from it:

```
loaded tensormap with 5 blocks
for block 0

applying std_over_samples:
with float64 has nans: False
with float32 has nans: True

std at sample = 39 and feature 3 :
float64: 5.0324e-05
float32: nan

feature 3 values for system 39:
float64: [-0.27714047 -0.27715549 -0.27721998 -0.27707845]
float32: [-0.27714047 -0.2771555  -0.27721998 -0.27707845]

manual variance check (for instability):
float64: -0.07681134225468467
float32: -0.07681135088205338

```

Luthaf:
Fix #902 

# Reviewer checklist

### DIFF
--- a/python/metatensor_operations/metatensor/operations/_reduce_over_samples.py
+++ b/python/metatensor_operations/metatensor/operations/_reduce_over_samples.py
@@ -115,39 +115,46 @@ def _reduce_over_samples_block(
 
     block_values = block.values
     other_shape = block_values.shape[1:]
-    values_result = _dispatch.zeros_like(
+    values_sum = _dispatch.zeros_like(
         block_values, shape=(new_samples.shape[0],) + other_shape
     )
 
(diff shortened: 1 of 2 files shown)


DavideTisi (approved):
I see now you only have squared quantities and not difference of squares. 
It look great and also more readable

PR #914, opened 2025-04-28
Fix non-conservative output request in ASE calculator

frostedoyster:


python/metatensor_torch/metatensor/torch/atomistic/ase_calculator.py
```diff
@@ -605,13 +605,13 @@ def _ase_properties_to_metatensor_outputs(self, properties):
             output.explicit_gradients = ["cell"]
 
         metatensor_outputs = {"energy": output}
-        if "forces" in properties and self._non_conservative:
+        compute_forces_and_stresses = "forces" in properties or "stress" in properties
+        if compute_forces_and_stresses and self._non_conservative:
```
Luthaf, line 609: so why do we need to always compute both together?

frostedoyster: We don't but, assuming you're running NPT, calculating them together gives you both for one model evaluation. If you calculate them separately when they're requested, you evaluate twice and pay twice (throwing away the 2x speed-up of non-conservative models)

Luthaf: Because ASE calls `calculate` twice with separate properties?

frostedoyster: Yes, it's here: https://wiki.fysik.dtu.dk/ase/_modules/ase/md/npt.html#NPT

PR #920, opened 2025-05-06
Add `non_blocking` parameter to `to()`

### LINKED ISSUE #838: Allow non-blocking device transfer for `System` and `TensorMap`
When training on a batch of `System`s, potentially a lot of tensors have to be sent to the device (let's say 16 systems with 4 tensors each makes 64 data transfers). At the moment, all transfers are blocking (i.e. `cudaDeviceSynchronize` is called after each transfer). It would probably be faster to allow the user to transfer without the need to synchronize

Luthaf:
Fix #838 (except for `System`, the corresponding change will happen in https://github.com/metatensor/metatrain).

# Reviewer checklist

### DIFF
--- a/python/metatensor_torch/tests/block.py
+++ b/python/metatensor_torch/tests/block.py
@@ -281,7 +281,9 @@ def test_to():
     check_dtype(block, torch.float32)
     check_dtype(block.gradient("g"), torch.float32)
 
-    converted = block.to(dtype=torch.float64)
+    # we use non_blocking=True for some of the calls to `.to` below as a smoke test,
+    # making sure the parameter is accepted by this function
+    converted = block.to(dtype=torch.float64, non_blocking=True)
     check_dtype(converted, torch.float64)
     check_dtype(converted.gradient("g"), torch.float64)
 
@@ -296,7 +298,7 @@ def test_to():
         devices.append(torch.device("cuda"))
 
     for device in devices:
-        moved = block.to(device=device)
+        moved = block.to(device=device, non_blocking=True)
         assert moved.device.type == torch.device(device).type
         assert moved.gradient("g").device.type == torch.device(device).type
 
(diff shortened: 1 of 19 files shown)


HaoZeke (commented):
Implementation and documentation look good, however I'm a bit confused about the testing pattern. Setting `non_blocking=True/False` doesn't affect any of the test results since it is purely a data-transfer fine-tuning operation. If so, we might as well **not** change the defaults in the test.

AFAIK testing this would require some kind of profiling check.

Luthaf:
You are right that testing this properly would be very hard, and I think no doable in CI. 

The current tests changes are mainly there as a smoke test, ensuring the parameter exists and the code does not crash when these are used.

HaoZeke:
> You are right that testing this properly would be very hard, and I think no doable in CI.
> 
> The current tests changes are mainly there as a smoke test, ensuring the parameter exists and the code does not crash when these are used.

Makes sense, perhaps a comment clarifying this would be good, to prevent a false sense of security down the line :D

python/metatensor_torch/tests/block.py
```diff
@@ -281,7 +281,7 @@ def test_to():
     check_dtype(block, torch.float32)
     check_dtype(block.gradient("g"), torch.float32)
 
-    converted = block.to(dtype=torch.float64)
+    converted = block.to(dtype=torch.float64, non_blocking=True)
```
HaoZeke, line 284: Maybe something as simple as

```suggestion
    # smoke test for non_blocking, see gh-920
    converted = block.to(dtype=torch.float64, non_blocking=True)
```

Luthaf: this test is also properly checking the rest of `to`. But I'll find a way to indicate this!

Luthaf: Better now?

HaoZeke: Yup, excellent.

HaoZeke (approved):
LGTM, thanks @Luthaf

Issue #775, opened 2024-10-28
Allow serialization when the data is something else than `float64`


Luthaf:
Currently, serialization is done inside `metatensor-core`, which does not know anything about the actual data, but only has access to the methods in [`mts_array_t`](https://github.com/metatensor/metatensor/blob/d92dffebb7bd2a28e444a5b9b681c7c11dc1d449/metatensor-core/src/data.rs#L72-L166). The relevant one here is `data`, which only allows taking out `f64` data.

It would be nice not to have to convert to float64 just for data serialization, and lift this limitation. This means adding a function in `mts_array_t` to access other data types. Instead of inventing our own API here, we should reuse dlpack and implement #56.

Luthaf:
Done in #1025

Issue #568, opened 2024-04-03
Add `LabelsDict` helper class


Luthaf:
There are a lot of time where we would like to pass some data for each block/each sample/each property. For now we pass this as a list of the right size, but it sometimes seems that a dict would be nicer. Unfortunately, `LabelsEntry` is not hashable and can not be made so in TorchScript. 

What we could do instead is provide a specialization of `Dict[Labels, T]` as a separate class, that would store all values in a list and use `Labels.position` for O(1) lookup of keys:

```py
class LabelsDict:
    def __init__(self, ...):
        self._labels = Labels(...)
        self._values = [...]

    def get(self, key, default=None):
        # O(1) lookup
        index = self._labels.position(key)
        if index is None:
            return default
        else:
            return self._values[index]

    # all the other dict functions
```
    
Here is how it could be used

```py
keys = Labels(...)
# two ways of constructing it:
a = LabelsDict(keys, [1, 2, 3, 4])

b = LabelsDict(
    ["a", "b"],
    {
        (0, 0): 1,
        (0, 1): 2,
        (0, 2): 3,
        (1, 0): 4,
    },
)

# Supports the usual dict API:
a[keys[0]]
a[keys[0]] = 5
a.get(keys[0], 42)

for k, v in a.items():
    pass
```

Luthaf:
Having this class would impact `metatensor-learn` API quite a bit, so let's discuss this idea during the next meeting!

PR #921, opened 2025-05-08
Provide a custom `nn.Module` class for better integration with metatensor data

### LINKED ISSUE #875: Compatibility of TensorMaps and TensorBlocks with buffers in PyTorch model?
Greetings  to my fellow metaverse peeps 👋 

There are several cases where it would be nice for the user to be able to save TensorMaps and TensorBlocks as buffers in PyTorch Models, which are then communicated through the state dicts of the models during checkpoint routines. Example within the context of MLIPs would be composition weights (baseline energies), which was something that I ran into recently with @frostedoyster.

AFAIK this is not yet possible, and hence one needs to come up with rather clunky workarounds to avoid such an issue: https://github.com/metatensor/metatrain/pull/522

It'd be nice if this issue could be discussed and resolved, thanks!

Luthaf:
This adds a `mts.learn.nn.Module` class that can be used instead of `torch.nn.Module` and provides integration between torch and metatensor data in two ways:

- calling `module.to()` will also move Labels, TensorBlock, TensorMap, as well as dict/list/tuple containing them to the correct dtype/device
    - in C++ mode we need a separate class to implement `to` since `torch::jit::Module` and `torch::nn::Module` are completely separate from `torch.nn.Module`
    - same when running TorchScript code from Python, we have to monkeypatch `torch.jit.RecursiveScriptModule`
- `state_dict()` and `load_state_dict` will properly save and load metatensor data
    - this is also true for float32 data/on device data. For now the code will convert to float64 and back below the hood, until #775 is fixed

TODO: 
- [x] document the format used in state_dict
- [x] rebase once #920 is merged

Fix #875 

# Reviewer checklist


Luthaf:
Still TBD: using this inside the different modules. One issue will be that until we update `load_atomistic_model` to use the custom C++ Module, we'll need to keep the code that moves data to dtype/device in `forward`.

metatensor-torch/tests/test-module.pt
jwa7, line 1: This looks like it's created by the `create-test-module.py` - should it still be here?

python/metatensor_learn/metatensor/learn/nn/_module.py
```diff
@@ -0,0 +1,352 @@
...
+            some_changed = some_changed or changed
+            updated.append(updated_value)
+
+        if some_changed:
+            return tuple(updated), True
+
+    return value, False
+
+
+def _serialize_metatensor(value):
```
jwa7, line 122: How about `LabelsEntry`?

jwa7 (commented):
Looks good on a first pass, but a couple of comments:

metatensor-torch/tests/test-module.pt
Luthaf: The issue is that we need an installation of metatensor-torch to create this file, and it is used to then test metatensor-torch itself … I'd like to dynamically generate this file, but doing so might make running the tests a lot harder. Hence why this file is committed to this repository.

python/metatensor_learn/metatensor/learn/nn/_module.py
Luthaf: I'd leave it out for now, unless there is a pressing need for it. I'd rather implement #568 as a solution here.

Adding it would first require to define the serialization function for LabelsEntry in C++.

metatensor-torch/src/module.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+
+using namespace metatensor_torch;
+
+template <typename T>
+bool is_custom_class(torch::IValue ivalue) {
+    if (ivalue.isCustomClass()) {
+        // this is inspired by the code inside `torch::IValue.toCustomClass<T>()`
+        auto* expected_type = torch::getCustomClassType<torch::intrusive_ptr<T>>().get();
+        return ivalue.type().get() == expected_type;
```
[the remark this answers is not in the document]
HaoZeke: Also perhaps better to use `c10::` over `torch::` to be more self documenting (veryminor nit)

metatensor-torch/src/module.cpp
```diff
@@ -0,0 +1,131 @@
...
+                    "as values are not supported"
+                );
+            }
+            return std::make_pair(updated, true);
+        }
+    } else if (ivalue.isList()) {
+        const auto& list = ivalue.toList();
+        auto updated = c10::impl::GenericList(list.elementType());
+        auto all_changed = true;
+        auto some_changed = false;
```
HaoZeke, line 70: Minor nit again, I'd prefer these to be set to falsy and updated to truthy in the loop, i.e. `false and true` instead of true and false.

PR #924, opened 2025-05-12
Handle empty vector before dispatchiong to concatenate when inserting a Label dimension

### LINKED ISSUE #600: `mts.torch.insert_dimension` error for Labels of zero length
If a `Labels` object is of zero length, one can insert a dimension in the normal way, passing a tensor of values of zero length:

```py
import torch

samples = mts.Labels(names=["a", "b"], values=torch.tensor([[0, 1]])[:0])  # [:0] to make it empty
print(samples)
samples = samples.insert(index=2, name="c", values=torch.tensor([3] * len(samples), dtype=torch.int32))
print(samples)
```
outputs:
```
Labels(
    a  b
)
Labels(
    a  b  c
)
```

However, when doing so on a TensorMap, one passes an `int` to the arg `values` such that Labels columns for each block are dynamically set to the correct dimension. In torch mode this raises an 
```py
import torch

samples = mts.Labels(names=["a", "b"], values=torch.tensor([[0, 1]])[:0])  # [:0] to make it empty
tensor = mts.TensorMap(
    keys=mts.Labels(names=["key"], values=torch.tensor([[0]])),
    blocks=[
        mts.TensorBlock(
            samples=samples,
            components=[],
            properties=mts.Labels.single(),
            values=torch.zeros((len(samples), 1)),
        )
    ],
)
mts.insert_dimension(tensor, axis="samples", name="d", index=3, values=4)
```
raises:
```
  File "/opt/miniforge3/envs/rho/lib/python3.11/site-packages/metatensor/operations/_dispatch.py", line 198, in concatenate
    if isinstance(arrays[0], TorchTensor):
        _check_all_torch_tensor(arrays)
        return torch.cat(arrays, axis)
               ~~~~~~~~~ <--- HERE
    elif isinstance(arrays[0], np.ndarray):
        _check_all_np_ndarray(arrays)
RuntimeError: torch.cat(): expected a non-empty list of Tensors
```
I think the fix requires the logic to be changed just before dispatching, i.e. to only call `_dispatch.concatenate` on non-zero length values arrays and manually handle zero-length ones. I have not yet looked deeply into this: just opening the issue for now, will come back to it later.

RMeli:
Close #600.

This PR adds a check that arrays are non-empty before dispatching to concatenate. Otherwise, returns an empty tensor with the same type as values.

# Reviewer checklist

### DIFF
--- a/python/metatensor_operations/metatensor/operations/_manipulate_dimension.py
+++ b/python/metatensor_operations/metatensor/operations/_manipulate_dimension.py
@@ -12,6 +12,13 @@ def _check_axis(axis: str):
         )
 
 
+def _concat(values: Array, n: int, axis: int):
+    print(">>> DBG", values, values.shape, n)
+    if n == 0:
+        return _dispatch.empty_like(values, shape=(0,))
+    return _dispatch.concatenate([values] * n, axis=axis)
+
+
 @torch_jit_script
 def insert_dimension(
     tensor: TensorMap,
@@ -78,7 +85,7 @@ def insert_dimension(
 
     if axis == "keys":
         if values_was_int:
-            label_values = _dispatch.concatenate([values] * len(keys), axis=0)
+            label_values = _concat(values, len(keys), axis=0)
 
         keys = keys.insert(index=index, name=name, values=label_values)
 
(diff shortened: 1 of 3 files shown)


Luthaf (approved):
Thanks!

PR #927, opened 2025-05-27
Add `drop_empty_blocks` operation

### LINKED ISSUE #786: Implement a `drop_empty_blocks` function in `metatensor-operations`
It could be useful to have a `drop_empty_blocks` operation, where blocks with any 0-length dimensions are dropped.

alessandroforina:
In this case, a method called drop_empty_blocks has been implemented, whose function is to drop all blocks contained in a TensorMap that have any of their dimensions equal to 0.

fix #786

# Reviewer checklist

### DIFF
--- a/docs/src/operations/reference/manipulation/drop-blocks.rst
+++ b/docs/src/operations/reference/manipulation/drop-blocks.rst
@@ -2,3 +2,5 @@ drop_blocks
 ===========
 
 .. autofunction:: metatensor.drop_blocks
+
+.. autofunction:: metatensor.drop_empty_blocks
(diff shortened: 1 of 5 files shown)


jwa7 (dismissed):
Hey @alessandroforina - thanks for the PR! It's looking good so far, but we'll have to also make this available in `metatensor.torch.operations` and add a test to make sure it's torchscript-able.

You can take a look at the source code for the metatensor-torch version of `drop_blocks` (https://github.com/metatensor/metatensor/blob/main/python/metatensor_torch/tests/operations/drop_blocks.py) for inspiration. Essentially what all you need to do is create another test function in that file for `drop_empty_blocks`. This will be very similar to the `test_drop_blocks` function that is already there.

Luthaf (approved):

PR #936, opened 2025-07-04
Patch RecursiveScriptModule even if `metatensor-learn` is not available

Luthaf:
Follow up to #921 

Someone could be using a model created with metatensor-learn and exported to TorchScript, and not have metatensor-learn installed when running the model. In this case we still want to properly handle metatensor data, so we need the code for this to always be executed when importing metatensor-torch.

# Reviewer checklist


python/metatensor_learn/metatensor/learn/nn/_module.py
```diff
@@ -85,8 +85,8 @@ def _apply(self, fn, recurse=True):
         return result
 
 
-# WARNING: used by metatensor.torch.learn, do not update without increasing the version
-# number as for a breaking change.
+# WARNING: this is duplicated in metatensor.torch._module, make sure to change both
+# versions of the function at the same time
```
HaoZeke, line 89: Does it make sense to have an API version string or something to check that it is the same as in the duplicated code? (same comment for the other duplicates)

python/metatensor_torch/metatensor/torch/_module.py
```diff
@@ -0,0 +1,194 @@
...
+        qualified_name = value._type().qualified_name()
+        if qualified_name.startswith("__torch__.torch.classes.metatensor"):
+            return value._type().name() == typename
+
+    return False
+
+
+# WARNING: this is duplicated in metatensor.learn.nn._module, make sure to change both
+# versions of the function at the same time
+def _metatensor_data_to(value, dtype, device):
```
HaoZeke, line 32: Same nit as earlier.

HaoZeke (commented):
Minor procedural nit..

PicoCentauri (commented):
I don't know but couldn't we just code generate 😅 it at installation to avoid explicit duplication? Feel free to just say no as code generation is dangerous territory.

HaoZeke:
> I don't know but couldn't we just code generate 😅 it at installation to avoid explicit duplication? Feel free to just say no as code generation is dangerous territory.

Not a bad plan but it'd make editable installs really tough IMO.

python/metatensor_learn/metatensor/learn/nn/_module.py
HaoZeke: but this could be for a runtime check across versions right? So the user wouldn't have to worry about differing copies?

HaoZeke:
I wonder if we later want to have some kind of documentation shortcut role like `devnote` or something to make sure these are not missed.

Luthaf:
You mean in the code itself? I tend to use comments for stuff that devs should read, and docstring for user-facing informations

Issue #368, opened 2023-09-12
Adding code coverage


agoscinski:
Coverage is helpful to understand what part of the codes are not tested yet, and to give a rough measure on stability of the code. We can more or less copy the way it has been done in rascaline. I will try to do the Python coverage, someone else needs to do the coverage for the C-API

- [ ] Python coverage that can simply added with pytest-cov
- [ ] C-API
- [ ] 
Reference how it was done in rascaline
https://github.com/Luthaf/rascaline/pull/126
https://github.com/Luthaf/rascaline/pull/200

Luthaf:
Updating the issue: we now have code coverage for Python, but are still missing it for the Rust and C++ code.

Luthaf:
Done in #939 and #1088

PR #939, opened 2025-07-10
[ENH] Add rust coverage

HaoZeke:
Partially implements #368. Didn't use the older Tarpaulin approach from `rascaline` because this instrumenting the source instead of sampling is better supported and more stable. Plus [cargo-llvm-cov](https://github.com/taiki-e/cargo-llvm-cov) also supports regions in codecov.

Locally I am testing this via:

```bash
rustup component add llvm-tools 
cargo +stable install cargo-llvm-cov --locked
cargo llvm-cov --all-features --workspace --no-fail-fast --codecov --output-path rust_codecov.json --exclude metatensor-python
# or to view the HTML just --html instead of --codecov
```
The more traditional route also works:

```bash
CARGO_INCREMENTAL=0 RUSTFLAGS='-Cinstrument-coverage' LLVM_PROFILE_FILE='cargo-test-%p-%m.profraw' cargo test
grcov . --binary-path ./target/debug/deps/ -s . -t html --branch --ignore-not-existing --ignore '../*' --ignore "/*" -o target/coverage/html
```

In any case to extract information from the `lcov` there is `genhtml`:

```bash
genhtml lcov.info --output-directory out
```

Though the `grcov` HTML is often nicer. FWIW it does seem like a pretty useful thing to have.. since there's not all that much coverage at the moment.

image

One thing to consider is converting the `coverage` files and merging them if we can try to make sure that any of the bindings cover the functions used instead of the current per-binding approach.

TODO:
- [x] Might want branch coverage (discuss)
- [x] Maybe `--include-ffi`
  - Though that should be in the follow up instrumenting the C/C++ stuff 

 ~- [ ] Tests updated (for new features and bugfixes)?~ N/A.

# Reviewer checklist

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -2,7 +2,7 @@ coverage:
   status:
     project:
       default:
-        target: 80%
+        target: 70%
     patch: off
   ignore:
     - "tests/.*"
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -2,7 +2,7 @@ coverage:
   status:
     project:
       default:
-        target: 70%
+        target: 78%
     patch: off
   ignore:
     - "tests/.*"
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly
+          toolchain: stable
           components: llvm-tools-preview
           override: true
 
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -73,6 +73,9 @@ jobs:
             --no-fail-fast --codecov \
             --exclude metatensor-python \
             --output-path rust_codecov.json
+        env:
+            # Use the CPU only version of torch when building/running the code
+            PIP_EXTRA_INDEX_URL: https://download.pytorch.org/whl/cpu
 
       - name: upload to codecov.io
         uses: codecov/codecov-action@v5
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -14,8 +14,26 @@ jobs:
   coverage:
     runs-on: ubuntu-22.04
     name: collect code coverage
+    # Kanged from [1]
+    strategy:
+      matrix:
+        llvm-version: [20]
+    env:
+      CMAKE_C_COMPILER: clang-${{ matrix.llvm-version }}
+      CMAKE_CXX_COMPILER: clang++-${{ matrix.llvm-version }}
+      CC: clang-${{ matrix.llvm-version }}
+      CXX: clang++-${{ matrix.llvm-version }}
+
     steps:
       - uses: actions/checkout@v4
+
+      - name: setup LLVM
+        run: |
+            wget https://apt.llvm.org/llvm.sh
+            chmod +x llvm.sh
+            sudo ./llvm.sh ${{ matrix.llvm-version }}
+            sudo apt-get update
+            sudo apt install --no-install-recommends clang-${{ matrix.llvm-version }}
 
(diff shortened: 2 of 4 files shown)


Luthaf:
So one advantage of tarpaulin is that we could use it to see how code in the C and C++ tests was exercising the C API defined in Rust. Is that what `--include-ffi` would enable here as well?

.github/workflows/coverage.yml
```diff
@@ -22,6 +22,16 @@ jobs:
         with:
           python-version: "3.13"
 
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
```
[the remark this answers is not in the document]
HaoZeke: `codecov` has some nicer branch coverage with nightly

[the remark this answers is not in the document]
HaoZeke: Specifically this is just for branch coverage in general https://github.com/taiki-e/cargo-llvm-cov/issues/8

HaoZeke:
Looking at https://app.codecov.io/gh/metatensor/metatensor/pull/939/indirect-changes

The coverage looks reasonable to me, I'll open a follow up to run the C/Cpp with `clang` to enable `--include-ffi`.

HaoZeke:
Now with the instructions in the added documentation local view it is much closer to the CodeCov output and includes `--include-ffi`

image
image


Still not exactly the same though weirdly.

.codecov.yml
```diff
@@ -2,7 +2,7 @@ coverage:
   status:
     project:
       default:
-        target: 80%
+        target: 70%
```
Luthaf, line 5: Let's keep this as high as possible

```suggestion
        target: 78%
```

.github/workflows/coverage.yml
```diff
@@ -56,9 +66,20 @@ jobs:
             coverage combine .tox/*/.coverage
             coverage xml
 
+      - name: Generate code coverage
+        id: coverage
+        run: |
+          cargo llvm-cov --all-features --workspace \
+            --no-fail-fast --codecov \
+            --exclude metatensor-python \
```
Luthaf, line 74: Do you also want to add `--include-ffi` here? And why is `metatensor-python` excluded here and not in the instructions?

Luthaf:
CI is failing, maybe because the job to collect coverage is using gcc and not clang as a C++ compiler?

.github/workflows/coverage.yml
```diff
@@ -14,14 +14,42 @@ jobs:
   coverage:
     runs-on: ubuntu-22.04
     name: collect code coverage
+    # Kanged from [1]
```
Luthaf, line 17:
```suggestion
    # Adapted from https://www.scivision.dev/github-actions-latest-llvm-clang-flang/
```

.github/workflows/coverage.yml
```diff
@@ -56,9 +84,23 @@ jobs:
             coverage combine .tox/*/.coverage
             coverage xml
 
+      - name: generate code coverage
```
Luthaf, line 87:
```suggestion
      - name: generate rust code coverage
```

.github/workflows/coverage.yml
```diff
@@ -56,9 +84,23 @@ jobs:
...
       - name: upload to codecov.io
         uses: codecov/codecov-action@v5
         with:
           fail_ci_if_error: true
-          files: coverage.xml
+          files: coverage.xml,rust_codecov.json
           token: ${{ secrets.CODECOV_TOKEN }}
+
+# References
+# [1]: https://www.scivision.dev/github-actions-latest-llvm-clang-flang/
```
Luthaf, line 106:
```suggestion
```

Luthaf (commented):
Looks good and the report on codecov is very reasonable

PR #948, opened 2025-07-28
Added the possibility to not check gradients on `equal_metadata`

pfebrer:
In https://github.com/metatensor/metatrain/pull/693 we need a function that checks if all the labels are the same but doesn't check the gradients.

In this PR I added the possibility to `metatensor.equal_metadata` the possibility to avoid checking the gradients so that we can use the function for that purpose.

Also added an extra test to check that `check_gradients=False`. 

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/3630008078.zip)

### DIFF
--- a/python/metatensor_operations/tests/equal_metadata.py
+++ b/python/metatensor_operations/tests/equal_metadata.py
@@ -466,3 +466,10 @@ def test_remove_last_component(tensor_map):
 
     new_tensor = TensorMap(keys=tensor_map.keys, blocks=new_blocks)
     assert not mts.equal_metadata(tensor_map, new_tensor)
+
+
+def test_ignore_gradients(tensor_map):
+    """Check that we can compare tensormaps while ignoring gradients."""
+    new_tensor = mts.remove_gradients(tensor_map)
+    assert not mts.equal_metadata(tensor_map, new_tensor, check_gradients=True)
+    assert mts.equal_metadata(tensor_map, new_tensor, check_gradients=False)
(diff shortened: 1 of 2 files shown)


python/metatensor_operations/tests/equal_metadata.py
```diff
@@ -466,3 +466,21 @@ def test_remove_last_component(tensor_map):
...
+        block = tensor_map[key].copy()
+        new_block = TensorBlock(
+            values=block.values,
+            samples=block.samples,
+            properties=block.properties,
+            components=block.components,
+        )
+        new_blocks.append(new_block)
+
+    new_tensor = TensorMap(keys=tensor_map.keys, blocks=new_blocks)
```
Luthaf, line 484:
```suggestion
    new_tensor = mts.remove_gradients(tensor_map)
```

Luthaf (commented):
Why is this flag needed compared to using `mts.remove_gradients` before checking? I.e. why do you want to check only the values and not the gradients?

python/metatensor_operations/tests/equal_metadata.py
pfebrer: This is simply because I didn't know that `remove_gradients` existed. In this case I just let Copilot do the job :) And apparently it does not know too much metatensor yet haha

pfebrer:
This is in principle ready

Luthaf (approved):

PR #974, opened 2025-09-08
Rewrite `join` to no longer use `keys_to_xxx`

Luthaf:
We previously had to add an extra "tensor" dimension in the keys, and the check to know whether we could or not remove this extra dimension was accidentally quadratic, taking a lot of time when using join to batch data when training ML models.

This is a breaking change since the API for `mts.join` will change. Instead of always adding a dimension, the user can now pick if they want to add a dimension tracking the origin of joined samples/properties with `add_dimension`. The `sort_samples` option was also removed.

Some benchmarks, batching the MAD training set. The two options with v0.3.4 where to either set `remove_tensor_name=True`, or to set it to `False` and manually call `mts.remove_dimension` after the fact.

<details>

```py
import time

import ase.io
import torch

def frame_to_tensor(frame, index):
    e = torch.tensor(frame.get_potential_energy())
    f = torch.tensor(frame.get_forces())

    block = mts.TensorBlock(
        values=e.reshape(1, 1),
        samples=mts.Labels("system", torch.tensor([[index]])),
        components=[],
        properties=mts.Labels("energy", torch.tensor([[0]])),
    )

    grad_samples = torch.zeros((len(f), 3), dtype=torch.int32)
    grad_samples[:, 0] = 0
    grad_samples[:, 1] = index
    grad_samples[:, 2] = torch.tensor(list(range(len(f))))
    gradient = mts.TensorBlock(
        values=-f.reshape(-1, 3, 1),
        samples=mts.Labels(["sample", "system", "atom"], grad_samples),
        components=[mts.Labels("xyz", torch.tensor([[0], [1], [2]]))],
        properties=mts.Labels("energy", torch.tensor([[0]])),
    )

    block.add_gradient("positions", gradient)
    return mts.TensorMap(mts.Labels("_", torch.tensor([[0]])), [block])


frames = ase.io.read("mad-train.xyz", ":")
print("done loading")

tensors = []
for i, frame in enumerate(frames):
    tensors.append(frame_to_tensor(frame, i))
print("done converting")


def batch_join(tensors, batch_size):
    start = 0
    while start < len(tensors):
        # # old
        # _ = mts.join(
        #     tensors[start : start + batch_size],
        #     axis="samples",
        #     remove_tensor_name=True,
        # )

        # # old manual remove
        # tensor = mts.join(tensors[start : start + batch_size], axis="samples")
        # _ = mts.remove_dimension(tensor, axis="samples", name="tensor")

        # new
        _ = mts.join(tensors[start : start + batch_size], axis="samples")
        start += batch_size


for batch_size in [8, 16, 32, 64, 128]:
    print(f"============ BATCH SIZE {batch_size}")

    # warmup
    start = time.time()
    _ = batch_join(tensors[: 3 * batch_size], batch_size=batch_size)
    stop = time.time()

    n_iters = 2
    n_batches = len(frames) // batch_size
    print(f"estimated benchmark time: {(stop - start) / 3 * n_batches * n_iters} s")

    start = time.time()
    for _ in range(n_iters):
        _ = batch_join(tensors, batch_size=batch_size)
    stop = time.time()

    print(f"{(stop - start) / n_iters * 1e3} ms")
    print(f"{(stop - start) / (n_batches * n_iters) * 1e3} ms/batch")

```

</details>

timing
timing-per-batch

# Reviewer checklist

### DIFF
--- a/docs/src/operations/reference/manipulation/join.rst
+++ b/docs/src/operations/reference/manipulation/join.rst
@@ -2,3 +2,5 @@ join
 ====
 
 .. autofunction:: metatensor.join
+
+.. autofunction:: metatensor.join_blocks
--- a/python/metatensor_learn/tests/collate.py
+++ b/python/metatensor_learn/tests/collate.py
@@ -122,7 +122,7 @@ def test_group_and_join_tensormaps():
         blocks=[
             TensorBlock(
                 values=np.ones((2, 1)),
-                samples=Labels(["sample_index", "tensor"], np.array([[0, 0], [2, 1]])),
+                samples=Labels(["sample_index"], np.array([[0], [2]])),
                 components=[],
                 properties=Labels(["p"], np.array([[0]])),
             )
@@ -194,9 +194,7 @@ def test_group_and_join_torch_tensormaps():
         blocks=[
             TensorBlock(
                 values=torch.ones((2, 1)),
-                samples=Labels(
-                    ["sample_index", "tensor"], torch.tensor([[0, 0], [2, 1]])
-                ),
+                samples=Labels(["sample_index"], torch.tensor([[0], [2]])),
                 components=[],
                 properties=Labels(["p"], torch.tensor([[0]])),
             )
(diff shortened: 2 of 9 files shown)


Luthaf:
> A first superficial reaction, add_dimension is not very telling, I'd use something like add_tensor_index or tensor_index_label.

So the point for `add_dimension` was that we have the same argument in `join_blocks`, so anything with `tensor` in the name does not work IMO. The added name is also user-defined: `add_dimension="foo"`, `add_dimension="this_is_the_old_tensor"`.

We could do `joined_index_name: Optional[str]` or `add_joined_index: Optional[str]` if you think it's clearer!

jwa7 (changes requested):
This is great thank you @Luthaf 🚀 Just some docs changes requested. Personally I think the `add_dimension` is fine as is - it does what it says on the tin

PR #965, opened 2025-08-25
[ENH] Use `dlpack`

HaoZeke:
Post #964. Superseeds #933. Closes #56. Closes #775. Closes #870.

# Contributor (creator of pull-request) checklist

 - [ ] Tests updated (for new features and bugfixes)?
 - [ ] Documentation updated (for new features)?
 - [ ] Issue referenced (for PRs that solve an issue)?

# Reviewer checklist

 - [ ] CHANGELOG updated with public API or any other important changes?

PR #984, opened 2025-10-01
Split metatensor's C++ header into multiple files

### LINKED ISSUE #982: [MAINT] breakup `metatensor.hpp`
Despite `clangd` it is a major headache to switch around a ~3000 LOC CPP file. I'd recommend either breaking into `.hpp` files or even `.inl` (if we want templates that's clearer).

HaoZeke:
Closes #982. Needed for more progress on #965.

This PR refactors the monolithic C++ header (`metatensor.hpp`) into a modular, header-only library structure to significantly improve the developer experience with minimal source changes.

# Reviewer checklist


Luthaf:
This is not vesin, so we can assume that the user is installing the code through CMake. We don't have to keep support for a single header either because end users will have to install the library as well, so they can not use the "copy header inside my project" strategy =)

HaoZeke:
> You should try to tone down your LLM, it uses hyperbolic language ("separate, complex entity", "cumbersome") for no reason …
> 
> > Plus providing a single-header distribution from modular sources is pretty popular in the C++ space, like xtb/tblite, catch2, nlohmann/json;
> 
> catch2 and json are both pure C++ library that can be used as header only, this does not apply for us. Not sure about xtb.
> 
> > Lowest Friction for All Build Systems:
> 
> Not applicable, people have to use CMake to build this library.
> 
> > Mostly because I didn't want to touch the CMake build to handle installing a bunch of new files
> 
> I'm happy to do the cmake changes =)

Fair enough, if the initial breakdown looks OK, could you push the CMake changes?

metatensor-core/CMakeLists.txt
```diff
@@ -408,7 +413,14 @@ write_basic_package_version_file(
     COMPATIBILITY SameMinorVersion
 )
 
-install(FILES ${METATENSOR_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+install(FILES ${METATENSOR_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/metatensor)
```
Luthaf, line 416: This is a breaking change, it would require `#include <metatensor/metatensor.hpp>` by users, which I don't think we want to do.

```suggestion
install(FILES ${METATENSOR_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
```

metatensor-core/CMakeLists.txt
```diff
@@ -408,7 +413,14 @@ write_basic_package_version_file(
...
+install(DIRECTORY ${PROJECT_SOURCE_DIR}/include/cxx
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/metatensor
```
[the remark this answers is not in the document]
HaoZeke: Yes but that'd less-cleanly separate the `C` and `C++` headers

metatensor-core/CMakeLists.txt
```diff
@@ -365,9 +365,13 @@ set(METATENSOR_HEADERS
...
 file(MAKE_DIRECTORY ${METATENSOR_INCLUDE_DIR})
 
+file(GLOB_RECURSE METATENSOR_CXX_HEADERS
+    "${PROJECT_SOURCE_DIR}/include/*.hpp"
+)
+
 set_target_properties(metatensor::shared PROPERTIES
     IMPORTED_LOCATION ${METATENSOR_SHARED_LOCATION}
-    INTERFACE_INCLUDE_DIRECTORIES ${METATENSOR_INCLUDE_DIR}
+    INTERFACE_INCLUDE_DIRECTORIES "${METATENSOR_INCLUDE_DIR};${PROJECT_SOURCE_DIR}/include"
```
Luthaf, line 374: I don't understand why this is needed?

PR #986, opened 2025-10-01
chore(std): use CPP17 throughout

### LINKED ISSUE #985: [MAINT] Bump min `cxx` standard
Currently we have `CPP17` on `metatensor-torch` and `CPP11` on `metatensor`. Could we shift to `CPP17` throughout? OTOH, I'd like:

### `std::optional` in `Labels`

The current `Labels::position` function returns `-1` when it finds no entry. "magic numbers" are fun, but potentially buggy. Someone could forget to check for it and use the result as a valid index.

```cpp
// C++11 Style: Returns -1 on failure
int64_t position(const int32_t* entry, size_t length) const;
```

C++17's `std::optional` completely removes this ambiguity. The function signature now explicitly communicates that it might not return a value. This forces callers to handle the "not found" case correctly, making the code fundamentally safer.

```cpp
// C++17 Style: Returns std::nullopt on failure
std::optional<int64_t> position(const int32_t* entry, size_t length) const;
```

### Correctness at Compile Time with `if constexpr`

The `NDArray::operator==` currently uses `memcmp` for all data types. This function behaves incorrectly for floating-point types due to, [well floating points being floating points](https://people.cs.pitt.edu/~cho/cs1541/current/handouts/goldberg.pdf), with `NaN` values and `-0.0` vs `0.0` and what not.

```cpp
// C++11 Style: Risky for floats
return std::memcmp(lhs.data(), rhs.data(), /*...*/) == 0;
```

C++17's `if constexpr` would enable type-specific logic that the compiler selects with zero runtime overhead. So a fast `memcmp` for integers while implementing a correct, element-wise comparison for floats. This fixes a potential bug and improves correctness.

```cpp
// C++17 Style: Safe and efficient for all types
if constexpr (std::is_floating_point_v<T>) {
    // ... safe element-wise comparison ...
} else {
    // ... fast memcmp for integers ...
}
```

### Better Performance with `std::string_view`

Our functions that take `const std::string&` arguments force a memory allocation when a user passes a string literal (e.g., `"positions"`). This creates unnecessary overhead.

```cpp
// C++11 Style: Allocates for string literals
void add_gradient(const std::string& parameter, ...);
```

Changing the signature to `std::string_view` eliminates this allocation. A `string_view`, being a view of any C-style string or `std::string`, would make the API both faster and more flexible.

```cpp
// C++17 Style: Zero-allocation for string literals
void add_gradient(std::string_view parameter, ...);
```

Given that all modern toolchains support C++17 and `metatensor-torch` already uses it, aligning the core library makes sense for consistency and future development (plus I get `nestd::namespaces` for #984).

HaoZeke:
Closes #985. More reasons given in the issue.

 - [X] Tests updated (for new features and bugfixes)?
 - [X] Documentation updated (for new features)?

# Reviewer checklist

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -29,8 +29,7 @@ Additionally, you will need to install the following software, but you should
 not have to interact with them directly:
 
 - **CMake**: we need a CMake version of at least 3.16.
-- **a C++ compiler** we need a compiler supporting C++11 (C++17 for
-  ``metatensor-torch``). GCC >= 7, clang >= 5 and MSVC >= 19 should all work.
+- **a C++ compiler** we need a compiler supporting C++17. GCC >= 7, clang >= 5 and MSVC >= 19 should all work.
 
 .. _rustup: https://rustup.rs
 .. _tox: https://tox.readthedocs.io/en/latest
--- a/metatensor-core/CMakeLists.txt
+++ b/metatensor-core/CMakeLists.txt
@@ -370,7 +370,7 @@ set_target_properties(metatensor::shared PROPERTIES
     INTERFACE_INCLUDE_DIRECTORIES ${METATENSOR_INCLUDE_DIR}
     BUILD_VERSION "${METATENSOR_FULL_VERSION}"
 )
-target_compile_features(metatensor::shared INTERFACE cxx_std_11)
+target_compile_features(metatensor::shared INTERFACE cxx_std_17)
 
 if (WIN32)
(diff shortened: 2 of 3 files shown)


HaoZeke:
> Do you want to also do the changes from #985 in here?

No I think those should be after / on top of #984.

Luthaf:
Restarting CI since macOS runners are having a hard time today

Luthaf (approved):

PR #1001, opened 2025-10-29
chore(pyproject): fixup as per packaging guide

HaoZeke:
i.e. https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license

xref: https://github.com/python-attrs/attrs/issues/1376

This is needed to pass CI on https://github.com/metatensor/metatensor/pull/997

### DIFF
--- a/python/metatensor_core/pyproject.toml
+++ b/python/metatensor_core/pyproject.toml
@@ -4,7 +4,7 @@ dynamic = ["version", "authors"]
 requires-python = ">=3.10"
 
 readme = "README.rst"
-license = {file = "LICENSE"}
+license = "BSD-3-Clause"
 description = "Python bindings for metatensor"
 
 dependencies = ["numpy"]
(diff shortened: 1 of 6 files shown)


python/metatensor_core/pyproject.toml
```diff
@@ -4,7 +4,7 @@ dynamic = ["version", "authors"]
 requires-python = ">=3.10"
 
 readme = "README.rst"
-license = "BSD-3-Clause"
+license = {file = "LICENSE"}
```
Luthaf: This seems wrong. The packaging guide says

> The new format for license is a valid [SPDX license expression](https://packaging.python.org/en/latest/glossary/#term-License-Expression) consisting of one or more [license identifiers](https://packaging.python.org/en/latest/glossary/#term-License-Identifier). The full license list is available at the [SPDX license list page](https://spdx.org/licenses/). The supported list version is 3.17 or any later compatible one.

```toml
[project]
license = "GPL-3.0-or-later"
# or
license = "MIT AND (Apache-2.0 OR BSD-2-Clause)"
```

Is the CI failure in the other PR because we have a too old version of setuptools? (that would have been caught by `--check-build-dependencies`)?

Luthaf:
I think this is really an issue with `uv` missing some kind of `--check-build-dependencies`. For now we can maybe update the build dependencies manually inside the virtualenv (like we do update `pip` which I think might also pull a newer setuptools)

PR #1010, opened 2025-11-12
Allow building multiple projects in parallel

Luthaf:
uv will try to install all the projects at the same time, but `git update-index` will lock the git index, and fail when called in parallel.

So we modify the code to write updates to a separate index file and atomically replace it once updated

# Reviewer checklist

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -69,7 +69,7 @@ deps =
     {[testenv]packaging_deps}
 
 commands =
-    uv build --wheel python/metatensor_core --no-build-isolation --out-dir {envtmpdir}/dist
+    uv build --wheel python/metatensor_core --no-build-isolation --out-dir {envtmpdir}/dist --quiet
 
 
 [testenv:core-tests]


scripts/git-version-info.py
```diff
@@ -72,9 +75,68 @@ def n_commits_since_last_tag(tag_prefix):
...
 
 
+def acquire_git_index_lock(lock: Path, timeout=2.0, poll=1e-3) -> bool:
+    start = time.time()
+    while True:
+        try:
+            fd = os.open(str(lock), os.O_CREAT | os.O_EXCL | os.O_WRONLY)
+            with os.fdopen(fd, "w") as f:
+                f.write(str(os.getpid()))
+            return True
```
HaoZeke:
```suggestion
    while True:
        try:
            with lock.open(mode="xw") as f:
                f.write(str(os.getpid()))
            return True
```

HaoZeke (changes requested):
Almost there, looks great. Generally would like less `os` and more `pathlib` :D

xref: https://github.com/astral-sh/uv/issues/16665

HaoZeke (approved):
LGTM, thanks @Luthaf

PR #1060, opened 2026-02-22
feat(mtsio): `mmap` loader

### LINKED ISSUE #972: Load serialized data with `mmap`
Sometimes (typically when loading only a subset of data, or when loading from specific distributed FS) it is faster to use mmap instead of open/read to load data from files.

Our serialization format should be able to support this by virtue of using `STORED` compression in the ZIP file, meaning we should be able to `mmap` the whole thing and then use offset inside the mmap range as data pointers for the different arrays.

This has the potential to make training ML models faster when all the target data are stored on disk, e.g. using metatrain with a DiskDataset.

Related to #946

HaoZeke:
Closes #972.

Add `load_mmap` / `load_block_mmap` across all layers for zero-copy loading of `.mts` files via memory-mapped I/O. Data arrays point directly into the mmap'd file (lazy page-in by the OS); labels are still loaded eagerly.

  | Layer | New API |
  |-------|---------|
  | C | `mts_tensormap_load_mmap`, `mts_block_load_mmap`, `mts_mmap_free` |
  | C++ | `metatensor::io::load_mmap`, `TensorMap::load_mmap`, etc. |
  | Rust | `metatensor::io::load_mmap`, `metatensor::io::load_block_mmap` |
  | Python (ctypes) | `metatensor.load_mmap`, `metatensor.load_block_mmap`, + static methods |
  | TorchScript | `metatensor.torch.load_mmap`, `metatensor.torch.load_block_mmap`, + `torch.ops.metatensor.*` |

# Notes

  - Uses a user-provided `mts_create_mmap_array_callback_t` so the core library stays array-agnostic—each binding constructs its own type (`MmapDataArray` in C++, `MmapNdarray` in Rust, `MmapOwner`-backed arrays in Python, `torch::from_blob` in PyTorch)
  - `mts_mmap_free` releases the underlying mapping
  - In TorchScript, `values()` has a DLPack fallback path that converts non-torch arrays (mmap-backed) into `torch::Tensor` via `DLManagedTensorVersioned`
  - Python ctypes layer uses `MmapCpuArray` which creates numpy views via `np.from_dlpack`

# Reviewer checklist


Luthaf:
Can you re-open targetting main?

PR #1061, opened 2026-02-22
feat(mtsio): selective `load_partial`

### LINKED ISSUE #946: Allow to load only a subset of a serialized TensorBlock/TensorMap
When training ML models where the data is stored inside a TensorMap/TensorBlock on disk, it would be very useful to have a single large file containing the data for a whole dataset, but only load parts of the file in memory at the time (for example one batch at the time). 

This could be done by passing "selection" Labels to the different `load` functions, and then only loading the corresponding data.

Unresolved questions:
- [ ] should this be done with a new function `mts_load_partial` or by adding a `mts_labels_t selection` parameter to `mts_load`?
- [ ] does the zip crate (which we use to load mts files) read the whole archive in memory and will need some adaptation, or cant it load individual files?
- [ ] Same question as above regarding the ability to load partial files from the archive and/or memory-map the files
- [ ] do we only want to do partial loading in the sample dimension, or should this also support partial keys and partial properties?

HaoZeke:
Closes #946. 

**Do not merge until ::**
- #1060 (mmap loader)

Add `load_partial` across all layers for selectively loading `.mts` files, filtering by keys, samples, and properties via `Labels::select` semantics. Uses memory mapping (`memmap2::Mmap`) for efficient random access—only selected data is copied into the output arrays. Returns standard callback-created arrays identical to `load()` in the same way as a load -> slice operation would as discussed with @Luthaf.

  | Layer | New API |
  |-------|---------|
  | C | `mts_tensormap_load_partial` |
  | C++ | `metatensor::io::load_partial`, `TensorMap::load_partial` |
  | Rust | `metatensor::io::load_partial` |
  | Python (ctypes) | `metatensor.load_partial` |
  | TorchScript | `metatensor.torch.load_partial`, `torch.ops.metatensor.load_partial` |
  | Julia | `mts_tensormap_load_partial` |

# Implementation

  - Memory-maps the file once, wraps the mapping in a `Cursor<&[u8]>` for the `zip` crate, and accesses data via pointer arithmetic + `copy_from_slice`
  - All three filter axes (keys, samples, properties) use `Labels::select` semantics—partial dimension match, missing entries silently ignored
  - After `create_array(shape)` allocates a zeroed array, mutable access is obtained via DLPack and selected rows/columns are written as bytes
  - Gradient samples are reindexed when parent samples are filtered (`HashMap<i32,i32>` old→new mapping)

# Mmap vs file seeking (benchmark results)

Both approaches were implemented and benchmarked. Mmap won decisively (2–3× faster across all filter configurations), so it was adopted as the sole implementation:

  | Filter | seek | mmap | mmap vs seek |
  |--------|------|------|-------------|
  | no filter | 2.28 ms | 718 µs | 3.2× faster |
  | key filter | 727 µs | 325 µs | 2.2× faster |
  | sample filter | 1.99 ms | 689 µs | 2.9× faster |
  | property filter | 2.29 ms | 762 µs | 3.0× faster |
  | combined | 660 µs | 297 µs | 2.2× faster |

Compared to `load()` (full load, 2.1 ms), `load_partial` with combined filtering is **7× faster**.

The seek variant required two file handles and repeated `seek+read` syscalls per row. The mmap variant maps the file once and lets the OS page in only the touched regions.

# Testing

9 Rust tests including element-level equivalence with `load()` + manual slice, gradient sample reindexing, and error handling.

# Reviewer checklist


Luthaf:
Can you re-open targetting main?

pfebrer:
+1, would be cool to have :)

PR #1086, opened 2026-04-08
Small fixes for v0.2.0

Luthaf:
The main one is about making fill_value an actual scalar array (since this is allowed by the DLPack API) instead of using an array with shape [1].

# Reviewer checklist

### DIFF
--- a/metatensor-core/include/metatensor/arrays.hpp
+++ b/metatensor-core/include/metatensor/arrays.hpp
@@ -159,6 +159,33 @@ namespace details {
         return false;
     }
 
+    template<typename Lhs, typename Rhs>
+    bool arrays_equal(const Lhs& lhs, const Rhs& rhs) {
+        if (lhs.shape() != rhs.shape()) {
+            return false;
+        }
+
+        using lhs_data_t = std::remove_cv_t<std::remove_pointer_t<decltype(lhs.data())>>;
+        using rhs_data_t = std::remove_cv_t<std::remove_pointer_t<decltype(rhs.data())>>;
+
+        static_assert(std::is_same_v<lhs_data_t, rhs_data_t>, "data types of lhs and rhs must be the same");
+
+        if (std::is_integral_v<lhs_data_t>) {
+            // compare integers with memcmp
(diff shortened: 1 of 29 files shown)


metatensor-core/include/metatensor/arrays.hpp
```diff
@@ -159,6 +159,33 @@ namespace details {
         return false;
     }
 
+    template<typename Lhs, typename Rhs>
+    bool arrays_equal(const Lhs& lhs, Rhs& rhs) {
```
HaoZeke:
```suggestion
    bool arrays_equal(const Lhs& lhs, const Rhs& rhs) {
```

Otherwise this wouldn't allow comparisons with `const` rvalues and stuff. Plus it's weird.

metatensor-core/include/metatensor/arrays.hpp
```diff
@@ -159,6 +159,33 @@ namespace details {
...
+            // make sure to handle NaN and ±0.0 correctly when comparing floating
+            // point data and everything else
+            const auto* lhs_ptr = lhs.data();
+            const auto* rhs_ptr = rhs.data();
+            for (size_t i=0; i<details::product(lhs.shape()); i++) {
+                if (lhs_ptr[i] != rhs_ptr[i]) {
+                    return false;
+                }
+            }
+            return true;
```
HaoZeke, line 186: Does everyone understand IEEE 754's `NaN == NaN` is correctly false? Do we want to document that or just leave it to the users to discover (not a new issue though)

HaoZeke (changes requested):
LGTM, one trivial to fix bug and a few clarifications

metatensor-core/include/metatensor/arrays.hpp
Luthaf: I think yes, this is not anything special here. Even numpy/torch do this.

HaoZeke (approved):

HaoZeke:
LGTM @Luthaf I'm assuming from the fixup commit there's to be some history cleanup? Merge at will afterwards. Thanks

Luthaf:
Yes, the fixup will be folded into the corresponding commit in the v0.2.0 branch.

PR #1094, opened 2026-04-16
Overhaul error handling

Luthaf:
We now have `mts_set_last_error` that can re-use the same thread local storage for error as libmetatensor. Both this function and `mts_last_error` allow saving arbitrary data together with the error, which is used to stored C++ or Python exceptions and retreive them later.

This was prompted by painful debugging of some errors where the error thrown by torch was lost to limbo and no longer accessible.

# Reviewer checklist

### DIFF
--- a/metatensor-core/CHANGELOG.md
+++ b/metatensor-core/CHANGELOG.md
@@ -39,6 +39,8 @@ a changelog](https://keepachangelog.com/en/1.1.0/) format. This project follows
   allowing to remove the need to copy all data back to CPU when creating labels.
   As part of this, `mts_labels_create` is now `mts_labels`, and
   `mts_labels_create_assume_unique` is now `mts_labels_assume_unique`.
+- `mts_last_error` can now return custom data and error origin together with the
+  last error message. These can be set by a new function `mts_set_last_error`.
 
 #### Added
 
@@ -46,6 +48,10 @@ a changelog](https://keepachangelog.com/en/1.1.0/) format. This project follows
   `mts_block_dtype`, `mts_tensormap_dtype` to access dtype and device of metatensor data.
 - There are new functions to work with `mts_labels_t`: `mts_labels_dimensions`,
   `mts_labels_values`, and `mts_labels_values_cpu`.
+- `mts_set_last_error` to set the last error that occured in a callback before
+  transferring control back to libmetatensor. The data set by this function can
+  then be retrieved with `mts_last_error` and used to store and rethrow
+  exceptions across the C API boundary.
 
 ### metatensor-core C++
 
--- a/metatensor-core/include/metatensor/errors.hpp
+++ b/metatensor-core/include/metatensor/errors.hpp
@@ -1,5 +1,8 @@
 #pragma once
 
+#include <cstring>
+#include <cstdio>
+
 #include <stdexcept>
 #include <string>
 #include <utility>
@@ -15,49 +18,26 @@ namespace metatensor {
     };
 
     namespace details {
-        /// Singleton class storing the last exception throw by a C++ callback.
-        ///
-        /// When passing callbacks from C++ to Rust, we need to convert exceptions
-        /// into status code (see the `catch` blocks in this file). This class
-        /// allows to save the message associated with an exception, and rethrow an
-        /// exception with the same message later (the actual exception type is lost
-        /// in the process).
-        class LastCxxError {
-        public:
-            /// Set the last error message to `message`
-            static void set_message(std::string message) {
-                auto& stored_message = LastCxxError::get();
-                stored_message = std::move(message);
-            }
-
-            /// Get the last error message
-            static const std::string& message() {
-                return LastCxxError::get();
-            }
-
-        private:
-            static std::string& get() {
-                #pragma clang diagnostic push
-                #pragma clang diagnostic ignored "-Wexit-time-destructors"
-                /// we are using a per-thread static value to store the last C++
-                /// exception.
-                static thread_local std::string STORED_MESSAGE;
-                #pragma clang diagnostic pop
-
-                return STORED_MESSAGE;
-            }
-        };
-
         /// Check if a return status from the C API indicates an error, and if it is
         /// the case, throw an exception of type `metatensor::Error` with the last
         /// error message from the library.
         inline void check_status(mts_status_t status) {
             if (status == MTS_SUCCESS) {
                 return;
-            } else if (status > 0) {
-                throw Error(mts_last_error());
-            } else { // status < 0
-                throw Error("error in C++ callback: " + LastCxxError::message());
+            } else if (status == MTS_CALLBACK_ERROR) {
+                const char* message = nullptr;
+                const char* origin = nullptr;
+                void* data = nullptr;
+                mts_last_error(&message, &origin, &data);
+                if (origin != nullptr &&std::strcmp(origin, "C++ exception") == 0 && data != nullptr) {
+                    std::rethrow_exception(*static_cast<std::exception_ptr*>(data));
+                } else {
+                    throw Error(message == nullptr ? "unknown error" : message);
+                }
+            } else {
+                const char* message = nullptr;
+                mts_last_error(&message, nullptr, nullptr);
+                throw Error(message == nullptr ? "unknown error" : message);
             }
         }
 
@@ -71,23 +51,57 @@ namespace metatensor {
             try {
                 function(std::move(args)...);
                 return MTS_SUCCESS;
-            } catch (const std::exception& e) {
-                details::LastCxxError::set_message(e.what());
-                return -1;
             } catch (...) {
-                details::LastCxxError::set_message("error was not an std::exception");
-                return -128;
+                auto* exception_ptr = new std::exception_ptr(std::current_exception());
+
+                const char* message = nullptr;
+                try {
+                    std::rethrow_exception(*exception_ptr);
+                } catch (const std::exception& e) {
+                    message = e.what();
+                } catch (...) {
+                    message = "C++ code threw an exception that was not an std::exception";
+                }
+
+                auto status = mts_set_last_error(
+                    message,
+                    "C++ exception",
+                    exception_ptr,
+                    [](void *ptr) { delete static_cast<std::exception_ptr*>(ptr); }
+                );
+
+                if (status != MTS_SUCCESS) {
+                    // If we failed to set the error, we are in a very bad state,
+                    // but we should still try to report the original error
+                    // message if possible.
+                    std::fprintf(stderr, "INTERNAL ERROR: unable to set last error after C++ callback failure (status: %d). ", status);
+                    if (message != nullptr) {
+                        fprintf(stderr, "C++ error was: %s\n", message);
+                    } else {
+                        fprintf(stderr, "Unknown C++ error\n");
+                    }
+                    delete exception_ptr;
+                }
+
+                return MTS_CALLBACK_ERROR;
             }
         }
 
         /// Check if a pointer allocated by the C API is null, and if it is the
-        /// case, throw an exception of type `metatensor::Error` with the last error
-        /// message from the library.
+        /// case, throw an exception of type `metatensor::Error` with the last
+        /// error message from the library.
         inline void check_pointer(const void* pointer) {
             if (pointer == nullptr) {
-                throw Error(mts_last_error());
+                const char* message = nullptr;
+                const char* origin = nullptr;
+                void* data = nullptr;
+                mts_last_error(&message, &origin, &data);
+                if (std::strcmp(origin, "C++ exception") == 0 && data != nullptr) {
+                    std::rethrow_exception(*static_cast<std::exception_ptr*>(data));
+                } else {
+                    throw Error(message);
+                }
             }
         }
-
     } // namespace details
 } // namespace metatensor
--- a/metatensor-core/src/c_api/status.rs
+++ b/metatensor-core/src/c_api/status.rs
@@ -1,69 +1,137 @@
 use std::panic::UnwindSafe;
 use std::cell::RefCell;
-use std::os::raw::c_char;
-use std::ffi::CString;
+use std::os::raw::{c_char, c_void};
+use std::ffi::{CString, CStr};
 
 use crate::Error;
+
+#[derive(Debug)]
+struct LastError {
+    message: CString,
+    origin: CString,
+    custom_data: *mut c_void,
+    custom_data_deleter: Option<unsafe extern "C" fn(*mut c_void)>,
+}
+
+impl std::ops::Drop for LastError {
+    fn drop(&mut self) {
+        if let Some(deleter) = self.custom_data_deleter {
+            unsafe { deleter(self.custom_data) };
+        }
+    }
+}
 
 // Save the last error message in thread local storage.
 //
 // This is marginally better than a standard global static value because it
 // allow multiple threads to each have separate errors conditions.
 thread_local! {
-    pub static LAST_ERROR_MESSAGE: RefCell<CString> = RefCell::new(CString::new("").expect("invalid C string"));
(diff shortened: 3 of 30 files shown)


rust/metatensor-sys/src/c_api.rs
```diff
@@ -273,30 +286,25 @@ extern "C" {
         names_count: usize,
         array: mts_array_t,
     ) -> *const mts_labels_t;
-    #[must_use]
```
HaoZeke, line 276: This doesn't seem right? Since every failure gives `MTS_CALLBACK_ERROR` with context set in `LAST_ERROR` doesn't this mean a Rust caller without `?` now gets no warning while the thread-local accumulates stale error state?

HaoZeke: Same consideration to all of these, I think there's maybe something in the binding generation to fix this, somethigng like `--must-use-type "mts_status_t"` maybe?

python/metatensor_core/metatensor/status.py
```diff
@@ -17,41 +18,73 @@ def __init__(self, message, status=None):
...
+        raise _get_exception()
+
+
+def _delete_exception(exception):
+    # decrement the reference count of the exception
+    exception_ptr = ctypes.py_object(exception)
+    ctypes.pythonapi.Py_DecRef(exception_ptr)
+
+
+_DELETE_EXCEPTION = ctypes.CFUNCTYPE(None, ctypes.c_void_p)(_delete_exception)
```
HaoZeke, line 41: This also doesn't seem right? After the Rust `Drop for LastError` calls the deleter, this will be an exception as the address (`PyObject*`) which is an integer right, so the decref hits the int and not the exception... so every exception through `_save_exception` is leaked.

HaoZeke: Should be done the same ways as `_get_exception`, i.e. something like `ctypes.cast(user_data, ctypes.py_object).value` maybe?

python/metatensor_core/metatensor/status.py
```diff
@@ -17,41 +18,73 @@ def __init__(self, message, status=None):
...
-            raise MetatensorError(last_error()) from e
-        else:
-            raise MetatensorError(last_error())
+        raise _get_exception()
+
+
+def _delete_exception(exception):
+    # decrement the reference count of the exception
+    exception_ptr = ctypes.py_object(exception)
+    ctypes.pythonapi.Py_DecRef(exception_ptr)
```
HaoZeke, line 38:
```suggestion
def _delete_exception(exception):
    if not exception:
        return
    obj = ctypes.cast(ctypes.c_void_p(exception), ctypes.py_object).value
    ctypes.pythonapi.Py_DecRef(ctypes.py_object(obj))
```

metatensor-core/include/metatensor/errors.hpp
```diff
@@ -15,49 +17,26 @@ namespace metatensor {
...
-                throw Error("error in C++ callback: " + LastCxxError::message());
+            } else if (status == MTS_CALLBACK_ERROR) {
+                const char* message = nullptr;
+                const char* origin = nullptr;
+                void* data = nullptr;
+                mts_last_error(&message, &origin, &data);
+                if (std::strcmp(origin, "C++ exception") == 0 && data != nullptr) {
+                    std::rethrow_exception(*static_cast<std::exception_ptr*>(data));
+                } else {
+                    throw Error(message);
```
HaoZeke: AFAIK this is mostly fine because the origin is a `CString` pointer which is never `NULL`, but since `mts_last_error` has `std::panic::catch_unwind` without the out-param on the panic branch, this could also still be `origin = nullptr` and ergo be UB? That'd also cause the caller to ignore the return status.

HaoZeke: So maybe something like
```cpp
const char* message = nullptr;
const char* origin = nullptr;
void* data = nullptr;
auto inner = mts_last_error(&message, &origin, &data);
if (inner != MTS_SUCCESS) {
    throw Error("failed to retrieve last error");
}
if (origin != nullptr && std::strcmp(origin, "C++ exception") == 0 && data != nullptr) {
    std::rethrow_exception(*static_cast<std::exception_ptr*>(data));
}
throw Error(message != nullptr ? message : "<no message>");
```

and it should be the same for `check_pointer`

metatensor-core/include/metatensor/errors.hpp
```diff
@@ -71,23 +50,43 @@ namespace metatensor {
...
+                    message = "C++ code threw an exception that was not an std::exception";
+                }
+
+                check_status(mts_set_last_error(
+                    message,
+                    "C++ exception",
+                    exception_ptr,
+                    [](void *ptr) { delete static_cast<std::exception_ptr*>(ptr); }
+                ));
+                return MTS_CALLBACK_ERROR;
```
HaoZeke, line 86: Similar ownership concern, `new std::exception_ptr` is only Rust owned after `mts_set_last_error` returns `MTS_SUCCESS`. If there's a panic from the `catch_unwind`, (`CString::new`?) there'd be a throw before deletion from `check_status` and the `new` exception pointer is orphaned.

python/metatensor_core/metatensor/status.py
```diff
@@ -17,41 +18,73 @@ def __init__(self, message, status=None):
...
+    # increment the reference count of the exception
+    exception_ptr = ctypes.py_object(e)
+    ctypes.pythonapi.Py_IncRef(exception_ptr)
+
+    lib.mts_set_last_error(
+        ctypes.c_char_p(str(e).encode("utf8")),
+        ctypes.c_char_p(b"Python exception"),
+        exception_ptr,
+        _DELETE_EXCEPTION,
+    )
```
HaoZeke: Same problem as the `catch_exceptions` in the C++ side, failure means a metatensor error after bumping the refcount, so this leaks. Needs a try except to decref on failure.

python/metatensor_core/metatensor/status.py
HaoZeke, line 1: So this is kinda messy from a GIL perspective. If there's a thread with `LAST_ERROR` and python `data` after `Py_finalize`, this will crash in `Py_DecRef` after the interpreter shuts down. I don't think this is going to be very common, but at-least for the main thread what about:

```python
import atexit

def _clear_last_error():
    from ._c_lib import _get_library
    try:
        _get_library().mts_set_last_error(None, None, None, None)
    except Exception:
        pass

atexit.register(_clear_last_error)
```

Still not great for worker threads, which are technically unsafe but.. can't think of anything rn.

metatensor-core/src/c_api/status.rs
```diff
@@ -1,69 +1,142 @@
...
+    custom_data: *mut c_void,
+    custom_data_deleter: Option<unsafe extern "C" fn(*mut c_void)>,
+}
+
+impl std::ops::Drop for LastError {
+    fn drop(&mut self) {
+        if let Some(deleter) = self.custom_data_deleter {
+            unsafe { deleter(self.custom_data) };
+        }
+    }
```
HaoZeke, line 21: If we do cast this to a PyObject as we should (see earlier Python comment) then this is UB, `custom_data` has to be not NULL, or we have to somewhere have `data=NULL` => `deleter=NULL`. 

Otherwise this is fine, C++ deleter is null-safe.

python/metatensor_core/metatensor/_c_api.py
```diff
@@ -159,8 +163,19 @@ def setup_functions(lib):
     lib.mts_version.restype = ctypes.c_char_p
 
     lib.mts_last_error.argtypes = [
+        POINTER(ctypes.c_char_p),
+        POINTER(ctypes.c_char_p),
+        POINTER(POINTER(None)),
+    ]
+    lib.mts_last_error.restype = _check_status
```
HaoZeke: If the rust side panics it'll recurse, though technically fine since `_get_exception` has it's own `if status !MTS_SUCCESS`. Might be cleaner to special case `mts_last_error` so the `restype` is `mts_status_t` and have `_get_exception` read the status directly.

rust/metatensor/src/errors.rs
```diff
@@ -1,45 +1,85 @@
...
+    } else {
+        unsafe { CStr::from_ptr(origin).to_str().unwrap_or("<invalid UTF-8 in error origin>") }
+    };
+
+    if !user_data.is_null() && origin == "Rust Error" {
+        let rust_error = unsafe {
+            user_data.cast::<Error>().as_ref().expect("should not be null")
+        };
+        return rust_error.clone();
+    }
```
HaoZeke, line 41: Hypothetical, but if there's a crate with origin string "Rust Error" and data that isn't `Box<metatensor::Error>` won't this read reinterpret as `Error` which would be UB on deref?

Basically similar concern with C++ and easily solved with namespaces.

metatensor-core/src/lib.rs
```diff
@@ -57,7 +53,7 @@ impl std::fmt::Display for Error {
             Error::Io(e) => write!(f, "io error: {}", e),
             Error::Serialization(e) => write!(f, "serialization format error: {}", e),
             Error::BufferSize(e) => write!(f, "buffer is not big enough: {}", e),
-            Error::External { status, context } => write!(f, "external error: {} (status {})", context, status.as_i32()),
+            Error::CallbackError => write!(f, "callback error, last error is {}", crate::c_api::get_last_error_message()),
```
HaoZeke: I think this depends on the thread-local state, not on the construction time state, so this will end up having errors unrelated to why / when the error was raised, should be captured at construction `Error::CallbackError { message: get_last_error_message() }` maybe?

metatensor-core/src/c_api/status.rs
```diff
@@ -1,69 +1,142 @@
...
 impl mts_status_t {
     pub fn is_success(self) -> bool {
-        self.0 == MTS_SUCCESS
+        self == mts_status_t::MTS_SUCCESS
     }
+}
 
-    pub fn as_i32(self) -> i32 {
-        self.0
+impl std::fmt::Display for mts_status_t {
```
HaoZeke, line 86: what about a `match *self`?

metatensor-core/CHANGELOG.md
HaoZeke, line 1: Does anyone need to know that the `MTS_BUFFER_SIZE_ERROR` is now size 4? Probably not?

metatensor-core/tests/cpp/blocks.cpp
```diff
@@ -1,6 +1,8 @@
 #include <fstream>
 #include <sstream>
 
+#include <iostream>
```
HaoZeke:
```suggestion
```

HaoZeke (changes requested):
I have another general query about this... so since this replaces `Err(Error::External { status, ... })` with `Err(Error::CallbackError)` isn't this somehow less interpreatable? What about storing the status on `Error::CallbackError`?

Otherwise this is very nice, and definitely useful :)

metatensor-core/include/metatensor/errors.hpp
Luthaf: Ah, I though `strcmp` already handled null pointers … good catch!

metatensor-core/src/c_api/status.rs
Luthaf: I'll change the deleter on the Python side to handle potentially NULL data, but in practice I don't see how the data could be null if the deleter is set? Since they are both set at the same time

metatensor-core/src/c_api/status.rs
Luthaf: Might be fine now, this is leftover from a time where I could not match on it. I'll update it!

python/metatensor_core/metatensor/_c_api.py
Luthaf: Yes, I change the restun type in `_c_lib`, but I can also change it here if you think it is cleaner.

python/metatensor_core/metatensor/status.py
Luthaf: good point. It should not happen very often (basically only if some code starts Python, run Python + metatensor, stops Python, keep running metatensor); but I'll think about what we can do.

python/metatensor_core/metatensor/status.py
Luthaf: `exception` should already be a `c_void_p`, no? I don't understand what this is trying to prevent, do you have an example of something failing?

Luthaf: ahh, you mean the code here will get an int as `exception`?

rust/metatensor-sys/src/c_api.rs
Luthaf: I need to check, but I think `#[must_use]` was not working because `mts_status_t` is just an alias to another type.

rust/metatensor/src/errors.rs
Luthaf: Yes, but this is also a mostly internal API for the different bindings that is not meant to be used outside of the C API. I guess this could fail if some C API user use `"Rust Error"` as their origin here, and then we hit this bit of code, but I don't see how namespace solve anything?

Luthaf:
> I have another general query about this... so since this replaces Err(Error::External { status, ... }) with Err(Error::CallbackError) isn't this somehow less interpreatable? What about storing the status on Error::CallbackError?

There is no longer a custom status for external/callback error, only `MTS_CALLBACK_ERROR`; so there is nothing to store!

metatensor-core/CHANGELOG.md
Luthaf: no, they should compile against the header, and will pick up the change automatically (this is only possible because it is part of an ABI breaking release!)

metatensor-core/src/lib.rs
Luthaf: true, then I'd rather not capture the message at all since it would be cost for nothing (in general this message should never be seen, and is only about debugging libmetatensor itself)

rust/metatensor-sys/src/c_api.rs
Luthaf: My bad, it works fine!

metatensor-core/src/c_api/status.rs
Luthaf: Ah no, this is a if so we can handle the case of us getting a random value here (since we can't enforce that callbacks don't return random numbers). I'll add a comment.

python/metatensor_core/metatensor/status.py
Luthaf: I guess the only solution is to tell whoever is doing weird things like this to run `mts_set_last_error(nullptr, nullptr, nullptr, nullptr)` themself before terminating the Python interpreter.

python/metatensor_core/metatensor/status.py
Luthaf: yes, good catch! I was missing a test to check this, so this function was never called

metatensor-core/include/metatensor/errors.hpp
```diff
@@ -1,5 +1,9 @@
 #pragma once
 
+#include <cstring>
+
+#include <iostream>
```
HaoZeke: Couldn't we just use `fprintf`? Unless we already include `iostream` I'd rather not just for `cerr`..

python/metatensor_core/metatensor/status.py
```diff
@@ -17,41 +19,100 @@ def __init__(self, message, status=None):
...
+            ctypes.c_char_p(str(e).encode("utf8")),
+            ctypes.c_char_p(b"Python exception"),
+            ctypes.c_void_p.from_buffer(exception_ptr),
+            _DELETE_EXCEPTION,
+        )
+    except Exception as err:
+        # if we failed to save the exception, we are in a very bad state, but we should
+        # still try to report the original error message if possible.
+        print("INTERNAL ERROR: unable to save last error after Python callback failure")
+        print(f"original error was: {e}, error while saving was {err}")
```
HaoZeke: print(file=sys.stderr)

HaoZeke (approved):
Two minor nits.

PR #1117, opened 2026-05-06
Only check major version in as_dlpack

Luthaf:
Trying to use `ExternalCPUArray` with Rust arrays in featomic was failing because the check on the minor version was the wrong way around, but checking again the semantics of `max_version` and how dlpack does versioning, I think we should be OK just checking the major version.

# Reviewer checklist

### DIFF
--- a/metatensor-core/include/metatensor/arrays.hpp
+++ b/metatensor-core/include/metatensor/arrays.hpp
@@ -1449,18 +1449,13 @@ public:
             throw Error("`stream` must be null for CPU data");
         }
 
-        DLPackVersion mta_version = {DLPACK_MAJOR_VERSION, DLPACK_MINOR_VERSION};
-        // SEMVER, so major must match, and the MTA minor must be below the minor(caller)
-        bool major_mismatch = max_version.major != mta_version.major;
-        bool minor_too_high = max_version.minor < mta_version.minor;
-        if (major_mismatch || minor_too_high) {
+        if (max_version.major != DLPACK_MAJOR_VERSION) {
             throw Error(
-                "SimpleDataArray supports DLPack version " +
-                std::to_string(mta_version.major) + "." +
-                std::to_string(mta_version.minor) +
-                ". Caller requested incompatible version " +
-                std::to_string(max_version.major) + "." +
-                std::to_string(max_version.minor)
+                "invalid `max_version` in SimpleDataArray::as_dlpack: "
+                "we got v" + std::to_string(max_version.major) + "." +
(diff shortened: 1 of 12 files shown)


HaoZeke (approved):
TBH this whole max version check was a pain. Happy to stick to major only until / if we ever hit a bug from being slightly not spec compliant..

metatensor-core/include/metatensor/arrays.hpp
```diff
@@ -1449,18 +1449,13 @@ class SimpleDataArray : public metatensor::DataArrayBase {
...
-                std::to_string(mta_version.major) + "." +
-                std::to_string(mta_version.minor) +
-                ". Caller requested incompatible version " +
-                std::to_string(max_version.major) + "." +
-                std::to_string(max_version.minor)
+                "invalid `max_version` in SimpleDataArray::as_dlpack: "
+                "we got v" + std::to_string(max_version.major) + "." +
+                std::to_string(max_version.minor) + ", but we support v" +
+                std::to_string(DLPACK_MAJOR_VERSION) + "." +
+                std::to_string(DLPACK_MINOR_VERSION)
```
HaoZeke, line 1458: Is there any point printing the minor here? Since we don't gate on it anyway anymore..

metatensor-core/src/labels/array.rs
```diff
@@ -217,8 +217,12 @@ unsafe extern "C" fn labels_array_as_dlpack(
         // minimum minor version requirement.
         if max_version.major != current.major {
             return Err(Error::InvalidParameter(
-                format!("`max_version` too high in as_dlpack: got {}, we support {}", max_version.major, current.major)
-            ))
+                format!(
+                    "invalid `max_version` in LabelsArray::as_dlpack: \
+                    we got v{}.{}, but we support v{}.{})",
```
HaoZeke, line 222: Same.

HaoZeke (commented):
Maybe something like we support all versions upto major+1 non inclusive.

metatensor-core/include/metatensor/arrays.hpp
Luthaf: Just makes for a nicer error message IMO, since the full version is what the user is giving us

HaoZeke: Yeah but then we're open to bug reports like hey why can I do `1.8` if "we support `1.5`" is printed 😅 

Or more likely someone silently cursing us and thinking everything is broken.

PR #1121, opened 2026-05-07
Remove CpuLabels in the Rust bindings

Luthaf:
They are more pain than benefit for now, especially when passing data around in featomic.

# Reviewer checklist


HaoZeke (approved):
Wasn't the point of the explicit type to be well, explicit? Nevertheless this is much more compact so in it goes.

Luthaf:
Yes, but there where also issues with lifetimes here, when calling this function on `&'a Labels`. We should be able to make it work, but later!

PR #1126, opened 2026-05-12
Fix equality check when arrays are not contiguous

Luthaf:
This was breaking Labels equality in torch down the line.

xref https://github.com/metatensor/metatrain/pull/1128

# Reviewer checklist

### DIFF
--- a/metatensor-core/include/metatensor/arrays.hpp
+++ b/metatensor-core/include/metatensor/arrays.hpp
@@ -19,12 +19,45 @@
 namespace metatensor {
 namespace details {
     /// Compute the product of all values in the `shape` vector
-    inline size_t product(const std::vector<size_t>& shape) {
+    template<typename Shape>
+    inline size_t product(const Shape& shape) {
         size_t result = 1;
         for (auto size: shape) {
-            result *= size;
+            result *= static_cast<size_t>(size);
         }
         return result;
+    }
+
+    template<typename Shape>
+    inline std::vector<int64_t> contiguous_strides(const Shape& shape) {
+        auto strides = std::vector<int64_t>(shape.size(), 1);
+        if (shape.empty()) {
+            return strides;
+        }
+
+        for (size_t i = shape.size(); i > 1; --i) {
+            auto dim = i - 1;
+            strides[dim - 1] = strides[dim] * static_cast<int64_t>(shape[dim]);
+        }
+        return strides;
+    }
+
+    template<typename Shape>
+    inline bool is_row_major_contiguous(const Shape& shape, const std::vector<int64_t>& strides) {
+        if (shape.size() != strides.size()) {
+            return false;
+        }
+
+        int64_t expected_stride = 1;
+        for (size_t i = shape.size(); i > 0; --i) {
+            auto dim = i - 1;
+            if (static_cast<int64_t>(shape[dim]) != 1 && strides[dim] != expected_stride) {
+                return false;
+            }
+            expected_stride *= static_cast<int64_t>(shape[dim]);
+        }
+
+        return true;
     }
 
     /// Get the N-dimensional index corresponding to the given linear `index`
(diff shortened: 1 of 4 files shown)


metatensor-core/include/metatensor/arrays.hpp
```diff
@@ -158,21 +191,66 @@ namespace details {
...
+        auto indices = std::vector<int64_t>(ndim, 0);
+        const auto* lhs_ptr = lhs.data();
+        const auto* rhs_ptr = rhs.data();
+
+        for (size_t i = 0; i < n_elements; i++) {
+            int64_t lhs_offset = 0;
+            int64_t rhs_offset = 0;
+            for (size_t dim = 0; dim < ndim; dim++) {
+                lhs_offset += indices[dim] * lhs_strides[dim];
+                rhs_offset += indices[dim] * rhs_strides[dim];
```
HaoZeke, line 236: This seems inefficient? It's linear offset for every element?

Maybe a private tracking tuple for the offsets and an incremental update would work better but .. can also leave it for later.

metatensor-core/include/metatensor/arrays.hpp
```diff
@@ -604,15 +694,14 @@ class NDArray {
...
-    /// is 0.
-    bool is_empty() const {
-        for (auto s: shape_) {
-            if (s == 0) {
-                return true;
-            }
-        }
-        return false;
+    /// Get the strides of this array in number of elements.
+    std::vector<int64_t> strides() const {
```
HaoZeke: So this is a by-value return, so a new one each time..

metatensor-core/include/metatensor/arrays.hpp
```diff
@@ -1619,6 +1709,16 @@ class SimpleDataArray : public metatensor::DataArrayBase {
         return data_->data();
     }
 
+    /// Get the strides of this array in number of elements.
+    std::vector<int64_t> strides() const {
+        return details::contiguous_strides(shape_);
+    }
```
HaoZeke: Also by-value.

metatensor-core/include/metatensor/arrays.hpp
```diff
@@ -416,15 +503,16 @@ class DLPackArray {
...
-        for (auto s: shape_) {
-            if (s == 0) {
-                return true;
-            }
-        }
-        return false;
+    /// Get the strides of this array in number of elements.
+    const std::vector<int64_t>& strides() const & {
+        return strides_;
+    }
```
HaoZeke, line 509: But this one is by reference. Maybe get the others to also be by-reference with a private member?

python/metatensor_torch/tests/labels.py
```diff
@@ -672,3 +672,15 @@ def forward(self, x: LabelsEntryWrap) -> LabelsEntryWrap:
...
+
+def test_equal_non_contiguous():
+    values_non_contiguous = torch.tensor([[1, 2, 3], [4, 3, 1]]).T
+    values_contiguous = values_non_contiguous.contiguous()
+
+    labels_non_contiguous = Labels(names=["a", "b"], values=values_non_contiguous)
+    labels_contiguous = Labels(names=["a", "b"], values=values_contiguous)
+
+    print("labels_non_contiguous", labels_non_contiguous)
+    print("labels_contiguous", labels_contiguous)
```
HaoZeke: Do we not have a hygine test preventing print statements in tests?

HaoZeke (changes requested):
Some minor comments.

python/metatensor_torch/tests/labels.py
Luthaf: We should, not sure why it did not catch it here, I'll check.

metatensor-core/include/metatensor/arrays.hpp
Luthaf: fair enough!

metatensor-core/include/metatensor/arrays.hpp
Luthaf: Why private tuple? This is only used in this function. We can always get back to it and speed it up later yes!

python/metatensor_torch/tests/labels.py
Luthaf: ah no, this is only in metatomic. I'll open an issue for this repo as well, we can do it later

HaoZeke (approved):
LGTM, thanks @Luthaf

HaoZeke:
Merge at will post-CI :)

PR #1128, opened 2026-05-12
Select between deep and shallow copies for TensorMap and TensorlBlock

Luthaf:
Ping @pfebrer who asked for it, since we are making another RC we can include the change.

The only surprising thing might be that `copy.copy(tensor)`/`copy.deepcopy(tensor)` will behave differently in -core and -torch. For metatensor-core, this will forward to `tensor.copy(deep=False)`/`tensor.copy(deep=True)`; but for metatensor-torch the deepcopy will go through a full serialization.

# Reviewer checklist

### DIFF
--- a/metatensor-torch/include/metatensor/torch/tensor.hpp
+++ b/metatensor-torch/include/metatensor/torch/tensor.hpp
@@ -27,8 +27,12 @@ public:
     /// Create a new `TensorMapHolder` for TorchScript.
     TensorMapHolder(Labels keys, std::vector<TensorBlock> blocks);
 
-    /// Make a copy of this `TensorMap`, including all the data contained inside
-    TensorMap copy() const;
+    /// Make a copy of this `TensorMap`.
+    ///
+    /// If `deep` is `true`, the blocks and their data will be copied as well;
+    /// otherwise, the new `TensorMap` will increase the reference count of the
+    /// blocks' data and return a shallow copy.
+    TensorMap copy(bool deep = true);
 
     /// Get the keys for this `TensorMap`
     Labels keys() const;
--- a/python/metatensor_core/metatensor/_block.py
+++ b/python/metatensor_core/metatensor/_block.py
@@ -175,13 +175,10 @@ class TensorBlock:
                 self._lib.mts_block_free(self._actual_ptr)
 
     def __copy__(self):
-        raise ValueError(
-            "shallow copies of TensorBlock are not possible, use a deepcopy instead"
-        )
+        return self.copy(deep=False)
 
     def __deepcopy__(self, _memodict):
-        new_ptr = self._lib.mts_block_copy(self._ptr)
-        return TensorBlock._from_ptr(new_ptr, parent=None)
+        return self.copy(deep=True)
 
     def __reduce__(self):
         raise NotImplementedError(
@@ -203,11 +200,30 @@ class TensorBlock:
         """
         return self.values.shape
 
-    def copy(self) -> "TensorBlock":
+    def copy(self, deep: bool = True) -> "TensorBlock":
         """
-        get a deep copy of this block, including all the data and metadata
+        Get a copy of this block, with the same values and labels. If ``deep`` is
+        ``True``, also make a full copy of the values; otherwise, the values in the new
+        block will share the same memory as those in this block.
+
+        :param deep: if ``True``, create a deep copy of the block
         """
-        return copy.deepcopy(self)
+        if deep:
+            new_ptr = self._lib.mts_block_copy(self._ptr)
+            return TensorBlock._from_ptr(new_ptr, parent=None)
+        else:
+            new_block = TensorBlock(
+                values=self.values,
+                samples=self.samples,
+                components=self.components,
+                properties=self.properties,
+            )
+
+            for parameter in self.gradients_list():
+                gradient = self.gradient(parameter)
+                new_block.add_gradient(parameter, gradient.copy(deep=False))
+
+            return new_block
 
     def __repr__(self) -> str:
         if self._actual_ptr is None:
--- a/python/metatensor_core/tests/blocks.py
+++ b/python/metatensor_core/tests/blocks.py
@@ -269,7 +269,7 @@ def test_gradients(block_components):
     assert_equal(gradient.values, np.full((2, 3, 2, 2), 11.0))
 
 
-def test_copy():
+def test_deep_copy():
     block = TensorBlock(
         values=np.full((3, 3, 2), 2.0),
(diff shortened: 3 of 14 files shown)


HaoZeke (commented):
My two cents on this is that most com chem folks I know don't try to understand the context of shallow/deep copying and will end up peppering everything with `deep=True` but I guess that's orthogonal to the issue.

metatensor-torch/include/metatensor/torch/tensor.hpp
```diff
@@ -27,8 +27,12 @@ class METATENSOR_TORCH_EXPORT TensorMapHolder: public torch::CustomClassHolder {
...
     TensorMapHolder(Labels keys, std::vector<TensorBlock> blocks);
 
-    /// Make a copy of this `TensorMap`, including all the data contained inside
-    TensorMap copy() const;
+    /// Make a copy of this `TensorMap`.
+    ///
+    /// If `deep` is `true`, the blocks and their data will be copied as well;
+    /// otherwise, the new `TensorMap` will increase the reference count of the
+    /// blocks' data and return a shallow copy.
+    TensorMap copy(bool deep = true);
```
HaoZeke, line 35: Doesn't this still have to be `const`?

```suggestion
    TensorMap copy(bool deep = true) const;
```

python/metatensor_core/tests/blocks.py
```diff
@@ -282,10 +282,58 @@ def test_copy():
...
 
     del block
 
     assert id(clone.values) != block_values_id
+    assert_equal(clone.values, np.full((3, 3, 2), 2.0))
+
+    # The samples are always shared
+    assert id(clone.samples.values) == samples_values_id
+    assert id(clone.components[0].values) == component_values_id
+    assert id(clone.properties.values) == properties_values_id
```
HaoZeke, line 297: Maybe nicer to not use `id`
...

```
assert clone.values.__array_interface__['data'][0] == block.values.__array_interface__['data'][0]
```
should probably work

python/metatensor_core/metatensor/_block.py
```diff
@@ -203,11 +200,30 @@ def shape(self):
...
+        if deep:
+            new_ptr = self._lib.mts_block_copy(self._ptr)
+            return TensorBlock._from_ptr(new_ptr, parent=None)
+        else:
+            new_block = TensorBlock(
+                values=self.values,
+                samples=self.samples,
+                components=self.components,
+                properties=self.properties,
+            )
```
HaoZeke, line 220: Shouldn't we actually have an `mts_block_copy_shallow` instead of doing this in the Python layer?

python/metatensor_core/tests/tensor.py
```diff
@@ -94,7 +94,7 @@ def test_constructor_errors():
         TensorMap(keys, ["block"])
 
 
-def test_copy():
+def test_deep_copy():
     # Do not use a fixture here because we want exactly on reference in the copy test.
```
HaoZeke:
```suggestion
    # Do not use a fixture here because we want exactly one reference in the copy test.
```

HaoZeke (changes requested):
Almost there.

metatensor-torch/include/metatensor/torch/tensor.hpp
Luthaf: It would be nice, but `block_by_id` is not const (since one can always modify the block). For now this will be fine, but we should spend some time to improve const correctness someday

python/metatensor_core/metatensor/_block.py
Luthaf: We can't, there is no API in mts_array_t to make a shallow copy of the values. And I don't think we want to add one.

python/metatensor_core/tests/blocks.py
Luthaf: TBH I don't see how this is better than `id`? We don't use id in the torch tests because they recreate new Python objects for the same tensor, but we actually want to have the same python object here

HaoZeke: Yes but the CPython allocator recycles memory and this basically tests if the bindings are caching the right object instead of the underlying data?

Shouldn't happen here but if `block.values` gets garbage collected it could still allow `id()` checks to be true... I thought it was generally just a bad idea to use `id` because of 

> Two objects with non-overlapping lifetimes may have the same id() value.

https://docs.python.org/3/library/functions.html#id

python/metatensor_core/metatensor/_block.py
HaoZeke: but that diverges the capabilities of the C/C++ stuff and the Python stuff right

metatensor-torch/include/metatensor/torch/tensor.hpp
HaoZeke: xref: https://github.com/metatensor/metatensor/issues/1130

pfebrer:
So that I understand, this means that in metatensor-torch `copy.copy(tensormap)` will actually do a deep copy?

But `tensormap.copy(deep=False)` will work fine in both cases?

Luthaf:
> So that I understand, this means that in metatensor-torch copy.copy(tensormap) will actually do a deep copy?

I'm not 100% sure, but I think it will do a shallow copy as well, but at the `torch::intrusive_ptr` level instead of creating a new TensorBlock/TensorMap with new references to the same data.

> But tensormap.copy(deep=False) will work fine in both cases?

yes!

python/metatensor_core/metatensor/_block.py
Luthaf: Yes, but this is already the case for a lot of things

python/metatensor_core/tests/blocks.py
Luthaf: but here they have overlapping lifetimes

HaoZeke: fair enough ^_^

python/metatensor_core/metatensor/_block.py
HaoZeke: Would be nice to track later, someday an actual unified thing would be nice.

HaoZeke (approved):
Lets get it in prod for @pfebrer, thanks @Luthaf

pfebrer:
Ah ok, as long as `tensormap.copy(deep=False)` works, then great! I imagine it is faster than the deep copy, no? Have you tested? haha

Luthaf:
I have not tested, but yes, a deep copy is doing strictly more work, so it will be slower for sure. How much slower will depend on how much data is stored in the block/TensorMap

PR #1148, opened 2026-06-16
Make `Labels` fully compatiable with jvp mode

GardevoirX:
As a follow-up of #1055 

Basically, the following operations are not compatible with jvp mode:
- creation of `torch::Tensor` from `std::vector`
- using `Tensor::data_ptr` when `Tensor::has_storage() == False`

To reproduce (thanks to chatty):
```python
import torch
from metatensor.torch import Labels

print('torch', torch.__version__)
try:
    print('metatensor.torch', mts.__version__)
except Exception:
    pass

labels = Labels(['a', 'b'], torch.tensor([[0, 0], [1, 0], [2, 1]], dtype=torch.int32))
other = Labels(['a', 'b'], torch.tensor([[1, 0], [3, 1]], dtype=torch.int32))
selection = Labels(['a'], torch.tensor([[1], [2]], dtype=torch.int32))
entry_tensor = torch.tensor([1, 0], dtype=torch.int32)

x = torch.tensor([1.0, 2.0], requires_grad=True)
tangent = torch.tensor([0.1, 0.2])

def as_float(extra):
    if extra is None:
        return torch.tensor(0.0)
    if isinstance(extra, torch.Tensor):
        return extra.to(torch.float32).sum() * 0.0
    if isinstance(extra, int):
        return torch.tensor(float(extra)) * 0.0
    if isinstance(extra, bool):
        return torch.tensor(float(extra)) * 0.0
    return torch.tensor(0.0)

def run_case(name, op):
    def fn(x):
        return x.sum() + as_float(op())
    try:
        eager = fn(x)
    except Exception as e:
        print(f'{name:40s} eager_error {type(e).__name__}: {str(e).splitlines()[0]}')
        return
    try:
        y, dy = torch.func.jvp(fn, (x,), (tangent,))
        print(f'{name:40s} OK y={float(y):.1f} dy={float(dy):.1f}')
    except Exception as e:
        print(f'{name:40s} JVP_ERROR {type(e).__name__}: {str(e).splitlines()[0]}')

cases = [
    ('values.sum', lambda: labels.values.sum()),
    ('column.sum', lambda: labels.column('a').sum()),
    ('position tensor constant outer', lambda: labels.position(entry_tensor)),
    ('position tensor constructed inner', lambda: labels.position(torch.tensor([1, 0], dtype=torch.int32))),
    ('position list', lambda: labels.position([1, 0])),
    ('select.sum', lambda: labels.select(selection).sum()),
    ('union.values.sum', lambda: labels.union(other).values.sum()),
    ('intersection.values.sum', lambda: labels.intersection(other).values.sum()),
    ('difference.values.sum', lambda: labels.difference(other).values.sum()),
    ('union_and_mapping first map sum', lambda: labels.union_and_mapping(other)[1].sum()),
    ('union_and_mapping second map sum', lambda: labels.union_and_mapping(other)[2].sum()),
    ('intersection_and_mapping map sum', lambda: labels.intersection_and_mapping(other)[1].sum()),
    ('difference_and_mapping map sum', lambda: labels.difference_and_mapping(other)[1].sum()),
    ('append.values.sum', lambda: labels.append('c', torch.tensor([5, 6, 7], dtype=torch.int32)).values.sum()),
    ('insert.values.sum', lambda: labels.insert(1, 'c', torch.tensor([5, 6, 7], dtype=torch.int32)).values.sum()),
    ('permute.values.sum', lambda: labels.permute([1, 0]).values.sum()),
    ('remove.values.sum', lambda: labels.remove('b').values.sum()),
    ('rename.values.sum', lambda: labels.rename('b', 'bb').values.sum()),
    ('to(cpu).values.sum', lambda: labels.to('cpu').values.sum()),
    ('Labels.empty.values.sum', lambda: Labels.empty(['x', 'y']).values.sum()),
    ('Labels.range.values.sum', lambda: Labels.range('x', 3).values.sum()),
    ('Labels.single.values.sum', lambda: Labels.single().values.sum()),
    ('Labels constructor values.sum', lambda: Labels(['x'], torch.tensor([[0], [1]], dtype=torch.int32)).values.sum()),
]

for name, op in cases:
    run_case(name, op)
```
Every case should pass with this branch, and fail with the main branch

# Reviewer checklist

### DIFF
--- a/metatensor-core/tests/utils/mod.rs
+++ b/metatensor-core/tests/utils/mod.rs
@@ -200,6 +200,7 @@ pub fn create_python_venv(build_dir: PathBuf) -> PathBuf {
         let mut cmd = Command::new(find_python());
         cmd.arg("-m");
         cmd.arg("venv");
+        cmd.arg("--clear");
         cmd.arg(&build_dir);
 
         run_command(cmd, "python to create virtualenv with `venv`");
--- a/metatensor-core/tests/utils/mod.rs
+++ b/metatensor-core/tests/utils/mod.rs
@@ -200,7 +200,6 @@ pub fn create_python_venv(build_dir: PathBuf) -> PathBuf {
         let mut cmd = Command::new(find_python());
         cmd.arg("-m");
         cmd.arg("venv");
-        cmd.arg("--clear");
         cmd.arg(&build_dir);
 
         run_command(cmd, "python to create virtualenv with `venv`");
--- a/metatensor-core/tests/utils/mod.rs
+++ b/metatensor-core/tests/utils/mod.rs
@@ -307,7 +307,7 @@ pub fn setup_torch_pip(python: &Path) -> PathBuf {
 /// Install metatensor in a Python virtualenv with pip, and return the
 /// CMAKE_PREFIX_PATH for the installed metatensor.
 pub fn setup_metatensor_pip(python: &Path, source_dir: &Path) -> PathBuf {
-    pip_install(python, &["setuptools>=77", "packaging>=23", "cmake"], PipInstallOptions::default());
+    pip_install(python, &["setuptools>=77", "wheel", "packaging>=23", "cmake"], PipInstallOptions::default());
 
     pip_install(
         python,
--- a/metatensor-core/tests/utils/mod.rs
+++ b/metatensor-core/tests/utils/mod.rs
@@ -200,7 +200,6 @@ pub fn create_python_venv(build_dir: PathBuf) -> PathBuf {
         let mut cmd = Command::new(find_python());
         cmd.arg("-m");
         cmd.arg("venv");
-        cmd.arg("--clear");
         cmd.arg(&build_dir);
 
         run_command(cmd, "python to create virtualenv with `venv`");
@@ -308,7 +307,7 @@ pub fn setup_torch_pip(python: &Path) -> PathBuf {
 /// Install metatensor in a Python virtualenv with pip, and return the
 /// CMAKE_PREFIX_PATH for the installed metatensor.
 pub fn setup_metatensor_pip(python: &Path, source_dir: &Path) -> PathBuf {
-    pip_install(python, &["setuptools>=77", "wheel", "packaging>=23", "cmake"], PipInstallOptions::default());
+    pip_install(python, &["setuptools>=77", "packaging>=23", "cmake"], PipInstallOptions::default());
 
     pip_install(
         python,
--- a/metatensor-torch/src/labels.cpp
+++ b/metatensor-torch/src/labels.cpp
@@ -557,26 +557,30 @@ std::tuple<Labels, torch::Tensor> LabelsHolder::difference_and_mapping(const Lab
 
 torch::Tensor LabelsHolder::select(const Labels& selection) const {
     auto options = torch::TensorOptions().dtype(torch::kInt64).device(torch::kCPU);
-    auto selected = torch::zeros({this->count()}, options);
-    auto selected_count = static_cast<size_t>(selected.size(0));
+    std::vector<uint64_t> selected(this->count());
+    auto selected_count = selected.size();
 
     auto device = this->device();
     if (this->count() == 0) {
-        return selected.to(device);
+        auto tensor_selected = torch::zeros({0}, options);
+        return tensor_selected.to(device);
     }
 
     labels_.select(
         selection->labels_,
-        // we assume that we can pass a pointer to `int64_t` instead of
-        // `uint64_t`, which should be fine on all platforms since they all use
-        // 2-complement representation and the values should always be positive.
-        reinterpret_cast<uint64_t*>(selected.mutable_data_ptr()),
+        selected.data(),
         &selected_count
     );
+    selected.resize(selected_count);
 
-    selected.resize_({static_cast<int64_t>(selected_count)});
+    auto selected_i64 = std::vector<int64_t>();
+    selected_i64.reserve(selected.size());
+    for (auto value: selected) {
+        selected_i64.push_back(static_cast<int64_t>(value));
+    }
+    auto tensor_selected = torch::tensor(selected_i64, options);
 
-    return selected.to(device);
+    return tensor_selected.to(device);
 }
 
 struct LabelsPrintData {
--- a/metatensor-torch/src/labels.cpp
+++ b/metatensor-torch/src/labels.cpp
@@ -516,26 +516,30 @@ Labels LabelsHolder::set_union(const Labels& other) const {
 
 std::tuple<Labels, torch::Tensor, torch::Tensor> LabelsHolder::union_and_mapping(const Labels& other) const {
     auto options = torch::TensorOptions().dtype(torch::kInt64).device(torch::kCPU);
-    auto first_mapping = torch::zeros({this->count()}, options);
-    auto second_mapping = torch::zeros({other->count()}, options);
-    std::vector<int64_t> first_mapping_vector(this->count());
-    std::vector<int64_t> second_mapping_vector(other->count());
+    auto first_mapping_vector = std::make_shared<std::vector<int64_t>>(this->count());
+    auto second_mapping_vector = std::make_shared<std::vector<int64_t>>(other->count());
 
     auto result = torch::make_intrusive<LabelsHolder>(labels_.set_union(
         other->labels_,
-        first_mapping_vector.data(),
-        first_mapping_vector.size(),
-        second_mapping_vector.data(),
-        second_mapping_vector.size()
+        first_mapping_vector->data(),
+        first_mapping_vector->size(),
+        second_mapping_vector->data(),
+        second_mapping_vector->size()
     ));
-
+    
     auto device = this->device();
-    for (size_t i = 0; i < first_mapping_vector.size(); i++) {
-        first_mapping.index_put_({static_cast<int64_t>(i)}, first_mapping_vector[i]);
-    }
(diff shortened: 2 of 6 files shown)


metatensor-torch/src/labels.cpp
```diff
@@ -557,26 +557,30 @@ std::tuple<Labels, torch::Tensor> LabelsHolder::difference_and_mapping(const Lab
 
 torch::Tensor LabelsHolder::select(const Labels& selection) const {
     auto options = torch::TensorOptions().dtype(torch::kInt64).device(torch::kCPU);
-    auto selected = torch::zeros({this->count()}, options);
-    auto selected_count = static_cast<size_t>(selected.size(0));
+    std::vector<uint64_t> selected(this->count());
```
Luthaf: I don't understand why this doesn't work? Since the tensor is not on the meta device. 

I don't love using a std::vector here, since this adds two copies of the data and the associated slowdown

Luthaf:
Can you please add the example in this PR as tests?

metatensor-torch/src/labels.cpp
GardevoirX: Because in jvp mode a tensor is wrapped with a `TensorWrapper`, which (idk where to find the source and chatty told me and seemed to be so) has no storage (similar to meta tensor) but all other things, like `dtype` and `device`. 
https://github.com/pytorch/pytorch/blob/7d287fc446e5b99bca5d1d6de467ce5897a223fc/aten/src/ATen/functorch/TensorWrapper.h#L17-L42

maybe the problem of two copies can be solved through `torch::from_blob`

Luthaf: yes, but my understanding was that this wrapping only applies to inputs to the jvp function, not anything happening inside compiled C++ code

GardevoirX: It's applied to the tensor created inside a function where we request jvp, an example from chatty:

`ext.cpp`:
```cpp
#include <torch/extension.h>
#include <cstdint>
#include <memory>
#include <vector>

// Does not use the JVP input. It creates a tensor internally through C++ torch API
// and then treats it as a raw writable C buffer.
torch::Tensor internal_zeros_then_data_ptr() {
    auto options = torch::TensorOptions().dtype(torch::kInt64).device(torch::kCPU);
    auto selected = torch::zeros({4}, options);
    std::cout << "has_storage: " << selected.has_storage() << std::endl;
    auto* ptr = selected.mutable_data_ptr<int64_t>();
    for (int64_t i = 0; i < 4; i++) {
        ptr[i] = i + 10;
    }

    return selected;
}

// Metadata-safe version: write a real C++ buffer first, then expose it as tensor.
torch::Tensor internal_vector_then_from_blob() {
    auto data = std::make_shared<std::vector<int64_t>>(4);
    for (int64_t i = 0; i < 4; i++) {
        (*data)[i] = i + 10;
    }

    auto options = torch::TensorOptions().dtype(torch::kInt64).device(torch::kCPU);
    return torch::from_blob(
        data->data(),
        {static_cast<int64_t>(data->size())},
        [data](void*) {},
        options
    );
}

PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
    m.def("internal_zeros_then_data_ptr", &internal_zeros_then_data_ptr);
    m.def("internal_vector_then_from_blob", &internal_vector_then_from_blob);
}
```

`run.py`
```python
import torch
import repro_cpp_internal_tensor_jvp_ext as ext

x = torch.tensor([1.0, 2.0], requires_grad=True)
tangent = torch.tensor([0.1, 0.2])

def as_float(t):
    return t.to(torch.float32).sum() * 0.0

def run(name, op):
    def fn(x):
        # x is only here to activate jvp. op() creates tensors internally in C++.
        return x.sum() + as_float(op())

    print(f"\n{name}")
    print("eager:", fn(x))
    try:
        y, dy = torch.func.jvp(fn, (x,), (tangent,))
        print("jvp:", y, dy)
    except Exception as e:
        print("jvp error:", type(e).__name__, str(e).splitlines()[0])

run("C++ torch::zeros + mutable_data_ptr", ext.internal_zeros_then_data_ptr)
run("C++ std::vector + from_blob", ext.internal_vector_then_from_blob)
```

`setup.py`
```python
from setuptools import setup
from torch.utils.cpp_extension import BuildExtension, CppExtension

setup(
    name="repro_cpp_internal_tensor_jvp_ext",
    ext_modules=[
        CppExtension(
            "repro_cpp_internal_tensor_jvp_ext",
            ["ext.cpp"],
            extra_compile_args={"cxx": ["-O0"]},
        )
    ],
    cmdclass={"build_ext": BuildExtension.with_options(use_ninja=False)},
)
```

```bash
python setup.py build_ext --inplace
python run.py
```
output:
```
C++ torch::zeros + mutable_data_ptr
has_storage: 1
eager: tensor(3., grad_fn=<AddBackward0>)
has_storage: 0
jvp error: RuntimeError Cannot access data pointer of Tensor that doesn't have storage

C++ std::vector + from_blob
eager: tensor(3., grad_fn=<AddBackward0>)
jvp: tensor(3., grad_fn=<AliasBackward0>) tensor(0.3000)
```

Luthaf: ok, so they change global state inside torch when running jvp … Thanks for the clear example!

metatensor-core/tests/utils/mod.rs
```diff
@@ -307,7 +308,7 @@ pub fn setup_torch_pip(python: &Path) -> PathBuf {
 /// Install metatensor in a Python virtualenv with pip, and return the
 /// CMAKE_PREFIX_PATH for the installed metatensor.
 pub fn setup_metatensor_pip(python: &Path, source_dir: &Path) -> PathBuf {
-    pip_install(python, &["setuptools>=77", "packaging>=23", "cmake"], PipInstallOptions::default());
+    pip_install(python, &["setuptools>=77", "wheel", "packaging>=23", "cmake"], PipInstallOptions::default());
```
Luthaf: We don't need wheel anymore, bdist_wheel is now part of setuptools.

metatensor-core/tests/utils/mod.rs
```diff
@@ -200,6 +200,7 @@ pub fn create_python_venv(build_dir: PathBuf) -> PathBuf {
         let mut cmd = Command::new(find_python());
         cmd.arg("-m");
         cmd.arg("venv");
+        cmd.arg("--clear");
```
Luthaf: We don't want to clear here, ideally we should be able to re-use the venv across multiple test runs

python/metatensor_core/pyproject.toml
```diff
@@ -36,6 +36,7 @@ changelog = "https://docs.metatensor.org/latest/core/CHANGELOG.html"
 [build-system]
 requires = [
     "setuptools >=77",
+    "wheel",
```
Luthaf: Same, should not be required since setuptools v70: https://setuptools.pypa.io/en/latest/history.html#v70-1-0

metatensor-torch/src/labels.cpp
```diff
@@ -332,7 +332,11 @@ Labels LabelsHolder::permute(std::vector<int64_t> dimensions_indexes) const {
     }
 
     auto new_names = std::vector<std::string>();
-
+    auto dimensions_indexes_tensor = torch::empty(
+        {static_cast<int64_t>(dimensions_indexes.size())},
+        torch::TensorOptions().dtype(torch::kInt64).device(torch::kCPU)
+    );
+    size_t i = 0;
     for (auto index : dimensions_indexes) {
```
Luthaf: Can you rewrite this loop to use i as a loop index and then access `dimensions_indexes[i]` as needed?

metatensor-torch/src/labels.cpp
```diff
@@ -431,15 +437,35 @@ torch::optional<int64_t> LabelsHolder::position(torch::IValue entry) const {
...
+        const int32_t* entry;
+        std::vector<int32_t> entry_vector;
+        if (values.has_storage()) {
+            entry = static_cast<const int32_t*>(values.data_ptr());
+        } else {
+            for (int64_t i = 0; i < values.size(0); i++) {
+                entry_vector.push_back(values[i].item<int32_t>());
+            }
+            entry = entry_vector.data();
+        }
```
Luthaf, line 450: I don't get this one. If there is no storage, how can you get items?

GardevoirX: By editing the example showed above as this
```c++
torch::Tensor internal_zeros_then_data_ptr() {
    auto options = torch::TensorOptions().dtype(torch::kInt64).device(torch::kCPU);
    auto selected = torch::zeros({4}, options);
    std::cout << "has_storage: " << selected.has_storage() << std::endl;
    std::cout << "the first value: " << selected[0].item<int32_t>() << std::endl;
    auto* ptr = selected.mutable_data_ptr<int64_t>();
    for (int64_t i = 0; i < 4; i++) {
        ptr[i] = i + 10;
    }
    return selected;
}
```
the output becomes
```
C++ torch::zeros + mutable_data_ptr
has_storage: 1
the first value: 0
eager: tensor(3., grad_fn=<AddBackward0>)
has_storage: 0
the first value: 0
jvp error: RuntimeError Cannot access data pointer of Tensor that doesn't have storage

C++ std::vector + from_blob
eager: tensor(3., grad_fn=<AddBackward0>)
jvp: tensor(3., grad_fn=<AliasBackward0>) tensor(0.3000)
```
So yes one can call `.item` even if the tensor doesn't have storage

PR #1124, opened 2026-05-08
feat: mmap loader

HaoZeke:
Reopens #1060.

# Contributor (creator of pull-request) checklist

 - [ ] Tests updated (for new features and bugfixes)?
 - [ ] Documentation updated (for new features)?
 - [ ] Issue referenced (for PRs that solve an issue)?

# Reviewer checklist

 - [ ] CHANGELOG updated with public API or any other important changes?

PR #1165, opened 2026-07-16
Add `register_buffer` on `mts.learn.nn.Module`

### LINKED ISSUE #1027: Implement our version of `Module.register_buffer` and `torch.nn.Buffer`
Currently our `Module` can only move `metatensor` stuff, which causes problems in metatensor/metatrain/pull/938 when we try to move `torch.Tensor` with `.to`. After discussing with @Luthaf , the solution might be adding these two things.

Luthaf:
Only data stored with this function will be moved by `to()` and included in the `state_dict`.

This is required for mmap loading in #1124: we need the data to be aligned to load it with mmap, meaning we must specify alignment for files in the .mts, which is only available on recent version of the zip crate. But recent version of the zip crate also create files with a different size, which breaks metatrain (previous change had to be reverted in https://github.com/metatensor/metatensor/pull/1016).

So the idea is to release this, update metatrain to use it, and then change the zip version used in the code to enable mmap loading.

Fix #1027

# Reviewer checklist


Luthaf:
(might be worth to also do a metatrain PR pulling from this branch to make sure everything works before going through a release, I'll see if I find the time)

metatensor-torch/src/module.cpp
```diff
@@ -133,10 +136,36 @@ void Module::to_impl_(
...
+                "metatensor-learn and use `register_buffer` explicitly "
+                "to remove this warning."
+            );
+            for (const auto& attr : module.named_attributes(false)) {
+                attr_names.push_back(attr.name);
+            }
+        }
+
+        for (const auto& name : attr_names) {
+            if (module.hasattr(name)) {
```
GardevoirX: should there be an else-branch?

GardevoirX: and the similar logic in `python/metatensor_torch/metatensor_torch/_module.py` doesn't have this `if`

python/examples/learn/6-nn-custom-module.py
```diff
@@ -0,0 +1,290 @@
...
+assert shorthand.labels.names == ["p"]
+
+
+# %%
+#
+# Non-persistent buffers
+# ----------------------
+#
+# Just like PyTorch, ``register_buffer`` accepts a ``persistent`` argument.
+# Non-persistent buffers are excluded from ``state_dict`` but are still moved by
```
GardevoirX, line 222: Would be better if this could also be in the documentation?

python/metatensor_learn/metatensor_learn/nn/_module.py
```diff
@@ -77,14 +166,33 @@ def _apply(self, fn, recurse=True):
...
+        or isinstance_metatensor(value, "TensorMap")
+    ):
+        return True
+    elif isinstance(value, dict):
+        return any(_contains_metatensor(v) for v in value.values())
+    elif isinstance(value, (list, tuple)):
+        return any(_contains_metatensor(v) for v in value)
+    return False
+
+
```
GardevoirX, line 197: not sure if `mts.learn.nn.Module` can also be considered metatensor type

metatensor-torch/tests/create-test-module.py
```diff
@@ -21,69 +23,101 @@ def _create_tensor(key_name):
...
+                _create_tensor(name),
+            ),
         )
 
         # check nested modules
         self.a = LabelsModule(name)
         self.b = BlockModule(name)
         self.c = TensorModule(name)
+        self.d = EmptyModule(name)
+        self.e = EmptyModule(name)
```
GardevoirX: Can we add the vanilla `tensor` into here? When I was playing with this stuff half a year ago, there was this bug:
```python
import torch
from metatensor.torch.learn.nn import Module

class TestModule(Module):
    def __init__(self):
        super().__init__()
        self.register_buffer("tensor", torch.full((5,), 2.0))
        
t = TestModule()
ts = torch.jit.script(t)
ts.to("cuda")
```
This made `_parse_rsm_attributes` complain about `RuntimeError: failed to parse line in attributes: '     2'`

GardevoirX (commented):
LGTM, but I have some questions!

python/metatensor_learn/metatensor_learn/nn/_module.py
Luthaf: it is already a nn.Module, so it should be handled outside of this function when looping over sub-modules

metatensor-torch/tests/create-test-module.py
Luthaf: sure, I'll do this!

metatensor-torch/src/module.cpp
Luthaf: no, the if is not needed, this is leftover from a previous version

python/examples/learn/6-nn-custom-module.py
Luthaf: It was already there, but I tried to clarify more. Let me know!

GardevoirX (approved):
Thanks!

PR #1167, opened 2026-07-22
Prevent end-of-line warnings when computing version number

johannes-spies:
Similar to `metatomic`, we prevent warnings about CRLF/LF conversions when computing the version number.

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatensor/actions/artifacts/8546556377.zip)


Luthaf (approved):
Thanks!

PR #1178, opened 2026-08-10
Convert mixed dict and list values in Module.to

HaoZeke:
## Summary

`Module.to` converts metatensor values inside tuples and leaves the rest.
Dicts and lists raised:

```
dict containing both metatensor and non-metatensor data as values are not supported
```

That is the GROMACS crash on `examples/ml-mm`.
`mtt export` of the pet-mad-xs checkpoint (metatrain 2026.3.1, metatensor-learn 0.5) does not write `_mts_buffer_names`.
The C++ fallback then walks every attribute and hits a mixed dict.

Same conversion as tuples: move Labels / TensorBlock / TensorMap, keep the other values.

## Test plan

- [x] `test_mixed_data` in `metatensor_torch` and `metatensor_learn` (dict + list, dtype move, non-metatensor entry stays)
- [ ] `tox -e learn-torch-tests` on `tests/module.py::test_mixed_data`
- [ ] `tox -e torch-tests` on `tests/learn/module.py::test_mixed_data`
- [ ] cookbook `ml-mm` `gmx_mpi mdrun` after `libmetatensor-torch` has this change


Luthaf:
So this only applies to models exported with an old version of metatensor-learn, right? In this case, I'd rather have a "legacy" flag, and keep the throwing code around, wrapped inside a `if (!legacy)` check. This way we can remove the legacy code cleanly in the future.

HaoZeke:
Yes — only old exports (no `_mts_buffer_names`). Modern modules still throw on mixed dict/list.

Wired it as an internal `legacy` flag on the conversion helpers, set from the existing signal rather than a new public knob:

- `legacy = !module.hasattr("_mts_buffer_names")` in C++ `Module::to` and the Python scripted `_apply_metatensor` path
- `if (!all_changed && !legacy) throw` for dict/list; tuples still convert mixed either way
- pure `learn.nn.Module` always has `_mts_buffer_names`, so `.to()` stays strict

That matches the pet-mad / GROMACS case (attribute walk on old scripted models) and keeps the throw for new code so mixed containers do not get papered over. When the no-`_mts_buffer_names` fallback is removed, the legacy convert branch goes with it.

Tests: modern raise restored, legacy convert unit tests, tuple convert, plus C++ Module + Legacy module fixtures.

HaoZeke:
@Luthaf the `legacy` gate is in, and CI is green after the ruff wrap on the new tests. Ready for another look when you have a minute.

Luthaf (approved):

PR #1182, opened 2026-09-11
Check if libmetatensor/libmetatensor_torch are already loaded

Luthaf:
Otherwise we could end up in a state where some engine loads libmetatensor, and then calls back to Python which loads a separate library.

Similar to the TODO item in https://github.com/metatensor/metatomic/pull/324, we need to allow C/C++ code to load the libraries, and then re-use the already loaded libraries from Python callbacks.

# Reviewer checklist


GardevoirX (approved):
LGTM thanks!

PR #1184, opened 2026-09-14
Allow for correct treatment of `mts.learn.nn.Module` to nested empty lists

GardevoirX:
The current `Module.to` cannot handle some specific situations in metatrain. For example, when moving this dict, it identifies `[[]]` as non-metatensor data
```python
{'energy': [[]], 'non_conservative_stress': [[Labels
                    xyz_1
                        0
                        1
                        2, Labels
                      xyz_2
                        0
                        1
                        2]]}
```
This is because there is nothing to move in the nested empty lists, and thus `some_changed` is always `False` in this branch, when dealing with `[[]]`
https://github.com/metatensor/metatensor/blob/3b78c07f9fbf85ef1d5848ae9e063dcbe603b2d7/python/metatensor_learn/metatensor_learn/nn/_module.py#L236-L259

~~These codes also misclassify `((), [], {})` into non-metatensor data~~ <- This isn't torchscriptable
https://github.com/metatensor/metatensor/blob/3b78c07f9fbf85ef1d5848ae9e063dcbe603b2d7/python/metatensor_learn/metatensor_learn/nn/_module.py#L236-L238

Setting `some_changed` or the second return value as `True` solves both problems

# Reviewer checklist


python/metatensor_learn/tests/module.py
```diff
@@ -54,8 +54,9 @@ def __init__(self, name):
         list_value = [Labels([name], values)]
         tuple_value = tuple([Labels([name], values)])
         nested_value = {
-            "dict": {42: [[Labels([name], values)], []], 50: []},
+            "dict": {42: [[Labels([name], values)], []], 50: [], 404: [[]]},
```
Luthaf, line 58: Let's also add something with one more level of nesting here, just in case

Luthaf (commented):
Should we also add something similar to the C++ code in https://github.com/metatensor/metatensor/blob/main/metatensor-torch/src/module.cpp?

EricBoittier (approved):
LGTM

Luthaf:
The code looks good, can you add a CHANGELOG item for it in all relevant packages?

